Issue #55, opened 2015-01-07
Pressing an Unknown Key Crashes BDX


SolarLune:
Hey, there. So, it would appear that if you press a key that is unknown to BDX (e.g. Print Screen or NumLock), it will crash the engine. Here's the error
```
Exception in thread "LWJGL Application" java.lang.NullPointerException
        at com.nilunder.bdx.inputs.GdxProcessor.keyDown(GdxProcessor.java:21)
        at com.badlogic.gdx.backends.lwjgl.LwjglInput.processEvents(LwjglInput.java:300)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:199)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:114)
```

GoranM:
:smile:

SolarLune:
Hah, thanks a lot!

... Now I need to fix it. :|

GoranM:
The codeToLog hash doesn't have anything for keys that are not defined on Keyboard, so `keyboard.codeToLog.get(code)` will return null when GDX provides a code we don't handle. Trying to reference `k.hit`, in that case, will cause a NullPointerException.

You just need a clause to avoid that exception, and that should solve the problem.

Although, if you don't feel like doing it, that's no problem - just let me know.

SolarLune:
Thanks for the tip - I've got a patch up for merging (#58).

SolarLune:
Committed.

PR #58, opened 2015-01-09
GdxProcessor Fix

SolarLune:
... to make sure the key exists before modifying its (non-existent) log. Should fix bug #55.


GoranM:
Thanks.

PR #59, opened 2015-01-09
Added blend mode setter and getter.

SolarLune:
Of course, the blend constants would eventually be moved over to BDX, but I think it'd be nice to at least expose the functionality here.

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -9,6 +9,7 @@ import javax.vecmath.Vector4f;
 
 import com.badlogic.gdx.graphics.g3d.Material;
 import com.badlogic.gdx.graphics.g3d.ModelInstance;
+import com.badlogic.gdx.graphics.g3d.attributes.BlendingAttribute;
 import com.badlogic.gdx.graphics.g3d.attributes.ColorAttribute;
 import com.badlogic.gdx.math.Matrix4;
 import com.badlogic.gdx.math.Vector3;
@@ -404,6 +405,29 @@ public class GameObject implements Named{
 			ca.color.set(r, g, b, a);
 
 		}
+	}
+
+	public int[] blendMode(){
+
+		BlendingAttribute ba = (BlendingAttribute) modelInstance.materials.first().get(BlendingAttribute.Type);
+
+		int[] a = {ba.sourceFunction, ba.destFunction};
+
+		return a;
+
+	}
+
+	public void blendMode(int src, int dest){
+
+		for (Material mat : modelInstance.materials){
+
+			BlendingAttribute ba = (BlendingAttribute) mat.get(BlendingAttribute.Type);
+
+			ba.sourceFunction = src;
+			ba.destFunction = dest;
+
+		}
+
 	}
 
 }


GoranM:
I'm not sure about this.

What are you using it for?

SolarLune:
Well, I would use it to set the blend mode of an object (to be additive, for example)...? Not sure what you mean.

GoranM:
> I'm not actively using it in a game right now, but I could.

I meant: Why would you need to?

I'm guessing alpha blend doesn't look that good for certain particle effects (like fire, for example), so you want the ability to set additive?

> I'm unsure of how it would be done otherwise...? Maybe on materials if they eventually get their own BDX classes?

Yes. It's something we would need to think about, but in either case, we shouldn't stuff GameObject with every graphics-related method that comes to mind.

Even "color" was kinda pushing it.

SolarLune:
I'd say color *was* really necessary. That's a pretty basic part of a game engine. Putting it into materials would be a bit awkward to use regularly, as it is awkward to use currently without the helpful `color()` function. Even if it wasn't, `object.mesh.materials.first().color()` is kind of a handful to type every time you want to change the color of something. I figure you should expose some general usage tools now, and if you really want to put stuff elsewhere, then one could transfer the code out of GameObject when the Material class is written. Better to have it in the interim than be another thing to add on the pile to do later.

And yeah, additive blending (or other kinds of blending) are very useful for different special effects. Particle effects like hitsparks or flames rely pretty heavily on additive blending to look good.

GoranM:
Something like `object.materials.color(r, g, b, a)` would make more sense, when one wants to set the color of all materials.

How an object is rendered depends on its materials, so methods that manipulate some form of render state belong either on the container (for bulk operations), or on the material itself.

For something like Text, which can only utilize one material, a color method makes sense, but for GameObject, which can have multiple materials, it doesn't.

I would argue the same for blend mode, only stronger, because it's nowhere near as common in use as color would be (for most projects).

> Better to have it in the interim than be another thing to add on the pile to do later.

I don't see the benefit of adding more methods to GameObject that clearly don't belong there.

We would eventually have to move them, and the act of "moving" a method is far more energy intensive (and far more likely to cause problems) than simply adding a new one; You're effectively breaking the API, and then managing all the related issues.

I think it's much better to wait until we can nail down a proper design for the graphics side of things.

SolarLune:
Eh, it seems like having an extra class for containers of each type of resource would be extra unnecessary chaff that could be sheared from the codebase (MeshArrayListNamed, MaterialArrayListNamed, etc). No need to make the API more complex than necessary. Just put things where they would be with efficiency in mind, rather than being too worried about where they should be logically.

Having the code in the codebase available before moving exposes it to more testing, and allows users to implement usage of it easier in the future. Better to find an issue with it now than find it along with every other issue the new Material class could / would have.

Anyway, closed.

GoranM:
You wouldn't need a MeshArrayListNamed, because each object can only have one mesh at any given time, and having an extra class for a specialized material container wouldn't really present a significantly higher level of complexity (just an additional level of indirection).

I'm worried about where things should logically reside, because that tends to be the basis of an elegant API.

If you're using something so often, to the point where typing an extra word has a significant impact on your productivity, you should abstract things in your own methods, as every other programmer would; An API that provides a logical structure doesn't hamper that.

As for your testing argument: What's to test? You're simply setting the blend mode for all materials - it works.

But ok - You think this is important, and since we already have color, it's not really that much worse, so I'll merge it in.

rafcolson:
I'm confused about this attribute. Looking at the code it's a getter and a setter of a list, but I don't understand what is supposed to be the source or destination.

So it gets the blend mode of the first material on the object and it sets all materials of an object to a specific blend mode. I think this is not the way to do it. Much better to just get the material of the object and change whatever has to be changed about the material.

This is what I had in mind. Not only logical, but more practical in my opinion.
```java
mesh.materials.get(material_index).someAttribute(some_value);
```
or
```java
mesh.materials.get(material_name).someAttribute(some_value);
```
or
```java
for (Material m: mesh.materials) {
   m.someAttribute(some_value);
}
```
and
```java
boolean some_value = mesh.materials.get(material_index).someAttribute();
```

SolarLune:
@rafcolson - For the setter, It sets the [blend mode](http://www.glprogramming.com/red/chapter06.html#name1) (the source and destination blend modes) of the materials of the mesh. You can use this to change how materials are blended; for example, to be blended Additively, like the BGE does with the Add alpha setting in the material window. For constants, you can use the LibGDX's GL_ blend mode constants (which would probably be best moved into BDX when the Material class is created and blending mode is more easy to set and get). 

For the getter, I thought that (like color()) it would be awkward to return a list of each blend mode for each material in the mesh, so it just goes with the first one. It's relatively unlikely that you'd have different blend modes for each material in a mesh - you'd probably have them separated into different objects.

@GoranM - It's cool, I'll refrain from PRs for additional features until 0.1 drops. Thanks for the merge.

rafcolson:
@SolarLune - Thanks for that information on blend modes. And I see now, that, like the color(), it's convenient that it's an attribute of GameObject. But I expect that in future releases such changes could also be made through the material itself.

PR #61, opened 2015-01-14
Added timeLeft() function to Timer class.

SolarLune:
Renamed time() function to the easier to grasp timeUp() function.
Cleaned up some imports in SpriteAnim.java and fixed it to use timeUp().


SolarLune:
@rafcolson - No problem! I don't see where there's any float output, though - it's all double...? The inputs are float, though (as they were before).

rafcolson:
You're right. Forget about my remark. I didn't read carefully. And thanks again.

GoranM:
I think this:

```java
if (action.time())
    action();
```

Is more elegant (semantically) than this:

```java
if (action.timeUp())
    action()
```

In most cases, you're trying to determine if it's time for a certain action to take place, not when time "expires" (which is what "up" implies).

That said: The "interval" semantics I used are probably not as familiar, or as expected as the more popular "countdown" semantics, so I think Timer should be redesigned to better support both:

```java
Timer timer = new Timer(10); // ten seconds - [time(10);]
timer.time(5); // changed my mind - want 5 seconds - [delta = 5;]
timer.time(); // timeLeft, but I omit "Left", because it's implied in this design
timer.done(); // [time() <= 0 ? true : false;]
timer.restart(); // [time(delta);]

Timer action = new Timer(); // 1 second by default
action.tick(); // [d = done(); if (d) restart(); return d;]
```

Does that make sense?

PS: In the future, it would probably be a good idea to create a suggestion issue, where we could discuss these matters in detail, before you go on to make a PR (as suggested on the [Contributing page](https://github.com/GoranM/bdx/wiki/Contributing)).

rafcolson:
Yes, it makes more sense to me. But I'm just not used to the countdown way of doing things. I'd rather have a function to get the time.
```java
import com.nilunder.bdx.utils.*;

Float time = Time() // gets the time since the time started
Time.reset() // sets the time to zero
```
And also this:
```java
import com.nilunder.bdx.utils.*;

Integer tic = Ticker() // gets the amount of tics elapsed since the ticker started
Ticker.reset() // sets the ticker to zero
```

But I guess the Timer function could serve for this also. If I understand correctly you must restart the timer after it was done in order to set the time again? Or will it restart also after it's done when setting the time? For example:

```java
package com.raco.game;

import com.nilunder.bdx.utils.Timer;
import java.util.Map;
import java.util.HashMap;

public class TimeScheduler extends GameObject {
	private Timer timer = new Timer();
	private Map<String, Float> map = new HashMap<String, Float>();
	private List<String> list;
	private Integer index = 0;

	public void init() {
		map.put("logo_screen", 3.0f);
		map.put("intro_screen", 10.0f);
		map.put("title_screen", 20.0f);
		map.put("preview_screen", 30.0f);
		list = (List<String>) map.values().toArray();
		timer.time(map.get(list(index)));
	}

	public void main() {
		if (timer.done()) {
			index += 1;
			if (index == list.size()) {
				end();
				return
			}
			timer.time(map.get(list(index)));
		}
	}

}
```

SolarLune:
I thought it was confusing to have time() check to see when the timer's "finished". I had to look in the source (or on the Wiki) to find the function I wanted, which is, generally, bad in a class as simple as Timer. That's why I renamed it to "timeUp()". It's the function that returns true when the timer's time is up. If you'd rather have it be "done()", I don't like that name as much as it strays away from time terminology, but it's okay I suppose. Just not "time()". It doesn't really make sense.

```java
Timer timer = new Timer(10); // ten seconds - [time(10);]
timer.time(5); // changed my mind - want 5 seconds - [delta = 5;]
timer.time(); // timeLeft, but I omit "Left", because it's implied in this design
timer.done(); // [time() <= 0 ? true : false;]
timer.restart(); // [time(delta);]

Timer action = new Timer(); // 1 second by default
action.tick(); // [d = done(); if (d) restart(); return d;]
```

Yeah, this is OK. I think a pause() and resume() function would be useful as well, as this would allow you to actually pause and resume timers (like for when a game is paused).

OK, thanks for the info, and sorry about that. I thought I was making some fairly easy to understand commits, but I'll make an issue for additions in the future.

GoranM:
> Just not "time()". It doesn't really make sense.

It makes sense if your timer variable carries the name of the action, as I demonstrated before:

```java
if (action.time())
    action();
```

But I guess most people wouldn't think to do that, so I think the change is warranted.

For "timeUp": Yea, partly, I see your point, but I also think one could confuse it with a function that returns true if the timer is counting up, toward the limit you set, and false if counting down to zero.

"done" seems very clear, while avoiding that potential confusion.

> I think a pause() and resume() function would be useful as well, as this would allow you to actually pause and resume timers (like for when a game is paused).

Well, it may be better to just modify Timer to pause internally when the Scene is paused, so that developers don't have to do that themselves.

But yes, we should still have pause() and resume(), because there are cases where the developer would want to manage that directly, for other purposes.

SolarLune:
"action.time()" still doesn't make sense. It's the time for the action... What?

I don't see how or why anybody could get "timeUp()" confused; there's no functionality for the timer to run in any other direction than forward. Even if there were, it would still be a value increasing or decreasing toward the target value; why would it matter if it were counting up to a desired time or down to zero? It's going to "go off" in a specified amount of time.

Pausing and resuming would be nice to add. I imagine pausing timers internally for paused scenes would be a good idea.

GoranM:
> "action.time()" still doesn't make sense. It's the time for the action... What?

Did you ever hear of "lunch time"? How about "[morphin time](https://www.youtube.com/watch?v=MalLny7MT6E)"?

It's pretty common terminology, really: "x time" -> "time to do x".

> there's no functionality for the timer to run in any other direction than forward. Even if there were, it would still be a value increasing or decreasing toward the target value

With "done", you don't have to know any of that.

> I imagine pausing timers internally for paused scenes would be a good idea.

Yes, but I'm not sure on how that should be implemented.

Right now, I'm thinking about something like `scene.timer()` to create a new Timer instance, so that each scene could grab and manipulate timers instanced within it ... It feels clunky.

GoranM:
Given the new Timer class, I think it's ok to close this.

PR #67, opened 2015-01-18
Access to Blender Game Properties through GameObject.props

rafcolson:
I'll update the API as soon as it's merged. Thanks for your help on this, Goran.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -395,6 +395,7 @@ def srl_objects(objects):
 
         name_object[obj.name] = {
             "type": obj.type,
+            "properties": {n: p.value for n, p in obj.game.properties.items()},
             "transform": transform,
             "parent": obj.parent.name if obj.parent else None,
             "model": mesh_name,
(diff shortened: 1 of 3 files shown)


GoranM:
Looks good.

Thanks!

rafcolson:
Unfortunately there seems to be an issue with my build. I don't know if I'm doing something wrong, but when I'm building an *.apk I'm getting an exception. With v0.0.76 everything works fine, so it must have something to do with the commits regarding GameObject.props.

GoranM:
There's nothing in your commits that could possibly break android builds - I just tested it, and it works fine on my system.

Did you remember to recompile bdx by running ant? Did you create a brand new project to test?

If you answered yes to questions above, try doing this for BDX:

```
ant clean
ant
```

And then for the project:

```
gradlew clean
gradlew android:assembleDebug
```

SolarLune:
@rafcolson - Is this working? No matter what and where I put it, props.has("property name") always makes BDX crash with a null pointer exception...? Printing out the value of "props" just returns null.

rafcolson:
>There's nothing in your commits that could possibly break android builds - I just tested it, and it works fine on my system.

That's reassuring. Thanks for the pointers, I'll try if that helps.

GoranM:
It would help if you could describe your overall development setup, and the step by step procedure you follow to test your BDX changes.

PR #74, opened 2015-01-19
Added subimage() getter and setter to SpriteAnim.

SolarLune:
Added playHead() getter and setter to be able to modify the playHead of the current animation.
Cleaned up imports a bit.

### DIFF
--- a/src/com/nilunder/bdx/components/SpriteAnim.java
+++ b/src/com/nilunder/bdx/components/SpriteAnim.java
@@ -49,7 +49,7 @@ public class SpriteAnim extends Component {
 			playHead = 0;
 		}
 
-		public void playHead(int subimage){playHead = subimage;}
+		public void playHead(int frame){playHead = frame;}
 
 		public int playHead(){return playHead;}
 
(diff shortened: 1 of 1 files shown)


GoranM:
I'm not sure about "subimage" as a name - I mean, you're changing the frame, so it would make more sense to have methods like:

```java
public int frame();
public void frame(int frame);
```

Wouldn't it?

SolarLune:
There's already a frame() function in SpriteAnim.java. To avoid confusion, I think it might be best to avoid naming it "frame()" (or name the frame() function something else).

GoranM:
The currently existing "frame" functions are private, and deal exclusively with Vectors. Even internally, type context makes the distinction quite clear (not to mention that it's enforced by the compiler); it's a complete non-issue externally, because users can't access private functions.

SolarLune:
OK, I would change it, then, but we can't have two frame() getters.

GoranM:
Oh, I just assumed that Java could distinguish between frame() - int and frame() - Vector2f, but I guess not ...

Anyway: It would probably be best to rename private frame methods to uvFrame, so we can just use frame for both public methods.

SolarLune:
Done!

GoranM:
Ok, looks good.

Thanks.

PR #77, opened 2015-01-20
Bug fix: createproj.set_android_sdk_version

rafcolson:
As I told you before, I have problems with building, but this fix shouldn't be having anything to do with that. Thanks for the tips.


GoranM:
I don't remember exactly, but I think there was a case where those directories didn't exist (SDK downloaded partially, or something like that), so yea, I set some default, but I think we could just as well do nothing in that case (although, I'm not sure).

Anyway:

Your "url" semantics seem unclear. I think 99% of people would look at that and think that maybe you're getting stuff via http, somehow, or something like that. Why not simply use "dir" for "directory"? I think that would make sense to most people (unlike "url").

The arguments of get_version could be a little more clear. Instead of just "k" for the function, you should name it something relevant, like dir_to_intv, because you're basically converting a directory name to an integer that represents the version. Also, the v prefix seems redundant for v_default - The name of the function implies that it's the version.

In set_version, what does 'p' stand for? It's not apparent from context, so maybe the name should be expanded a bit, to be more descriptive.

I think latest_build_tools_version should just be build_tools_version - You're getting whatever get_version provides, which could be latest, or default, or just the only available.

latest_platform_version should be compile_sdk_version - The ultimate goal is to set the compileSdkVersion in the gradle file, so I'm pretty sure that would make more sense.

Do set_android_sdk_build_tools_version and set_android_sdk_compile_sdk_version really need to exist as functions? It seems somewhat redundant, because you could just convert them into comments, and merge their internals into the main body of the function, without loosing anything on clarity (it would also be less code).

rafcolson:
Thanks, I'll make the changes. Is it okay to paste the code here? Or should I pull another request?

SolarLune:
@rafcolson - If you push to the same branch that you have on your fork, the changes will be applied to the same pull request automatically.

rafcolson:
Oh, so I should make a new branch of my master. Hmm, still not used to GitHub. A side-question, is there an easy way to revert a commit on your own branch? Thanks, SolarLune.

GoranM:
Your master seems up to date with my master (your commit, on top of current GoranM/bdx state), so I think you could just make the relevant changes, save the file, and do:

```
git commit --amend -am "Bug #68: Set compileSdkVersion for android."
```

Which will basically change the last commit, to include your new changes.

Then you just need to force push to your fork on github:

```
git push -f origin master
```

And I think that should do it (I assume you're working with git on the command line ... if you're using some other client, you can probably find documentation that explains how to do the same).

rafcolson:
I'm sorry for all the mess. It's terrible. But I think now I now how to do it.

First I'll pull a request to my master, but DON'T merge. Then the commit is available to pull a request to your master.

That avoids my master to get corrupted if my commits are no good. So I'm closing.. but I still have to investigate more how git works. Sorry for slowing you down.

Issue #68, opened 2015-01-18
Exception in build.gradle while building for android


rafcolson:
This bug was present from the beginning. I had to work around it by editing **build.gradle** in the android folder, line 3:
```java
compileSdkVersion 19
```
(When installing the add-on?) *buildToolsVersion* is being updated to my current Android SDK Build-tools version (see Android SDK Manager), but *compileSdkVersion* is not. I should have:
```java
android {
    buildToolsVersion "21.1.2"
    compileSdkVersion 21
    ...
```
Then it works.

Maybe this is an issue with gradle itself, but I felt it should be reported anyway.

For anyone who wants to build their BDX project into an Android *.apk, it should be helpful to inform about this issue. I could make a tutorial of all necessary steps to build for Android, but who knows, maybe it could be fixed..

GoranM:
This is one of those things that's not managed by the libgdx project generator, or gradle, so we need code that can find out, and subsequently set the proper version in the gradle file.

I already [do this](https://github.com/GoranM/bdx/blob/master/blender/bdx/ops/createproj.py#L210) for the build tools version, so we should probably do it for the sdk version as well.

You want to give that a try?

I think you can just search the platform directory in the android sdk, which should contain directories like android-\<version\>, so if you can just get the highest there, and set that in the build file, I think that should do it.

rafcolson:
> You want to give that a try?

I will. Thanks for the tips.

rafcolson:
Pulled request #77

GoranM:
You made a pull request, it's up to me to pull it.

Also, in commits that address issues, you should reference the issue in the commit message - don't reference the pull request from the issue.

rafcolson:
I'll keep in mind what you said.

rafcolson:
Well, it finally got in. Thanks.

PR #79, opened 2015-01-20
Bug #68: Set compileSdkVersion for android.

rafcolson:


GoranM:
This looks better, but I think that `get_version` should be renamed to `get_version_dir_name`, because that's what it actually does.

Anyway, to make the changes, I think you can just follow the procedure I outlined in your last PR. Just change `git push -f origin master` to `git push -f origin compileSdkVersion`.

rafcolson:
I finally found the Git Shell! I thought I needed Hub, ... I'm a mess. Anyway, again, I couldn't do exactly as you said because I pulled a request from a branch and I don't know how to clone that to my GitHub. So I merged it with my Master, adjusted the java script and with the Shell I followed your instructions.

Next time better, I suppose...

PR #85, opened 2015-01-21
Bug #68: Set compileSdkVersion for android.

rafcolson:


GoranM:
The old pull request from the original createSdkVersion branch didn't seem to update. Maybe you deleted, and then recreated the createSdkVersion branch?

In either case, this is fine, so I'll merge it.

PS: 

Good info: https://github.com/GoranM/bdx/wiki/Contributing#notable-resources

Issue #66, opened 2015-01-18
GameObject scale() and dimensions() not working for empties


rafcolson:
I'm not sure if this is supposed to be or not, but it sure would be nice these attributes worked with empties also.

Sometimes I use an empty for spawning random objects. The scale or dimensions of the empty is an easy reference for the size of the spawning area.

GoranM:
Well, technically, not a bug, since this is [something I did knowingly](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/GameObject.java#L295), but it should be made to work in the future, so maybe it's a good idea to treat it as a bug.

The easiest way to make it work is to make every object have a modelInstance, even if they don't have a visual representation (like empties and lights). There's already a defaultModel, so we could just create an instance from that, and set the object to be invisible.

That would also simplify other bits of code.

rafcolson:
Wouldn't that add more weight to the number of GL calls? I hope you can fix the performance issue with GL calls on Android. Then that won't be a disadvantage.

GoranM:
As far as I know, draw call overhead depends on which objects are [actually rendered](https://github.com/GoranM/bdx/blob/master/blender/bdx/gen/BdxApp.java#L59), not the number of existing model instances, so as long as we set the relevant objects to invisible, I think it should be fine.

For performance in general: I first need to confirm that rendering is the bottleneck. I mean, it's true that rendering is generally slow on mobile platforms, but it's possible that BDX is taking up a significant number of cycles, and that things could be much faster if optimized (in the right places).

Right now, BDX is "touching" all objects, even when not necessary:

The bullet body transform matrix is converted - [for every visible object, on every frame](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Scene.java#L474) - into the proper modelInstance transform matrix. If we were to move that logic into GameObject.transform, we could probably avoid a lot of that overhead, because that conversion would only happen when the object is actually moved.

Objects which don't have code in main, or running components, shouldn't have any additional logic overhead, but that's not quite true, because all of them are still being ["visited" by BDX](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Scene.java#L492), to actually call the main function, and that takes a bit of processing time, even if the function is empty. If we create (and manage) a list of objects that actually need to be processed, instead of iterating through scene.objects (which contains objects that don't), we could probably avoid some overhead.

[...]

PR #90, opened 2015-01-23
modelInstance for all objects

rafcolson:


rafcolson:
Forgot to put the Issue in the title, sorry.

#66

GoranM:
You should be able to use --amend to change the commit, along with the commit message (as I explained in the past).

The commit itself looks ok, but you would also need to search the codebase for
`modelInstance != null` and `modelInstance == null`, because now those conditionals are probably redundant, and should therefore be removed as a part of this PR.

PS: Don't recreate the branch, as I suspect you did for the last PR. Simply do the following:

```
git checkout modelInstance
```

Edit code, save files, then:

```
git commit --amend -am "modelInstance for all objects (#66)"
git push -f origin modelInstance
```

That should change this PR, so there should be no need to create a new one.

rafcolson:
It should be okay now.

GoranM:
Did you search the entire code-base for those patterns?

There's still a redundant check in GameObject.scale

rafcolson:
Sloppy of me. Done.

PR #94, opened 2015-01-24
Added valid boolean to GameObjects to check if they're valid.

SolarLune:
GameObjects now have a private field named `valid`, which gets set to true on creation and set to false on end()-ing.
Added valid() function check to check if a GameObject is valid / present in the game world.
Added parent checks to end() functions to remove GameObjects from parents' children lists if they're removed.

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -46,6 +46,7 @@ public class GameObject implements Named{
 	private GameObject parent;
 	private Matrix4f localTransform;
 	private Vector3f localScale;
+	private boolean valid;
 	
 	
 	public GameObject() {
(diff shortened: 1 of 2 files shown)


GoranM:
Instead of having the same code in `end` and `endNoChildren`, I think it would be better to just put it into scene.remove.

Also, note that calling `obj.parent(null)` will properly "unparent" the object in question ([as stated in the API reference](https://github.com/GoranM/bdx/wiki/API-GameObject#parent-p)), so there's no need to manipulate the parent's children list directly.

SolarLune:
Putting it into scene.remove sounds OK, but valid is currently a private field of GameObject, so it'd have to either be public (and so there wouldn't really be a need to have a valid() function), or there'd have to be a valid() setter (which doesn't make sense as it's not something that should be used by other objects or even the GameObject directly).

OK, obj.parent(null) can be used.

GoranM:
In the API reference, we point out that "Unless otherwise labeled, all fields should be considered read-only", so you can just make `valid` a public field, and developers can use that directly.

SolarLune:
PR Updated.

PR #95, opened 2015-01-24
Updated ArrayListNamed's int get() function.

SolarLune:
It now mirrors the String get() function in that if you request a value from the list outside of the scope of the list, it will return null rather than crash BDX.


GoranM:
I'm not sure about the overall utility of this addition.

I mean, I can't imagine one scenario where I would want get(int) to return null when out of bounds: I would actually prefer the crash, because indexing directly into the array implies clear expectations about capacity, which, if not met, probably constitute a bug.

SolarLune:
That's true, but I just thought it'd be good to have it be consistent. You're probably right, though. Closed.

Issue #93, opened 2015-01-23
init() and create() functions for GameObjects.


SolarLune:
Hey, there. So, I think both "init()" and "create()" functions might be useful for GameObjects. The main reason this would be a good idea, I think, is that BDX can, on scene creation, run all objects' init()  functions, and *then* run their create() functions. This would allow you to set up relationships, spawn scenes in, and otherwise set up the game from objects (in the init() function) and then be sure that they're actually usable before proceeding with them (in the create() function).

The main reason I'm asking for this is because I spawn a scene in from a game object's init() function, but I need to use an object in that spawned scene elsewhere. I could check to see if the scene exists in the main() function of the object that's going to access it, but that's a bit awkward. 

I suppose some level of "execution priority" would be really useful, alternatively; perhaps it could be implemented as a static variable in the GameObject's class (so just defining your custom class, extending it from GameObject, and then setting its executionPriority in the class itself would allow BDX to check it to make it run first, or last, or whatever)?

GoranM:
One could probably just change [this line](https://github.com/GoranM/bdx/blob/master/blender/bdx/exporter.py#L609), to:

```python
"objects": srl_objects(sorted(objects, key=lambda o: o.name)),
```

And then, in BDX, initial execution order should be based on python's sort order for object names (strings).

So, assuming that all your object names start with regular letters, you could just prefix the name of the "priority" object with a number, and that should produce desired results.

... Well, to avoid potential confusion, it would probably be better to write a bit of extra code, and consider only names that start with numbers:

```python
priority_objects = [o for o in objects if o.name[0].isdigit()]
other_objects = [o for o in objects if o not in priority_objects]
objects = sorted(priority_objects, key=lambda o: o.name) + other_objects
```

SolarLune:
It would seem that .java files and Java classes can't start with numbers, so that won't work. It's kind of a hack, but initial execution order based on Python's sort order for object names should work (so I guess you'd name whatever object you wanted to start first "AMyObject" or "AAMyObject" or whatever). 

I went with the first option to just alter the srl_objects line, but that didn't work because srl_objects uses a dictionary. So, I turned it into a SortedDictionary, but that still doesn't work. Looking into the issue currently.

GoranM:
> It would seem that .java files and Java classes can't start with numbers, so that won't work.
> ...
> Looks like it might have to do with templates being a HashMap (so it doesn't preserve the order that objects are inserted into it).

[...]

PR #99, opened 2015-01-25
Added ability to make an object initialize first.

SolarLune:
You can do this by putting "init_" in the name of that object.

### DIFF
--- a/src/com/nilunder/bdx/Scene.java
+++ b/src/com/nilunder/bdx/Scene.java
@@ -1,6 +1,7 @@
 package com.nilunder.bdx;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 
@@ -239,7 +240,19 @@ public class Scene implements Named{
 	}
 
 	private void addInstances(){
-		for (GameObject gobj : templates.values()){
+
+		ArrayList<GameObject> temps = new ArrayList<GameObject>(templates.values());
+
+		for (GameObject t : temps){
+
+			if (t.name.contains("init_")){
+
+				Collections.swap(temps, 0, temps.indexOf(t));
+
+			}
+		}
+
+		for (GameObject gobj : temps){
 			boolean onActiveLayer = gobj._json.get("active").asBoolean();
 			if (onActiveLayer && gobj.parent() == null){
 				GameObject g = clone(gobj);
--- a/src/com/nilunder/bdx/Scene.java
+++ b/src/com/nilunder/bdx/Scene.java
@@ -244,11 +244,9 @@ public class Scene implements Named{
 		ArrayList<GameObject> temps = new ArrayList<GameObject>(templates.values());
 
 		for (GameObject t : temps){
-
 			if (t.name.contains("init_")){
-
 				Collections.swap(temps, 0, temps.indexOf(t));
-
+				break;
 			}
 		}
 


rafcolson:
I don't know about this. It feels awkward. Couldn't we just add a priority flag like it is with the BGE. We'll probably have a script entry for the GameObject in the future anyway. I think this will make more sense to a lot of people. Having this dependency in the name... it just doesn't feel right. But that's only my opinion.

GoranM:
After the swap, you should break out of the loop; you're done at that point, so there's no need for subsequent iterations.

Also, I think it would be a good idea to remove blank lines from the for loop, to fit in with the overall style of the code-base, and to better imply that those directives implement a single higher-level operation/idea (placing the init object in front).

... Another small thing: When writing a PR that stems from a suggestion issue, it be would nice to reference that issue in the commit message, like:

`"Object with 'init_' prefix initializes first (#93)"`

SolarLune:
I don't much care for this solution either, to be honest. It does feel like a "non-definite" solution to a definite problem. Misspelling could still be an issue. Also, it doesn't make sense to add this here if we're going to add that custom BDX panel with script name specification and properties; a button to give objects execution priority would still be most desired.

GoranM:
> It does feel like a "non-definite" solution to a definite problem.

The definite problem is: One can't specify code that's guaranteed to run before objects initialize.

This solution solves that problem, as far as I can tell.

Could you elaborate a bit on what you mean exactly by "non-definite"?

> Misspelling could still be an issue

You could also assume that you checked the "initialize first" box in the panel, but you didn't, and your code doesn't run first, so you start wondering what's wrong - You check the panel, and find your error (just as you would with a misspelled prefix).

Or, since you now have one checkbox for each object, you mistakenly check the box for the wrong object.

The potential for human error is about the same.

> it doesn't make sense to add this here if we're going to add that custom BDX panel with script name specification and properties

Even with a panel, why have an additional option, and all the related code to support that?

There's no concrete benefit - It's just more code, in both the exporter and the engine API, for exactly the same functionality.

> execution priority

As I think we concluded in the referenced suggestion issue, there's no need for that: We just need the ability to specify an object that will initialize first.

SolarLune:
>The definite problem is: One can't specify code that's guaranteed to run before objects initialize. This solution solves that problem, as far as I can tell. Could you elaborate a bit on what you mean exactly by "non-definite"? ... The potential for human error is about the same.

Not quite. Changing the name of the object isn't as simple as just checking a box. You can misspell "init_"; you could capitalize it, you could add two i's or drop the _. This could easily be a problem if you're unfamiliar with the engine, and it's easy to misread "iinit_", "Init", or "init__" as "init_". It's not as simple as just checking a box to indicate this to the engine. To me, the primary thing is that the engine's easy-to-use; having a simple code-base should come second to that (though, obviously, it's also important).

This also provides a poor precedent in that naming objects can change their behavior, when naming objects should just be for the user's benefit and be the user's choice. The name shouldn't influence the behavior of an object at all; that's what the options in Blender are for. 

For example, it'd be kinda crazy to have to say "RB_ObjectName" to make your object recognized as a rigid body-enabled object in BDX, right? This is kind of a similar thing; it's kind of weird to have to name your object to indicate a property to BDX when there are a variety of other places where you have options and buttons to select and indicate other, perhaps even more important behaviors. It's inconsistent for no particular reason (to the user).

Ideally, there wouldn't need to be documentation for the engine, as everything's easy to understand and pretty much obvious. Naming your object to change its execution order isn't obvious - people would have to look it up to learn how this works. A button that has two states, a name, and a tool-tip that indicates this behavior is obvious - people probably wouldn't have to look it up to learn how it works.
 
> As I think we concluded in the referenced suggestion issue, there's no need for that: We just need the ability to specify an object that will initialize first.

Actually, being able to specify which object executes first could be useful in fast-moving games. For example, I have an InputMap class that I designed to handle input mapping for various controls. For it to work, I have to update it by running a poll() function. If I were designing a game that were really precise, like a fighting game, I think I'd really, really want the poll() function to happen first in every game frame.

I think it's also easier for users to understand an option that makes an object run first every frame than an option that makes an object initialize first and then drop that subsequently.

Maybe @rafcolson can weigh in on his thoughts.

rafcolson:
>The convention is easy to use, and it doesn't require us to code a specific panel, or to add more code to the public API.

I agree that it would be the easiest, quickest and probably simplest solution from a developer's point of view. But as a user myself, I would be more comfortable with a button next to the script field in Blender, if it were to be implemented.

>Not that I don't appreciate your feelings on the matter, but can you explain why this is not a good solution, or why alternatives would be better?

I hadn't much time today to explain why. I guess I just wanted to let you both know I don't think this would be the best solution so it wouldn't get merged. Now that I have a little time, I'll try to explain. Actually, the points are somewhat the same as those mentioned by SolarLune:

1. The ability to make an object run its script first belongs to the same category as the ability to change the script.
2. Users will immediately understand how it works.
3. Object naming as a means to define object properties is unconventional, regardless of what may be practiced. In my opinion, such practice should be a last resort.
4. I'm assuming that by default a script will be assigned to an object if it has the same name. So you would have to edit both names in order to give it priority. Compared to a simple mouse click ...

>For example, it'd be kinda crazy to have to say "RB_ObjectName" to make your object recognized as a rigid body-enabled object in BDX, right?

I think this is a good illustration of why "init_ObjectName" would be a bit strange.

In summary, I would think "the extra code" would be worth it. Look at it as a good investment. People will definitely use it. The question is, how would they prefer to use it.

GoranM:
> Changing the name of the object isn't as simple as just checking a box.

Assuming it's the right box, and that you actually clicked it ... I don't know - Maybe I'm just projecting from my own experience with misclicks, or clicking the wrong option for the wrong object, etc.

> To me, the primary thing is that the engine's easy-to-use; having a simple code-base should come second to that (though, obviously, it's also important).

Generally, I would agree, but there are cases where differences in ease-of-use are not large enough to justify the related delta in code complexity.

It's something that needs to be judged on a case-by-case basis, for the greatest overall benefit.

> This also provides a poor precedent in that naming objects can change their behavior, when naming objects should just be for the user's benefit and be the user's choice.

Well, technically, the "init_" prefix wouldn't change object behavior: All the code still runs in the same way, just as it would if the object didn't have that prefix.

The prefix is basically a hint for the system, to just initialize that object first.

> A button that has two states, a name, and a tool-tip that indicates this behavior is obvious - people probably wouldn't have to look it up to learn how it works.

That's a good point.

> Actually, being able to specify which object executes first could be useful in fast-moving games.

I think the existing solution would already ensure that the first initialized object is also the first that executes, because as long as it's first in the scene.objects list, it will be processed first in runObjectLogic.

Anyway:

I'm ok with adding the button/checkbox/whatever if/when we get a BDX object panel.

For now, it would probably make sense to merge this patch, because the implementation would essentially be the same on the engine side (instead of picking out an object with a specific prefix, we would have a specific name provided by the exporter, of the object with the "run first" checkbox ticked).

However, if you would rather wait until then, you can simply close this.

It's your choice.

rafcolson:
I also don't mind if this is merged as a temporarily implementation. So it's up to you, SolarLune.

SolarLune:
> After the swap, you should break out of the loop; you're done at that point, so there's no need for subsequent iterations.

Fixed.

> Also, I think it would be a good idea to remove blank lines from the for loop, to fit in with the overall style of the code-base, and to better imply that those directives implement a single higher-level operation/idea (placing the init object in front).

Done.

> I think the existing solution would already ensure that the first initialized object is also the first that executes, because as long as it's first in the scene.objects list, it will be processed first in runObjectLogic.

Oh, okay, cool. That should help out, then.

> ... Another small thing: When writing a PR that stems from a suggestion issue, it be would nice to reference that issue in the commit message, like: "Object with 'init_' prefix initializes first (#93)"

Done. OK, well, this can get merged in until the future BDX panel gets implemented. Thanks for your input, guys.

Issue #102, opened 2015-01-26
Incorrect children end()ing.


SolarLune:
Hey, there. Basically, there was an error dealing with looping through children and end() not working correctly as the children list would be changed when children were de-parented after being removed from the scene. Adding a PR to address this.

GoranM:
It would probably be a good idea to provide the actual error.

I assume it was a ConcurrentModificationException, but I shouldn't have to assume in a bug report.

SolarLune:
I didn't particularly feel the need to provide a lot of information given that I noticed the bug and provided a patch for it immediately.

In any case, that was the error, if I recall correctly. I'll keep it in mind in the future.

SolarLune:
Merged. Thanks, Goran.

PR #103, opened 2015-01-26
Addresses issue (#102). Creates a new list for GameObject.end().

SolarLune:
Creates a new list before looping through a GameObject's children to end them when using the end() function. Fixes broken end()ing functionality.

PR for issue (#102).

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -283,7 +283,7 @@ public class GameObject implements Named{
 	}
 	
 	public void end(){
-		for (GameObject g : children){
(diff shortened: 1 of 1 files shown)


GoranM:
The code is fine, but the description could be a bit better:

Instead of the "Addresses isssue" prefix, you should use the "Bug: " prefix, so that it's consistent with the convention that I and Raf already follow.

Also, I think it would be more concise to say that you "Iterate over copy of children in GameObject.end, to avoid ConcurrentModException", but that's just my opinion.

PS: Just in case if you don't know how, you can amend the existing commit with:

```
git commit --amend -m "New message"
```

That will essentially replace the latest commit, and then you can force push that to your github branch, which should make the changes visible here.

Issue #100, opened 2015-01-26
Properties erased by group instance


SolarLune:
Yo. It looks like an object's properties are completely erased by its group instance (if it belongs to one); shouldn't those properties be added to or overwritten by similarly named properties on group instances? For example:

On Guard, there's a property named "behavior", set to "stand". It also has a property named "health", set to 100. That makes it so that guards will, by default, stand there with 100 health (you could do this in code, of course, but this is an example).

But then, you decide that you want a specific guard to walk left. So, on a group instance of a Guard, you set the "behavior" property to "walk left"; I would then expect this instance of the Guard object to have a "behavior" property set to "walk left" *and* a "health" property set to 100, rather than just have a "behavior" property set to "walk left".

GoranM:
Raf's code doesn't merge properties.

I don't know if he would be interested in adding that support, but I guess you can assign this to him for now, and see what he says.

rafcolson:
I will look into it today.

rafcolson:
@SolarLune
This was something I overlooked totally. Right now properties are replaced completely, and that's not how it should be. See [here](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Scene.java#L311).

To merge both JsonValues I think I have to convert to HashMaps, or could there be another way doing this with JsonValues? I suggest to change GameObject.props to an HashMap instead then.

GoranM:
> I have to remember not to put references in the commit description until it is complete. It seems, that every time I change something to it, it will reference the issue also.

It's good to have a record of changes; shows how the solution evolved.

Don't remove the reference, just take the time to ensure that the commit is ready before you push to your github branch; GitHub only knows about the commits that you actually push.

rafcolson:
>Don't remove the reference, just take the time to ensure that the commit is ready before you push to your github branch; GitHub only knows about the commits that you actually push.

You're right. I shouldn't push every time I make a change.

GoranM:
It's up to the person who opened the bug (@SolarLune in this case) to confirm that it's resolved, and to ultimately close.

rafcolson:
Oh, I'm sorry.

SolarLune:
In my quick tests, it would appear that everything works as it should - object properties can be retrieved, and group instance properties of the same name overwrite the "original" object properties.

Thanks, @rafcolson. Issue's closed.

PR #104, opened 2015-01-26
Bug #100: Props support for group instance

rafcolson:
Note: Only properties sharing the same name will be overwritten.


rafcolson:
Let me know if there's a way to do this without converting to HashMap.

GoranM:
I don't know if this can all be done with JsonValue, instead of converting to HashMap (I kinda assumed that you would research that), but I guess a HashMap is fine.

You can cut a bit of code in your initialization process:

```java
g.props = new HashMap<String, JsonValue>();

for (JsonValue prop : gobj.get("properties")){
    props.put(prop.name, prop);
}
```

JsonValue supports Java iteration conventions, so that should work fine.

However, looking at this line:

```java
g.props.putAll(inst.props);
```

In this context, `g.props` references the original HashMap that was created for the template object, as you can see [here](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Scene.java#L284), so wouldn't that make props for all instances exactly the same?

rafcolson:
>I don't know if this can all be done with JsonValue, instead of converting to HashMap (I kinda assumed that you would research that), but I guess a HashMap is fine.

Unfortunately my research doesn't give the best results always, especially with Java, because I'm just learning it. Another reason definitely is that I'm searching for answers in the API's's and Stackoverflow. It's just not enough.

>JsonValue supports Java iteration conventions, so that should work fine.

See! I didn't know that.

>However, looking at this line:
```java
g.props.putAll(inst.props);
```
In this context, `g.props` references the original HashMap that was created for the template object, as you can see [here](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Scene.java#L284), so wouldn't that make props for all instances exactly the same?

Essentially, yes. An added instance (e.g. ```cube_inst = scene.add("Cube")```) will inherit all properties of the template. A group instance will too, but will preserve all properties that may be added to it.

How do you think it should be then, if I may ask?

rafcolson:
I don't understand why those merges pop up here, it's a bit frustrating. I'm closing this and reopening a new PR to avoid confusing.

Issue #108, opened 2015-01-27
least painful way to disable world gravity?


Nona2:
I basically want to script gravity for each character/state separately.  I turn off world gravity in Blender when using it alone, was wondering how I could do this in BDX, thanks in advance

edit: since it was right in front of me... also horizon color

SolarLune:
`scene.world.setGravity()`; not sure about how to set the horizon / clear / "empty space" color. I don't think BDX supports it "out of the box". BDX is built on LibGDX, though, so it should be possible. Looking into how to do this in LibGDX would probably be your best course of action until BDX supports the option, I think.

rafcolson:
I think adding support for this wouldn't be a bad idea. It wouldn't be hard to do either. I'll look into it.

Nona2:
thosell work thx...

PR #111, opened 2015-01-27
Support for Blender physics_gravity (#108)

rafcolson:


rafcolson:
Closing because I accidentally pulled two commits instead of one.

PR #113, opened 2015-01-27
Bug #100: Props support for group instance

rafcolson:


GoranM:
> An added instance (e.g. cube_inst = scene.add("Cube")) will inherit all properties of the template.

In the system, as it works now, "inherit" means: "set my props reference equal to the template props reference". Both references point to the same exact HashMap in memory.

> A group instance will too, but will preserve all properties that may be added to it.

Except that properties added to it will now also be available on `cube_inst`, and any other instance of "Cube" that was previously added. All added properties will be present on all other instances, because everything is referencing a single HashMap structure on the template.

When you call, putAll, you're adding to a single HashMap that all instances share.

> How do you think it should be then, if I may ask?

Well, I can't speak for @SolarLune, but I wouldn't want added/modified properties to show up on objects that shouldn't have them.

PS: Instead of merging upstream changes into your branch, you should rebase it.

Assuming you already updated master:

```
git checkout featureBranch
git rebase master
```

rafcolson:
>In the system, as it works now, "inherit" means: "set my props reference equal to the template props reference". Both references point to the same exact HashMap in memory.

I think I understand now. Made the adjustment. I hope it's okay now.

GoranM:
Well, that would certainly work, but you only need a new HashMap for objects that merge-in properties from a link group instance - It's perfectly fine for others to share a single HashMap, as far as I can tell.

rafcolson:
@GoranM Changed it.

Issue #105, opened 2015-01-27
Random.direction not normalized?


SolarLune:
Hey, there. It seems like Random.direction() isn't normalized; shouldn't it be?

GoranM:
I didn't feel it was necessary.

As far as I know, "direction" is not a concept that implies a specific magnitude. Also, one would typically want to set the length to something specific anyway, so normalizing it before-hand seemed like just wasting cycles.

SolarLune:
It seems like a direction would indeed imply a specific magnitude of 1 no matter what direction gets generated. If it's not normalized, then the length can vary widely, which contradicts the idea of generating a direction, since there's more data that varies - it's a random Vector, not a random direction (carried by a Vector). 

In addition, this opens up a position for bugs for users if they use the function expecting a vector of length 1 and then try using scale or mul instead of setting the length directly.

If you want both for efficiency, then this could be renamed to random.Vector, and random.Direction could return a normalized random.Vector.

rafcolson:
I would agree. However, I would just keep it at renaming to random.vector. And just add a method to Vector3f to normalize. So you could write: ```Vector3f v = random.vector().normalized()```

GoranM:
> If it's not normalized, then the length can vary widely, which contradicts the idea of generating a direction

I don't see how that would contradict the "idea of generating a direction": Vector(1, 1, 1) represents the same direction as Vector(99, 99, 99) - Direction is about where the arrow is pointing, not its length.

In any case: I don't mind much one way or another, so if you want Random.direction() to return a normalized vector, feel free to make that PR.

SolarLune:
It contradicts the idea because it's not generating a vector with random direction, it's generating a random vector. The difference is that you don't know what length the function will generate before-hand.

@rafcolson - There's already a normalized() function in Vector3f's. What do you think about having it return a normalized vector? Alternatively, @GoranM, what about renaming it random.Vector?

rafcolson:
@SolarLune Calling ```Random.vector()``` should return a random vector, not a random normalized vector, in my opinion. Having a method ```Random.vectorNormalized()``` seems a bit redundant if you could just call ```Random.vector().normalized()```.

GoranM:
> It contradicts the idea because it's not generating a vector with random direction, it's generating a random vector.

A random vector is in fact a vector with random direction.

[...]

PR #116, opened 2015-01-28
Issue #105 - Normalize random.direction().

SolarLune:
Change random.direction to random.vector.

### DIFF
--- a/src/com/nilunder/bdx/utils/Random.java
+++ b/src/com/nilunder/bdx/utils/Random.java
@@ -23,11 +23,17 @@ public class Random{
 	}
 
 	public static Vector3f direction(){
+		Vector3f vec = vector();
+		vec.normalize();
+		return vec;
+	}
+
+	public static Vector3f vector(){
 		ArrayList<Integer> ints = new ArrayList<Integer>();
 		ints.add(1);
 		ints.add(-1);
 		return new Vector3f(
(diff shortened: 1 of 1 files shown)


GoranM:
While length is (logically) irrelevant for Random.direction, restricting the length of Random.vector to 1 doesn't really make sense - It's not a random vector if no two components can be greater than 1/sqrt(2).

It's ok for Random.direction to have that restriction, but it wouldn't make sense for Random.vector.

To remove the last commit, you can do: `git reset --hard HEAD~` (on this branch, of course).

SolarLune:
OK, after some thought, I think I agree with you. Having two functions (one for a normalized vector (`direction`), and one for a random one (`vector`)) should be a happy medium.

GoranM:
Ok, just make it into a single commit, please:

```
git reset --soft HEAD~
git commit --amend -m "Add Random.vector(); direction() returns unit vector (#105)."
```

SolarLune:
I think that should do it.

GoranM:
Did you leave the `Issue #105 ` prefix in there on purpose?

One reference is sufficient, and for things like suggestions, I would prefer the postfix reference: `(#105)`.

Running the exact commit command I specified above should make the message proper.

SolarLune:
Ah, I didn't see that. My mistake.

Issue #115, opened 2015-01-28
Override GameObject.toString()


SolarLune:
Hey there. 

When you print out information about a collection of game objects (like the children list for GameObject), the console won't print out particularly useful information, but rather just a list of objects whose base class is "GameObject". 

It might be nice to override the GameObject's toString() function to print out its object name and hashstring (or even game object ID, if you wanted to create that field that increments with every added game boject and put that in there), rather than class name and hashstring to help you understand what objects are contained in what collections.

GoranM:
Instead of:

```
com.nilunder.bdx.GameObject@ffacb2
```

Something like:

```
Name<GameObject>@ffacb2
```

Would probably be better.

SolarLune:
Done; PR submitted.

SolarLune:
New PR submitted (rebasing didn't appear to solve the issue).

SolarLune:
PR's been merged. Thanks!

PR #119, opened 2015-01-28
(#115) - toString() returns object name, class name, and hashcode.

SolarLune:


GoranM:
Assuming you updated your master from bdx upstream (GoranM/bdx), you should have been able to do:

```
git checkout toString
git rebase master
```

Were there conflicts? If so, you're supposed to resolve those, of course, for changes to become visible in the commit.

Anyway, I'll merge this now - Just a small nitpick:

The reference convention we use for bugs is prefix:

```"Bug #666: rest of message."```

For anything else, postfix:

```"message (#667)"```

Issue #118, opened 2015-01-28
ghost toggle


Nona2:
In blender the only way i know how to toggle ghost is to parent it to a dumby object in the scene and set it to ghost.  Then to un ghost it i just unparent it.  Any chance of the ghosting being added to parent function or even better a toggling ghost function?

also if you are aware of a better way to toggle ghost in BGE that could be helpful too

rafcolson:
It's something on my most wanted list too. I think I can do it, so I'll give it a try.

@GoranM & @SolarLune : please say if I should stop assigning myself. I realize that I'm not that experienced and I don't like it when I'm slowing everything down.

SolarLune:
@rafcolson Psh, nah, man. You're doing great, and your solutions are solid. You're not slowing anything down.

GoranM:
You can't really slow anything down. I mean, if we know you're working in a particular area of the code base, we might keep that in mind, and try to stay out of your way (to help you avoid merge conflicts), but there's nothing about your work that could fundamentally prevent anyone from doing their work.

Worst case scenario: It takes you a bit more time, and you'll have to resolve some conflicts before your work can be merged.

In either case, you shouldn't rush: Quality > speed.

rafcolson:
I have a couple of suggestions with this. I would like to shorten the code to get ```meshName``` a bit. No need for ```mesh``` in [Scene.java#L177](https://github.com/rafcolson/bdx/blob/ghost/src/com/nilunder/bdx/Scene.java#L177).

I also like to rename keys [```"body"```](https://github.com/rafcolson/bdx/blob/ghost/blender/bdx/exporter.py#L413) and  [```"bounds"```](https://github.com/rafcolson/bdx/blob/ghost/blender/bdx/exporter.py#L414) to [```"body_type"```] and [```"bounds_type"```].

GoranM:
It all sounds good.

You can include those changes into the current PR for ghost, if you want. Or if you want to make a different PR with those commits, that would also be fine.

SolarLune:
@Nona2 - A ghost() function was merged to allow you to toggle ghosting for objects; it should be what you need, and enable you to close this issue.

Nona2:
thx guys

PR #122, opened 2015-01-29
GameObject.ghost() (#118)

rafcolson:


SolarLune:
Aww yissss

On a side-note, it's crazy to see how easy it is to expose features that Bullet supports but that the BGE doesn't (for whatever reason). It's fun to see others work on these kinds of features as well as work on them oneself.

Nice work, Raco.

rafcolson:
Thanks. Yes, there's a lot of potential in this. If I'm not mistaken, it's even possible to "change" objects from Dynamic to Static or No Collision.

Ah yes, I almost forgot to mention that I deliberately got "SENSOR" out of the picture in exporter.py because it is treated as a Static Ghost anyway. I removed an 'if' in method Bullet.setCollisionFlags().

Also, with objects that have No Collision it should be noted that ghost will be ```true```. Does it seem necessary to change this to ```null```?

GoranM:
> Yes, there's a lot of potential in this. If I'm not mistaken, it's even possible to "change" objects from Dynamic to Static or No Collision

Instead of ghost methods, it might be better to have bodyType methods - that would also remove the need for a ghost argument in Bullet.setCollisionFlags.

> I deliberately got "SENSOR" out of the picture in exporter.py because it is treated as a Static Ghost anyway. I removed an 'if' in method Bullet.setCollisionFlags().

The exporter should not "compress" options available in Blender, especially when there is no significant benefit in doing so; For now, we set the same collision flags for both sensor and ghost, but that may change.

In general, exporter output should aim to reflect data that's available; Which of that data we use, and how, is something to be codified in the engine.

rafcolson:
>Instead of ghost methods, it might be better to have bodyType methods - that would also remove the need for a ghost argument in Bullet.setCollisionFlags.

I'm not exactly sure what you mean by that. You mean to change bodyType to: "STATIC_GHOST", "DYNAMIC_GHOST" and "RIGID_BODY_GHOST" when ghost is enabled?

>For now, we set the same collision flags for both sensor and ghost, but that may change.

Yes, I thought this might be the case. So I'm changing it.

GoranM:
I mean, instead of a `GameObject.ghost(boolean ghost)` method, you can have a `GameObject.bodyType(String type)` method, which would basically just call `Bullet.setCollisionFlags(body, type)`.

rafcolson:
Yes, I just tried that. The problem is that it's a bit inconsistent unless we add the ability to change body types too. But you would choose not to add such an ability then using a method such as ```GameObject.bodyType(String type)``` for enabling/disabling ghost would be a bit messy.

What I mean is that users could think they can change the body type when it's not supported.

GoranM:
Oh, wait, "GHOST" is not a body type ... right, so it would probably be best to have ghost methods.

However, I don't think we would need any additional fields, because whether or not an object is  "GHOST" can be determined via `GameObject.body.getCollisionFlags()`, and setting an object to ghost can be done via GameObject.body.setCollisionFlags`.

rafcolson:
What do you mean by 'additional fields'? No ```GameObject.ghost()``` method you mean?

GoranM:
By fields, I mean variables.

You don't need a ghost boolean on a GameObject, and you don't need a bodyType string either (which also means that you don't need to write code to set them).

You can just have ghost methods like this:

```java
public boolean ghost(){
    return body.getCollisionFlags() & CollisionFlags.NO_CONTACT_RESPONSE;
}

public void ghost(boolean ghost){
    int flags = body.getCollisionFlags();
    int noContact = CollisionFlags.NO_CONTACT_RESPONSE;

    if (ghost)
        flags |= noContact;
    else
        flags &= ~noContact;

    body.setCollisionFlags(flags);
}
```

I think that will set the flags properly (assuming I used the right bitwise operation to add or remove the no contact flag from existing flags).

rafcolson:
Much, much better that way. Thanks. (I seem a bit useless, but hey, I'm learning a lot).

PR #123, opened 2015-01-30
GameObject.ghost() (#118)

rafcolson:


GoranM:
It seems like my in-code post doesn't add to the post count for this thread, even though it's shown here ...

rafcolson:
Made the change.

GoranM:
> however, now ghost() returns false even when it's true.

It should work (it works in my tests) ... Make sure to use this exact code:

```java
public boolean ghost(){
	int noContact = body.getCollisionFlags() & CollisionFlags.NO_CONTACT_RESPONSE;
	return noContact != 0 ? true : false;
}
```

Also, make sure that your bdx build is recompiled, and that core/libs (in your test project) are updated.

Using '==' works for now because there's a bug in my code: A static object set to ghost will only have the NO_CONTACT_RESPONSE flag, but it should have KINEMATIC_OBJECT and NO_CONTACT_RESPONSE.

When that bug is fixed, the '==' comparison will fail to return correct results for static objects set to ghost, because:

```java
valueOfFlags = KINEMATIC_OBJECT | NO_CONTACT_RESPONSE;
valueOfFlags != NO_CONTACT_RESPONSE;
```

The bitwise test should work in either case.

rafcolson:
Okay, I understand. Any chance this bug is in Bullet.java? Should I fix it?

GoranM:
So now it works in your tests?

> Any chance this bug is in Bullet.java?

Yes.

> Should I fix it?

If you want to, but it would probably be better to make a new PR for that - this one should just be for ghost.

Issue #125, opened 2015-01-30
Don't run main() for invalid objects


SolarLune:
Hey, there. So I'm trying to debug a behavior that appears every so often with my game, and I think I might have it down. It has to do with the order of execution - I end an object remotely, but it still gets its main() function run, thereby causing the problem I'm seeing. 

Anyway, it appears to be possible for a game object's main() function to be run even though it's been ended. The object remains in the objects list, and the objects in the `toBeRemoved` list only get actually removed after being updated. 

I think a possible solution would be to:

* Remove objects immediately instead of doing it in a group with the `toBeRemoved` list (probably slower), or
* Just not run main() on objects that have been invalidated (seems like a simple and easy solution).

rafcolson:
> The object remains in the objects list

This is unwanted behavior. What I don't understand is why objects aren't removed immediately. Is this slowing down everything else so much?

SolarLune:
I dunno. Removing an object from a list might be a performance concern, since the list has to shuffle things around depending on the object's position in the list. I guess that concern would become greater with more frequent end() calls and a larger objects list (which can, as I'm sure you know, easily grow into the hundreds or even thousands in size).

GoranM:
> why objects aren't removed immediately

Because that would throw a ConcurrentModificationException; We iterate through scene.objects to run the logic, so we can't directly alter scene.objects during that time, because that would invalidate the iteration.

> Removing an object from a list might be a performance concern, since the list has to shuffle things around depending on the removed object's position in the list.

Scene.objects is a LinkedList, so removing any given GameObject would simply set the "next" reference on the previous node, to point at the node that should come next:

```
A -> B -> C -> ...
// remove B
A ------> C -> ...
```

No need to shuffle things around: removal is an O(1) operation.

Anyway, we could put this in the loop that runs the logic:

```java
if (g.valid())
    continue;
```

It will cost some performance (every branch does), but that should solve the problem at hand.

SolarLune:
Ah, I see; I suppose that does make sense. 

PR submitted.

SolarLune:
PR merged. Thanks. Issue's closed.

PR #126, opened 2015-01-30
valid() check to prevent logic execution on invalidated objects (#125).

SolarLune:


GoranM:
Your indentation is a bit off - It may not be visible in your editor, but it's obvious if you view the commit source here, on github.

Also, no need for that comment; The code itself is pretty simple/clear.

SolarLune:
Fixed.

GoranM:
The unnecessary comment is still there.

The commit message is not sufficiently descriptive (the one you used before was fine).

SolarLune:
I don't know what you mean "the unnecessary comment", then. 

Oh, you mean the actual comment in the code. I'll fix it in awhile.

SolarLune:
Fixed.

Issue #130, opened 2015-01-31
rayMultiple()


SolarLune:
Hey, a function that returns a list of all objects intersected by a ray would be nice. 

I'd guess it would shoot a ray from an initial position toward the ending position, and on collision with an object, do it again from that new position toward the ending position. It would add the collided object to a list of objects that it collided with in case of multiple collisions. When there are no collisions or the only collisions are in the list of collided objects, the function's completed.

It would return that list of collided objects. This list would naturally be in the order of distance from the casting point outwards, so you could get the closest collision easily by just using get(0) or first().

Thoughts?

GoranM:
xray() - ArrayList\<RayHit\>

That would fit well, I think.

SolarLune:
OK, PR submitted. An interesting thing to note is that triangle mesh objects trigger collisions on both sides of their faces, while the other collision types only trigger a collision on the "outside faces". So a triangle mesh sphere, for example, will report two collisions by default.

Anyway, this behavior is ignored in the xray() function, as only one collision from each object is returned. I wonder if it's good to ignore such a behavior - perhaps you'd want to do something on both sides of a ray cast, like put a bullet hole on the entry and exit holes of wall, for example?

GoranM:
We could simply include all of it. I mean, xray() would return an ArrayList of RayHit objects, more than one of which could reference the same GameObject, and we would simply document that.

Or, I guess we could have an additional xray(boolean includeExit), which would include everything, while xray() would filter to have only one RayHit for one object.

I would be fine with either.

SolarLune:
Sounds good.

I've updated it in the opposite way to re-use as much code as possible; the normal xray() function has an additional optional argument to specify if you want all objects or just one RayHit per object; the short xray() function defaults to that one RayHit per object.

GoranM:
> Do you think the additional xray() function is really useful / necessary?

One less argument that people have to remember/type, for not too much additional code in the engine.

It's a net positive, in my view.

SolarLune:
OK. I've updated it again to fix the naming problems you mentioned.

SolarLune:
PR Merged. Thanks.

PR #132, opened 2015-01-31
xray() function to cast a ray through multiple objects (#130).

SolarLune:


GoranM:
It might be better to keep the semantics consistent with `RayHit`, so instead of `collisions`, `hits` would fit better, and `collidedObjects` could be renamed to `hitObjects`.

GoranM:
What about the diff? (note the edit above)

SolarLune:
Fixed.

Issue #76, opened 2015-01-20
Profiling


rafcolson:
>The easiest test you can run is to actually time the relevant functions, and see how long they take to process, relative to everything else: scene.update() + renderScene().
>
>You should be able to use com.badlogic.gdx.utils.TimeUtils:
>
>```java
long t = TimeUtils.millis();
functionYouWantToTime();
long tt = TimeUtils.millis();
long delta = tt - t;
```
>
>If there is truly a significant performance benefit, I would love to know, of course, but I think it's something we should discuss in a different issue (we're quite off-topic in this one).

I'll be using this issue for relevant information about profiling. Anyone is welcome to join in.

SolarLune:
That looks cool; I'd like to see a full-on profiling info function or dictionary to assist with finding optimizations. See [this](http://blenderartists.org/forum/showthread.php?355758-BDX-alpha-release-Testers-wanted&p=2801956&viewfull=1#post2801956) page of the BDX BlenderArtists thread.

rafcolson:
I assigned myself, but let this not keep you from joining in. My idea and implementation of a profiler probably won't be exactly what we need, but I'll pull some requests anyway. This way we could discuss it further.

rafcolson:
I've been working on [profiler-scene](https://github.com/rafcolson/bdx/tree/profiler-scene). Everything should work, but no objects are displayed. I think I'm close, though. Any help is appreciated.

I'm not very sure what sizes I should use to instantiate the Camera, is it resolution or units?

rafcolson:
@GoranM Could you please have a look at the API, if my additions regarding the Profiler are correct? Thanks.

SolarLune:
If the profiler runs by default and provides information on each aspect of the game, then it might be a good idea to put that somewhere on those pages. Maybe under the main heading of the Profiler, a bit could be written concerning how to use it.

For example, to just call percents() to get the hash map of each "aspect" of the engine running and the corresponding percentages, what each profiled stat corresponds to, what a is good spread, so to speak (i.e. if you have more on `Outside`, it means your game isn't using all of the computer's resources, and so is running at an excess, which is good), and so forth.

GoranM:
I'm not sure if there should be a link to the Profiler page under com.nilunder.bdx.utils: It's not really a class for the user to manage, or use directly - It's basically a system object with a few relevant methods that developers can actually use (start, stop, percents).

[...]

PR #134, opened 2015-02-01
Profiler test (#76)

rafcolson:
This is a simple profiler. It evaluates three processes in BdxApp, but to be honest, to me it doesn't say much. "scenes" takes all the time in comparison with "graphics" and "input". But it's a concept. Let me know what you think about it.

Edit:

You can get the profiler percentages (HashMap\<String, Float\>) like so:
```
System.out.println(Bdx.profiler.pcts());
```


SolarLune:
Cool. You would probably do well to split up the "scenes" section into several pieces, as that's logic execution, rendering, and transformation application.

rafcolson:
Yes, that's more practical. Updated.

SolarLune:
^ That could work well. I don't think there's any built-in method to draw text and bars apart from doing it in pure OpenGL, I suppose, so a profiler .bdx file could be packaged with BDX and, of course, be placed in the scenes folder. Consisting of text and bars. If you have a certain profiler variable on, then it would add in that scene (and, I suppose, not put it in with the scenes list to keep you from accidentally removing it during gameplay).

The actual updating of the profiler info could be done with a single source file that handles the scaling of those bars and text output on those text objects to represent the profiler data. It might be possible to hide that file by naming it with a "." at the beginning, too.

The alternative would be to do it in OpenGL, I guess; not too sure how difficult and how much extra code that would be to implement.

rafcolson:
With the overlay scene approach the Profiler scene should always be on top too.

GoranM:
Instead of calling static methods directly on the class, you should just create an instance of the Profiler on `Bdx.profiler`. It would be more consistent with other, globally available objects.

The "process" / "subProcess" semantics are confusing, because they already reference existing programming concepts (on first look, some programmers may think you're talking about code that runs in a different address space). Just use `startTime` and `stopTime` and simple/short names like that; for overall profiler times, use profStartTime and profStopTime. I think that would be less confusing, and less visual clutter.

Having to include all the code from one time call to another is a pretty big limitation; That should be redesigned into something a little more flexible, so we can do this:

```java
Profiler p = Bdx.profiler;

p.start("something");
// code
long millis = p.stop("something");
System.out.println(millis); // if I just want time for one run, instead of add on loop

// code I want to skip

p.start("blah");
// blah code
p.stop("blah");
// after blah
p.stop("after");
// more
p.stop("more");
```

rafcolson:
I 'll redesign it.

rafcolson:
It finally works, so you can have a look. In the meantime I'll start with constructing the Profiler overlay scene.

Is ```Bdx.profiler.pcts()``` (percentages) descriptive enough?

SolarLune:
`Bdx.profiler.percents()` or `Bdx.profiler.percentages()` would probably be fine; `pcts()` doesn't really sound like "Percentages" or "percents" to me. Looking great, Raco!

rafcolson:
Had to start all over because my master's remote's were all messed up.

Issue #136, opened 2015-02-01
scene.gravity


SolarLune:
It'd be nice to get and set the gravity of the scene through scene.gravity() functions.

SolarLune:
@rafcolson I wonder, is it necessary to have an internal gravity private property? You could just set and get the gravity directly. Thanks for the addition, though!

rafcolson:
Committed.

rafcolson:
@SolarLune This probably can be closed since it was committed. Just reminding.

SolarLune:
Ah, yes. Thanks.

PR #137, opened 2015-02-01
Scene.gravity (#136)

rafcolson:
I'm not sure how I could use ```world.getGravity(Vector3f out)``` instead of storing the variable, but any suggestions are welcome.


SolarLune:
```
Vector3f grav = new Vector3f();
world.getGravity(grav);
return grav;
```

Should work.

rafcolson:
Merged update.

PR #141, opened 2015-02-02
Profiler (#76)

rafcolson:
Edit:

This is a simple profiler. You can get the profiler percentages (LinkedHashMap\<String, Float\>) like so:
```
System.out.println(Bdx.profiler.percents());
```

### DIFF
--- a/blender/bdx/gen/BdxApp.java
+++ b/blender/bdx/gen/BdxApp.java
@@ -40,15 +40,21 @@ public class BdxApp implements ApplicationListener {
 
 	@Override
 	public void render() {
+		Bdx.profiler.start();
 		Gdx.gl.glViewport(0, 0, Gdx.graphics.getWidth(), Gdx.graphics.getHeight());
 		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
+		Bdx.profiler.stop("graphics");
 
 		Bdx.updateInput();
+		Bdx.profiler.stop("input");
 
 		for (Scene s : (ArrayListNamed<Scene>)Bdx.scenes.clone()){
 			s.update();
+			Bdx.profiler.start();
 			renderScene(s);
+			Bdx.profiler.stop("render");
 		}
+		Bdx.profiler.update();
 	}
 	
 	
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -7,6 +7,7 @@ import com.badlogic.gdx.files.FileHandle;
 import com.nilunder.bdx.inputs.*;
 import com.nilunder.bdx.audio.*;
 import com.nilunder.bdx.utils.ArrayListNamed;
+import com.nilunder.bdx.utils.Profiler;
 
 public class Bdx{
 
@@ -40,6 +41,7 @@ public class Bdx{
 
 	public static final float TICK_TIME = 1f/60f;
 	public static float time;
+	public static Profiler profiler;
 	public static ArrayListScenes scenes;
 	public static Sounds sounds;
 	public static Music music;
@@ -50,6 +52,7 @@ public class Bdx{
 
 	public static void init(){
 		time = 0;
+		profiler = new Profiler();
 		scenes = new ArrayListScenes();
 		sounds = new Sounds();
 		music = new Music();
--- a/src/com/nilunder/bdx/Scene.java
+++ b/src/com/nilunder/bdx/Scene.java
@@ -623,19 +623,24 @@ public class Scene implements Named{
 		
 		if (!paused){
 
+			Bdx.profiler.start();
 			runObjectLogic();			
+			Bdx.profiler.stop("logic");
 			
 			updateVisuals();
+			Bdx.profiler.stop("visuals");
 			
 			updateCamera();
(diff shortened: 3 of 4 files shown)


GoranM:
The `init` method is not a constructor, so it's never called. If you wish to make a constructor, you must use a public method, without a return type, that shares the name of the class:

```java
public Profiler(){
    // initialization code
}
```

Don't initialize the same data in multiple places - initialize only in the constructor.

`pcts` is not clear enough - use `percents`.

Instead of `profData`, perhaps it would be better to have the Profiler itself extend from 
`HashMap<String, HashMap<String, Long>>`.

`data` should be renamed to `times`.

`profileDeltaTime` should be renamed to `totalTime`, and it should represent all the time it takes to get from one update to another, not just the combined time of all the code that was measured; We want to know the percentage of time relative to everything else, including stuff that is outside BdxApp.render (in the BGE, that's what the "Outside" category represents).

As I demonstrated in my last post, I would want to have the option to just call the stop method for a category, even without a previous start:

```java
p.start("blah");
// blah code
p.stop("blah");
// after blah
p.stop("after");
// more
p.stop("more");
```

So, if there are circumstances where I want to time methods that follow one after the other, I can do so without calling start for all of them.

rafcolson:
Thanks for your comments. I will update it tomorrow.

rafcolson:
@GoranM Can you have a look? I chose ```totalStart()``` and ```totalStop()``` but let me know if you would like to rename those.

Obviously there would have to be an extra option to enable/disable the profiler. But do you mind that the profiler runs in the background even if it is disabled? Because I could reduce it to just one IF in ```Scene.java``` and also in  ```BdxApp.java``` (with the corresponding extra code). I guess I could just ignore the instantiating in ```Bdx.java```.

SolarLune:
An extra blend file wouldn't be necessary, just the .bdx file that gets generated from it. But yeah, if you could do it somehow in code, that would probably be good.

rafcolson:
Ah okay, I'll have to check the .bdx file then. Thanks. If it would be simpler, I'll use that.

Nananas:
This overlay scene can be hard-coded, together with a `DEBUG` conditional compilation flag. Like:
```
```

Nananas:
A release version of a game should not contain the profiler, right?
Afaik, there are no real macros/compiler flags in Java. Maybe use something like [this](http://www.javapractices.com/topic/TopicAction.do?Id=64)? 

Maybe two different `Profiler` files, where the one for the release build has the flag set to false, so the compiler just ignores everything related to the Profiler? (and two different build buttons, probably)

Or does gradle allow custom build flags (I could not find anything that worked)?

Just thinking out loud.

rafcolson:
>just the .bdx file for the overlay scene

On second thought I don't know what you mean with .bdx file.

Anyway, I'll try to do it dynamically.

SolarLune:
> On second thought I don't know what you mean with .bdx file.

When you export and run a BDX project, it exports each scene in the current blend file as a .bdx file. It can be found in the android/assets/bdx/scenes directory. I was saying that you could build the blend file, export it as a .bdx file, and then just include the resulting "profiler.bdx" file in new projects.

@Nananas - I don't see why a release version of the game couldn't contain the profiler as well. It probably wouldn't be active, but if it's a negligible pull on CPU, then I wouldn't even worry about it too much.

GoranM:
This doesn't seem to measure the time spent outside BdxApp.render.

Your use of super is redundant: You didn't override anything from the base class, so you don't need super to use original methods (just call them as you would call your methods).

Don't worry about the display, for now: In addition to a .bdx file, we'll also need a default font texture, and font data file, along with some systems to manipulate the size and position of the display (that will depend on resolution).

It's not something that should go into this PR, and it's probably going to more complicated than it seems.

I would agree with SolarLune that profiler's performance impact is negligible.

PS: Discussions that don't pertain (directly) to this PR are off topic.

rafcolson:
>This doesn't seem to measure the time spent outside BdxApp.render.
Where should the profiler start and stop?

GoranM:
In the same place, just like it did for the update method you had before.

rafcolson:
I can't really remember where I put it. Was it at the very beginning of BdxApp.java?

GoranM:
BdxApp.render is the BDX main loop, but there is code that runs outside of that area (operating system, various system services, etc), so given a profiler.update function, called at the end of BdxApp.render, the total time would be time since last update.

```
profiler.update: totalTime = timeNow - timeLast; timeLast = timeNow;
```

rafcolson:
Yes of course. Why didn't I see that.

rafcolson:
Updated.

GoranM:
Ok, so now you just have to calculate "outside" time: totalTime - allDeltas.

rafcolson:
Done. I also changed to LinkedHashMap, because I think "outside" should be on the bottom of the list, visually. Let me know if I should change the names. Maybe the use of the Profiler should be more intensive, like really going into functions to collect "physics" for example. I'm not familiar with everything, so I leave that up to you.

Thanks for your help. Most of the credit should go to you because of that, but I can't think of anything better, so I couldn't be happier.

GoranM:
Ok, this seems fine. Thanks!

Issue #140, opened 2015-02-02
Text objects share meshes


SolarLune:
Hey, there. So, by default, text objects share meshes; if you add one in-game, for example, and edit it, any others you add will similarly be edited. I think they should by default get unique meshes, since you rarely (if ever) want one text object to mirror another.

rafcolson:
Yes I think that would best.

GoranM:
Yea, I think it should just be a matter of calling useUniqueMesh on new instances of Text.

SolarLune:
OK, I'll assign this to myself, then.

SolarLune:
PR submitted.

GoranM:
Please remember to close resolved issues.

PR #143, opened 2015-02-03
(Bug #140) - Text objects use unique meshes.

SolarLune:


GoranM:
If I were to make a `TextObjectName.java`, and I wanted to override init, wouldn't I have to explicitly call `super.init()` for this to work?

Perhaps it would be better to simply call this in Scene.cloneNoChildren:

```java
t.useUniqueMesh();
```

SolarLune:
Fixed. I had to move the instanceof checks down so that I could use the `if g instanceof Text` check after the modelInstance was set.

PR #144, opened 2015-02-03
Added Collision group & mask data to rigidbody

Nananas:
I added this collision data to the `GameObject`. Don't know if this is really necessary...

When changing the group or mask, the `Rigidbody` should be re-added to the scene. -> Todo?

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -408,7 +408,9 @@ def srl_objects(objects):
                 "mass": 0 if static(obj) else obj.game.mass,
                 "friction": obj.active_material.physics.friction if obj.active_material else 0.5,
                 "restitution": obj.active_material.physics.elasticity if obj.active_material else 0,
-                "ghost": obj.game.use_ghost
+                "ghost": obj.game.use_ghost,
+                "group": sum([2**i for i, v in enumerate(obj.game.collision_group) if v]),
+                "mask": sum([2**i for i, v in enumerate(obj.game.collision_mask) if v])
             }
         }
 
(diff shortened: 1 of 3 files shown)


GoranM:
> When changing the group or mask, the Rigidbody should be re-added to the scene. -> Todo?

Is that all it takes? If you can make that work, and it tests fine, then I would include it, but I don't think it's a good idea to have additional GameObject methods if they don't do anything.

For now, it would probably be best to focus on just exporting and applying initially defined group/mask settings from blender (no need to change anything in GameObject until you can actually get it to work).

About your export code:

Is there a reason why `mask` is a negative number? I mean, in the [bullet documentation](http://www.bulletphysics.org/mediawiki-1.5.8/index.php?title=Collision_Filtering), there seems to be no such requirement - both bit fields are positive numbers.

Instead of extra variables, and multi-line loops, you could derive the bit field value like this:

```python
"group": sum([2**i for i, v in enumerate(obj.game.collision_group) if v])
```

It's less code, while being no less clear.

Nananas:
Haha yeah my calculations are odd. They where experimental... 
I started with the exact same code for `mask`, which would be logical. But I think I just messed up my mask settings in blender. Sorry about that. :innocent:

> "group": sum([2**i for i, v in enumerate(obj.game.collision_group) if v])

Nice, looks really smooth.

I removed the `GameObject` code and will make a new PR if I can make that change work.

Nananas:
It seems doing the following works:
```
// Inside a GameObject
scene.world.removeRigidBody(this.body);
scene.world.addRigidBody(this.body,newCollGroup ,newCollMask);
```

But this works too:
```
// Inside a GameObject
this.body.getBroadphaseProxy().collisionFilterGroup = newCollGroup;
this.body.getBroadphaseProxy().collisionFilterMask= newCollMask;
```

Which would be prefered/correct?

GoranM:
Instead of additional commits, you should have used the --amend option to change the existing commit.

You need to squish this into one commit.

Nananas:
Alright, thanks. I learned something nice today :+1:.

PR #145, opened 2015-02-04
Collision group & mask update

Nananas:
##### Continued from #144 : 

Allows change to the collision group and mask. (Also added getters and setters).

Added a `updateCollisionFilter()` function that removes and adds the `RigidBody` to the world. This can be extended with a `dirty` flag, only removing/adding the `RigidBody` when the group & mask have been changed...

(This is based of the `master` branch. I don't really know if this is very smart.

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -272,7 +272,7 @@ public class GameObject implements Named{
 		return touchingObjects.get(name) != null && 
 			touchingObjectsLast.get(name) == null;
 	}
-	
+
 	public float reactionForce(){
 		float force = 0;
 		int totalContacts = 0;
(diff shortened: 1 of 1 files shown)


GoranM:
`updateCollisionFilter()` seems unnecessary, and the same can be said for storing the value of group/mask on the game object: You can just re-add the body in the setters, and the getters can simply return the value of `body.getBroadphaseProxy().collisionFilterGroup/Mask`.

Nananas:
Updated.

Note: This will throw a nullpointerException if the rigidbody has not been added to the scene. Do you think a check is necessary? Or will it make debugging harder?

GoranM:
No need for that first commit, because it contains changes that have already been merged in from the last PR.

The second commit should be rebased on top of existing code in master.

Don't worry about checks for now.

Nananas:
Uh ok. I thought github handled the rebase on its own.
Is this better?

PR #147, opened 2015-02-04
Filtered ray

Nananas:
##### Continued from #144 or #145 :

Added collision group and mask to `ray` and `xray`.
I did not change `mouse.clicked()`, should there also be an additional method with group & mask filter?

### DIFF
--- a/src/com/nilunder/bdx/Scene.java
+++ b/src/com/nilunder/bdx/Scene.java
@@ -384,12 +384,19 @@ public class Scene implements Named{
 		world.removeRigidBody(g.body);
 		toBeRemoved.add(g);
 	}
-	
+
+
 	public RayHit ray(Vector3f src, Vector3f vec){
+		return ray(src, vec, (short)~0, (short)~0);
+	}
+	
+	public RayHit ray(Vector3f src, Vector3f vec, short group, short mask){
 		Vector3f to = new Vector3f(src);
 		to.add(vec);
 		
 		CollisionWorld.ClosestRayResultCallback rrc = new CollisionWorld.ClosestRayResultCallback(src, to);
+		rrc.collisionFilterGroup = group;
+		rrc.collisionFilterMask = mask;
 		
 		world.rayTest(src, to, rrc);
 		
@@ -403,8 +410,13 @@ public class Scene implements Named{
 		
 		return rh;
 	}
-	
+
+
 	public ArrayList<RayHit> xray(Vector3f src, Vector3f vec, boolean includeAll){
+		return xray(src, vec, includeAll, (short)~0, (short)~0);
+	}
+	
+	public ArrayList<RayHit> xray(Vector3f src, Vector3f vec, boolean includeAll, short group, short mask){
 
 		Vector3f startPos = new Vector3f(src);
 		Vector3f dist = new Vector3f(vec);
@@ -416,7 +428,7 @@ public class Scene implements Named{
 
 		while (!finished){
 
-			RayHit ray = ray(startPos, dist);
+			RayHit ray = ray(startPos, dist, group, mask);
 
 			if (ray != null){
 
--- a/src/com/nilunder/bdx/inputs/Pointer.java
+++ b/src/com/nilunder/bdx/inputs/Pointer.java
@@ -51,11 +51,14 @@ abstract class Pointer {
 		return dir;
 	}
 	
-	public RayHit ray(){
-
+	public RayHit ray(short group, short mask){
 		Vector3f v = rayDirection();
 		v.length(100);
 
-		return scene.ray(raySource(), v);
+		return scene.ray(raySource(), v, group, mask);
+	}
+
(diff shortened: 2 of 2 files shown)


Nananas:
The default values for:
`group` = 1
`mask`= -1

Which means only object with a mask with the first bit 1 will collide with the default ray. 
So I set the `group` filter to `-1`, which means all. I think this is more consistent: just collide with everything.

GoranM:
Instead of `-1`, you should use `~0` to specify all bits high (I doubt many people will know/remember the details of signed representation, so -1 could confuse people).

In Pointer.java, I think `ray()` can now simply call `ray(1, ~0)`

Assuming that ray filtering is no different from object filtering (in terms of how group and mask flags are used), the value of group should be irrelevant, as long as mask is set to `~0`, but I guess 1 would be a good default value, so that's fine.

Nananas:
Okie, I'll use `~0`.


> In Pointer.java, I think ray() can now simply call ray(1, ~0)
> 
> Assuming that ray filtering is no different from object filtering (in terms of how group and mask flags are used), the value of group should be irrelevant, as long as mask is set to ~0, but I guess 1 would be a good default value, so that's fine.

Well, about that, according to [this source](http://bulletphysics.org/mediawiki-1.5.8/index.php/Collision_Filtering):

> Note: As of 17/08/10, collision masks must match both ways for there to be a collision. The above example is incorrect. If WALL collides with NOTHING, it will actually collide with nothing even though PLAYER is set to collide with WALL. WALL must collide with PLAYER and PLAYER must collide with WALL for there to be a collision.

So I thought that making the group of a ray all bits high would be more logical, as this means collision with everything. Having an object with the first bit of the mask 0 would result in no collision with the ray...

Or is this the desired behavior?

GoranM:
Oh, I didn't notice that update ... They should have just updated the example itself.

Anyway, yea just use `~0` for both in that case.

Nananas:
Anything still to change? Should I make one commit of those maybe?

GoranM:
Oh, I didn't notice that you changed anything (I don't get notified unless you actually make a new comment).

Yeah, just make it into a single commit.

Nananas:
Single commit: done.

SolarLune:
Hey, Nananas, would you update the ray() and xray() function API ref. about using collision groups?

Nananas:
Uh, I added some quick info about the added functions. Also for #145.

However, I think the linked wiki document about collision filtering is pretty bad.

SolarLune:
Thanks. Yeah, it's definitely not as good as it could be. It might be worth it to write our own explanation on the API reference for the collision group functions, and link to that explanation in the ray() and xray() functions. A short isn't the same as the example in the Bullet wiki (BIT() function calls?), so mentioning that would probably also be worth mentioning, I think. 

As a side-note, why do ray() calls require both collision masks and collision groups? Couldn't it just require one and set both mask and group to be the same under the hood, since we only need the ray to check for a collision with another object (whichever group that object is in)? I thought groups are what collision group an object is in, and masks are which collision group objects collide with; but since rays don't collide with anything, do they need a group?

Or am I missing something?

Nananas:
Writing our own is a possibility.


> Couldn't it just require one and set both mask and group to be the same under the hood, since we only need the ray to check for a collision with another object (whichever group that object is in)?

Well, I think, that is because of the following, quoted from the [bullet physics collision filtering page](http://www.bulletphysics.org/mediawiki-1.5.8/index.php/Collision_Filtering)
> Note: As of 17/08/10, collision masks must match both ways for there to be a collision. The above example is incorrect. 

Which means something like this:
```
bool collide =
          (A.maskBits & B.groupBits) != 0 &&
          (A.groupBits & B.maskBits) != 0;
```

And I think this also applies to bullet physics. Does this explain it?

PS: as such, now that info site is even less useful :see_no_evil:

SolarLune:
OK, I get that a collision in the traditional sense only works if A collides with B and B collides with A (you can't have one-way collisions). What I don't get is why the ray function requires this, since it is a one-way collision check. 

Do we need to set the mask and group bits separately? Couldn't the ray's filter group be the default "all bits", with the filter mask being customizable?

Nananas:
I'm trying to come up with a situation where the group would be useful, but, unless you have a really complex game, having only the mask as filter seems to be enough...

So, do you prefer an additional `ray(short mask)` function, or just reduce functionality of the original `ray(group, mask)` by defaulting the group part to all?

SolarLune:
I'd think just reducing the functions to omit the group argument would be fine. Is there a situation where you'd need a separate group argument?

Nananas:
Bof, maybe when 8 mask bits isn't enough, and you have some complicated way of grouping the collision filters... maybe? Don't know :grimacing:.
I looked at some other implementations of raycasting collision filtering, and they all used only masks, so I guess, practically, the group part can be removed.

Should I or can you?

SolarLune:
You can do it.

Issue #92, opened 2015-01-23
Camera 3D to 2D function


SolarLune:
Hey, there. A "getScreenPosition()" function on the camera class would be useful; it'd convert a 3D Vector3f to a 2D Vector2f indicating the point's position on screen from (0, 0), to (1,1). I usually use such a function to do something like display information about a character next to them in flat 2D space in a GUI overlay scene (rather than doing it in 3D in the game scene).

GoranM:
I think it's a good idea, but I don't see the need for a "get" prefix: 
`screenPosition(Vector3f worldPosition)` seems like a nice name.

Also, when you say from (0,0) to (1,1), do you mean OpenGL traditional, where origin is at bottom left, or do you just mean normalized screen coordinates, where origin is at top left?

In either case, it seems like you would have to convert that to world coordinates of your overlay scene, and this conversion would require screen width and height.

It might be more convenient for screen position to derive from something like clip coordinates, where (0, 0) would be at the center, with X ranging from -1 to 1, and Y from -1 * height/width to 1 * height/width (assuming width > height - otherwise X ranges would be multiplied by inverse aspect ratio, while Y would be -1 to 1).

Perhaps that would be better?

SolarLune:
That sounds pretty good, I'd suppose. I had openGL coords in mind, I suppose, but there's no need for them to have any real base in OpenGL (I don't suppose). If clip coordinates are the precedent, then there's no reason to make it more complex.

screenPosition() is fine, and the position scheme sounds OK.

I'm not sure quite how this should be approached, though, as the base Camera class in BDX lacks the necessary properties to implement this feature (no view matrix is accessible). It also lacks a pointer to the associated LibGDX camera class that I assume it relates to when you actually create a camera in the Scene.java file. What's your idea for this? Should the Camera class extend the LibGDX camera class, or should the LibGDX camera be a property of the BDX Camera class (like Camera.data or something like that)?

GoranM:
I don't see the benefit in extending LibGDX PerspectiveCamera, or even holding a direct reference to it on the Camera class. There is only one PerspectiveCamera per scene, and it is available from scene.cam, which should be more than sufficient.

The "view matrix" is just the inverse of the camera transform, so that's not a problem.

Generally, I think the implementation would be similar to that of [raySource](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/inputs/Pointer.java#L24), but done in reverse.

SolarLune:
[...]

PR #162, opened 2015-02-07
screenPosition() function (#92). 

SolarLune:
Added `screenPosition()` function to convert world coordinates to screen coordinates (#92).


GoranM:
There goes your incentive to learn about projection. :)

PR #169, opened 2015-02-12
Profiler adjustments (#76)

rafcolson:
Points to consider:
- User defined categories may not start with ```"__"```. Is there a better alternative?
- User defined categories will be added at the bottom of the list. Would it be better to have a separate linked hash map for ```userPercents```?
- Eventually, when we have an overlay scene to display the profiler, should milliseconds also be displayed? If so, I also should add a hash map ```millis```.
- Now there's a method ```percents()```, but eventually, should this be removed? Because everything would already be displayed, which would make it a bit obsolete.

### DIFF
--- a/blender/bdx/gen/BdxApp.java
+++ b/blender/bdx/gen/BdxApp.java
@@ -40,19 +40,19 @@ public class BdxApp implements ApplicationListener {
 
 	@Override
 	public void render() {
-		Bdx.profiler.start();
+		Bdx.profiler.start("__graphics");
 		Gdx.gl.glViewport(0, 0, Gdx.graphics.getWidth(), Gdx.graphics.getHeight());
 		Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
-		Bdx.profiler.stop("graphics");
+		Bdx.profiler.stop("__graphics");
 
 		Bdx.updateInput();
-		Bdx.profiler.stop("input");
+		Bdx.profiler.stop("__input");
 
 		for (Scene s : (ArrayListNamed<Scene>)Bdx.scenes.clone()){
 			s.update();
-			Bdx.profiler.start();
+			Bdx.profiler.start("__render");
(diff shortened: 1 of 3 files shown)


GoranM:
> User defined categories may not start with "__". Is there a better alternative?

Well, you're not actually enforcing that, so its just a guideline. I guess it would be better to have a system that throws an exception if the developer tries to use a reserved name, but it's not critical requirement.

> User defined categories will be added at the bottom of the list. Would it be better to have a separate linked hash map for userPercents?

Adding to the bottom seems fine to me.

> Eventually, when we have an overlay scene to display the profiler, should milliseconds also be displayed? If so, I also should add a hash map millis.

Yeah, millis or deltas - that would be fine.

Instead of adding a new method, maybe it would be best to simply make those result hashes public members:

```java
prof = Bdx.profiler
prof.percents.get("category")
prof.deltas.get("category");
```

About your naming, in general:

It's a bit more difficult to understand code where adjacent names are all very similar. For example:

```java
startTime
stopTime
startTimes
```

See how they all kinda look the same?

This is easier to read:

```java
start
stop
startTimes
```

On a quick scan, you can quickly see the differences, so it's much more difficult to confuse one for the other - they're easier to distinguish.

So, in addition to relevant meaning for the names, you should also consider the overall scannability of your code - strike a good balance between the two.

It's not something you have to do for this patch (I mean, if you think it looks fine as it is, I'll merge it), but just something to keep in mind (or at least think about) in the future.

rafcolson:
>Well, you're not actually enforcing that, so its just a guideline. I guess it would be better to have a system that throws an exception if the developer tries to use a reserved name, but it's not critical requirement.

I would like that better also but I don`t know how to do it considering user categories should be identifiable. Those deltas may not be added to the total delta time. Maybe there is a direct way to know if a category happens in GameObject?

rafcolson:
Updated.

Issue #172, opened 2015-02-16
Bug: BDX crashes if you rename an image


SolarLune:
Hey, there. So BDX crashes on export of a project if you rename a resource (like an image). It looks like BDX uses the resource name in Blender (the name of the image, for example) as opposed to the path to the file as the filename of the resource to use.

GoranM:
You can fix it [here](https://github.com/GoranM/bdx/blob/master/blender/bdx/exporter.py#L437), if you want.

SolarLune:
PR Submitted.

SolarLune:
PR Merged. Thanks! Issue closed.

PR #175, opened 2015-02-23
(Bug #172) - Use filepath rather than the image name when exporting.

SolarLune:


GoranM:
Wouldn't it be better to use [os.path.basename](https://docs.python.org/3.4/library/os.path.html#os.path.basename)?

SolarLune:
Fixed.

Issue #185, opened 2015-03-14
Suggestion: new Component: "Track"


rafcolson:
Hi guys,

It would be nice to also have a Component to follow a target. This especially would be useful to add to a Camera with 3D games, in combination with the "Follow" component.

I didn't pull a request for this, because I stumbled into some issues with [alignAxisToVec](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/GameObject.java#L469). Like it is in the BGE, the orientation gets messed up after a while. So maybe there would be a better way to track to the target?

This is what I have now:

```java
package com.nilunder.bdx.components;

import com.nilunder.bdx.Component;
import com.nilunder.bdx.GameObject;
import com.nilunder.bdx.State;

import javax.vecmath.Vector3f;

public class Tracker extends Component {
	public GameObject target;
	public int axis;
	public Vector3f offset;
	
	public void set(GameObject target, int axis, Vector3f offset){
		this.target = target;
		this.axis = axis;
		this.offset = offset;
	}

	public Tracker(GameObject g, GameObject target, int axis){
		super(g);
		set(target, axis, new Vector3f());
		state = track;
	}
	
	public Tracker(GameObject g, GameObject target, int axis, Vector3f offset){
		super(g);
		set(target, axis, offset);
		state = track;
	}
	
	private State track = new State(){
		public void main(){
			if (!target.valid()){
				g.components.remove(this);
				return;
			}
			Vector3f vec = target.position().plus(offset).minus(g.position());
			vec.normalize();
			g.alignAxisToVec(axis, vec.plus(offset));
		}
	};
}
```

GoranM:
A "Track" component sounds like a good idea.

You can make a pull request for this, if you wish, and we can discuss the implementation there.

rafcolson:
Changed the title to "Track" and making a pull request now.

rafcolson:
PR merged.

PR #186, opened 2015-03-15
Track (#185)

rafcolson:
As mentioned in Issues, ```alignAxisToVec``` is not working properly, I think. The orientation gets messed up after a while. So maybe there would be a better way to track to the target? Or does ```alignAxisToVec``` need a fix?

### DIFF
--- a/src/com/nilunder/bdx/components/Track.java
+++ b/src/com/nilunder/bdx/components/Track.java
@@ -1,0 +1,38 @@
+package com.nilunder.bdx.components;
+
+import com.nilunder.bdx.Component;
+import com.nilunder.bdx.GameObject;
+import com.nilunder.bdx.State;
+
+import javax.vecmath.Vector3f;
+
+public class Track extends Component {
+	public GameObject target;
+	public Vector3f offset;
+	public String axis;
+	public String upAxis;
+	
+	private Vector3f upVec;
+	
+	public Track(GameObject g, GameObject target){
+		super(g);
+		this.target = target;
+		offset = new Vector3f();
+		axis = "Y";
+		upAxis = "Z";
+		upVec = new Vector3f(0, 0, 1);
+		state = track;
+	}
+	
+	private State track = new State(){
+		public void main(){
+			if (target == null || !target.valid()){
+				target = null;
+				return;
+			}
+			Vector3f vec = target.position().plus(offset).minus(g.position());
+			g.alignAxisToVec(upAxis, upVec);
+			g.alignAxisToVec(axis, vec);
+		}
+	};
+}


rafcolson:
As mentioned in Issues, ```alignAxisToVec``` is not working properly, I think. The orientation gets messed up after a while. So maybe there would be a better way to track to the target? Or doesn ```alignAxisToVec``` need a fix?

GoranM:
I don't really understand what you mean by "not working properly" - As long as the given axis aligns to a given vector, the method is correct. The direction of the other two axes is not relevant for the alignment operation, but if you want to have a specific "up" vector, you might get what you want by first aligning to Z, and then to the given vector:

```java
g.alignAxisToVec(2, new Vector3f(0, 0, 1));
g.alignAxisToVec(axis, vec);
```

Instead of multiple constructors, you could simply have this:

```java
public Track(GameObject g){
    super(g);
    axis = 1;
    offset = new Vector3f(0, 0, 0);
}
```

If users want to change those defaults, they can just set the public fields directly - That would be less code, while being equally convenient.

I don't think there's any need to remove the component once the target is invalid - You might want to set a different target at that point, and you should be able to use the same component for that. So, I think this would probably be better:

```java
if (!target.valid())
    target = null;
if (target == null)
    return;
```

And then, for the actual alignment, do you really need to normalize the vector? As far as I know, alignAxisToVec doesn't care about vector length.

rafcolson:
Changed the code. I had hoped there could be more control over the track-axis and the up-axis. When used on a Camera, there's the problem of the Camera pointing into the negative Z-direction. Also when setting ```axis``` to 2, the up-axis should be 0 or 1. So wouldn't it be better to specify both the track-axis and the up-axis? If so, we shouldn't use ```alignAxisToVec``` since this only aligns positive axis.

So do you have any ideas how this Track Component could have sort of the same parameters as the Blender's TrackTo Constraint?

GoranM:
> So wouldn't it be better to be able to specify both the track-axis and the up-axis?

Yea, that would probably be ok to add.

> So I guess my question is, do you have any ideas how this Track Component could have sort of the same parameters as the Blender's TrackTo Constraint?

It shouldn't be that difficult, but for now, just simple tracking is fine.

Oh, and change this:

```java
if (!target.valid())
    target = null;
if (target == null)
    return;
```

To this:

```java
if (target == null || !target.valid()){
    target = null;
    return;
}
```

So that we don't call valid() on a null target.

rafcolson:
Updated the code. Now the Track Component has full control over axis. By default the ```toAxis``` and ```upAxis``` are be "Y" and "Z". But if you would like a Camera track a GameObject, you could fill in the additional fields. You can also change the ```offset```, if needed.

```java
public class Cam extends Camera{
    private static Track track;
    private static GameObject target;

    public void init(){
        target = scene.objects.get("Sacky");
        track = new Track(this, target, "-Z", "X");
        track.offset = new Vector3f(0, 0, 1);
        components.add(track);
    }
}
```

I believe in this case it's better to have these two optional parameters in the constructor, because else you would have to spend another line setting those with the ```setAxis()``` function. But if you think it's unnecessary I'll remove it.

```java
track.setAxis("-Z", "X")
```

GoranM:
Most of this new code seems unnecessary.

I mean, all you really need is the upAxis field; The upVec can just be a constant Z, and for cameras, the proper upAxis would be Y.

rafcolson:
Well, I already tried every possible combination. For cameras it has to be "-ZX". If the up-axis would be "Y" then naturally the camera will pivot on the "Y", which is not what we want.

Another case, let's say your game is viewed from the top, the to-axis has to be "-Z".

Maybe I'm missing something here, so I'm curious if you'll come to the same conclusion after you tried out what you're suggestion. I think it's not possible. Here's the code I use on the camera

GoranM:
Could you please create a [packed BDX blend](https://www.youtube.com/watch?v=2bAcldnQxj0)?

rafcolson:
Done. The reason I copied the whole project is because the example not only uses the "Track" component but also the "Follow" component. Hope you don't mind.

GoranM:
It's still a .zip file: I need a .blend file.

rafcolson:
Here's a [blend](https://www.dropbox.com/s/0okpd7ac4dupolz/Track_test.blend?dl=0). But you'll need a build with the Track Component in it.

GoranM:
You can just make the component local to the class - no need for a special build.

As you can see in [this example](https://www.dropbox.com/s/g3pxpnx3s4lg64o/track_test_r1.blend?dl=1), the upVec can just be a constant, and the upAxis for the camera would just be Y.

That said, I think that using strings to specify an axis is probably a good idea, because we can't use integers to specify -X (can't have a -0). Might be best to create a version of alignAxisToVec which takes axis strings.

rafcolson:
Thanks.

I didn't realize the upVec should be constant.

>As you can see in this example, the upVec can just be a constant, and the upAxis for the camera would just be Y.

I'm not sure what you are suggesting, but I guess you'd prefer to evaluate if ```g``` is an instance of ```Camera```? Or should I keep the method I had?

The latter would give more control over particular cases where you may want to track the top of an object to some other object, let's say with planes. So I would prefer this over distinguishing between ```GameObject``` and ```Camera```.

What do you think? I could make some examples of what I think are user cases.

>That said, I think that using strings to specify an axis is probably a good idea, because we can't use integers to specify -X (can't have a -0). Might be best to create a version of alignAxisToVec which takes axis strings.

Of course a version of ```alignAxisToVec``` which takes axis strings would be a good addition, but to me it seems a bit trivial to be using this in the Track component, because you would have this extra code in the main loop (which else would be in ```init()``` only):
```java
if (axis.startsWith("-")){
	axis = axis.replace("-", "");
	direction = -1;
}else{
	direction = 1;
}
toAxis = "XYZ".indexOf(axis.charAt(0));
upAxis = "XYZ".indexOf(axis.charAt(1));
upVec = new Vector3f(0, 0, 1);
```

GoranM:
With the new alignAxisToVec method (just commited), you can simply have this:

```java
public class Track extends Component {
	public GameObject target;
	public Vector3f offset;
	public String axis;
	public String upAxis;
	
	private Vector3f Z;
	
	public Track(GameObject g, GameObject target){
		super(g);
		this.target = target;
		offset = new Vector3f();
		axis = "Y";
		upAxis = "Z";
		state = track;
		Z = new Vector3f(0, 0, 1);
	}
	
	private State track = new State(){
		public void main(){
			if (target == null || !target.valid()){
				target = null;
				return;
			}
			Vector3f vec = target.position().plus(offset).minus(g.position());
			g.alignAxisToVec(upAxis, Z);
			g.alignAxisToVec(axis, vec);
		}
	};
}
```

It's less code, and the logic seems very clear. Right?

rafcolson:
Works like a charm!

Thanks for the help.

GoranM:
> Just so you would know, you might have forgotten to use axis(int axis) in alignAxisToVec(int axis, Vector3f vec)

As far as I can tell, the current code is correct ... The axis method returns an axis vector, so there's no reason to use it when I'm just passing down to alignAxisToVec, which requires an axisName.

Anyway: I'll merge this now.

Issue #188, opened 2015-03-15
Suggestion: Mouse cursor hiding functions


SolarLune:
It'd be nice to have an ability to hide the system mouse cursor without capturing it. It looks like mouse.setCursorImage() does it; perhaps the ability to provide a null argument to just replace the cursor with a blank image / pixmap would be good?

GoranM:
Sounds good.

SolarLune:
PR submitted.

SolarLune:
PR merged. Thanks!

PR #189, opened 2015-03-16
Cursor Improvements (#188).

SolarLune:
Modified setCursorImage to hide the cursor if null is provided for the textureName.
Added another setCursorImage function call that only takes textureName for convenience.
Added showSystemCursor() function that shows the system cursor (undoes a custom cursor).


GoranM:
Wouldn't it make more sense if passing null to setCursorImage would undo the custom image, instead of hiding the cursor?

For mouse visibility, we could have a visible(boolean) method.

SolarLune:
Fixed as best as I knew how.

As for the PS - I find it more difficult to read lines right next to each other sometimes. I'll try to remember to align it with the style of code as it exists.

Issue #190, opened 2015-03-16
Suggestion: applyForceLocal() and velocityLocal()


rafcolson:
Actually, I was a bit surprised those aren't in yet. So how about it?

GoranM:
Sure, sounds good.

rafcolson:
Merged.

PR #192, opened 2015-03-17
applyForceLocal & velocityLocal (#190)

rafcolson:


GoranM:
You should just use the existing velocity method, instead of setLinearVelocity, so that you don't have to call body.activate().

PR #195, opened 2015-03-18
Basic 2D filter support and default shaders.

SolarLune:
Adds basic 2D filter (shader) support (only one filter for each Scene at the moment) to Bdx. The filter application system works similarly to the BGE's - a shader applied to one Scene is also applied to the underlying scenes.

Bdx.useFramebuffers variable added to help optimize things if you're not using filters. When not on, framebuffers aren't used and filters aren't applied. Bdx.useFramebuffers defaults to off and automatically gets turned on if you add a compiling filter to a scene, and turned off if a shader doesn't compile (so everything still renders correctly even without filters).

Also added default vertex and fragment 2D filters to assist in making customized versions (making it easy to copy and expound from to make the shader you want).

To-do features:

* Multiple filters per scene via a filter stack
* Custom Filter class to make setting filters in that stack easy
* Framebuffer downsampling
* Add depth texture uniform
* Add time uniform

EDIT: Oh, I forgot to open an issue for this here, sorry. It's in reference to [this](http://www.reddit.com/r/bdx/comments/2zdkpu/about_postprocessing/) thread over on /r/bdx.

### DIFF
--- a/blender/bdx/gen/BdxApp.java
+++ b/blender/bdx/gen/BdxApp.java
@@ -6,8 +6,13 @@ import com.badlogic.gdx.ApplicationListener;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.graphics.GL20;
 import com.badlogic.gdx.graphics.PerspectiveCamera;
+import com.badlogic.gdx.graphics.Pixmap;
+import com.badlogic.gdx.graphics.Texture;
+import com.badlogic.gdx.graphics.g2d.SpriteBatch;
+import com.badlogic.gdx.graphics.g2d.TextureRegion;
 import com.badlogic.gdx.graphics.g3d.ModelBatch;
 
+import com.badlogic.gdx.graphics.glutils.FrameBuffer;
 import com.nilunder.bdx.Bdx;
 import com.nilunder.bdx.GameObject;
 import com.nilunder.bdx.Scene;
@@ -19,11 +24,13 @@ public class BdxApp implements ApplicationListener {
 
 	public PerspectiveCamera cam;
 	public ModelBatch modelBatch;
+	public FrameBuffer frameBuffer;
+	public SpriteBatch spriteBatch;
 
 	@Override
 	public void create() {
 		modelBatch = new ModelBatch();
-		
+
 		Bdx.init();
 		Gdx.input.setInputProcessor(new GdxProcessor(Bdx.keyboard, Bdx.mouse, Bdx.allocatedFingers));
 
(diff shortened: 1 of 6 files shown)


rafcolson:
I'm curious. Does this enable to make materials smooth shaded?

SolarLune:
@rafcolson - Ah, no, this is for screen filters. I see, you misunderstood "default shaders". My mistake, I meant default 2D filters (which are shaders). I haven't done anything with lighting yet.

rafcolson:
Okay, thanks.

GoranM:
If you uncheck Transparency on the G_Sacky material, Sacky will be rendered behind the platform. I think this happens because you passed false in the FrameBuffer constructor, for the hasDepth argument; pass true instead, and it should work fine.

I don't think you need Scene.setShader: You can simply set the renderedBufferPass to a new ShaderProgram. In BdxApp.renderScene, you can check if it compiled properly, and if not, you should throw an exception which prints the compile log - If the shader code is wrong, it doesn't make sense to keep the game running.

I don't think you need Bdx.useFramebuffers: Whether or not you should render a scene to an offscreen buffer depends solely on Scene.renderedBufferPass. If it's null, that means there is no shader, and so, that scene should not render to an FBO.

SolarLune:
> If you uncheck Transparency on the G_Sacky material, Sacky will be rendered behind the platform. I think this happens because you passed false in the FrameBuffer constructor, for the hasDepth argument; pass true instead, and it should work fine.

Thanks, fixed.

> I don't think you need Scene.setShader: You can simply set the renderedBufferPass to a new ShaderProgram. In BdxApp.renderScene, you can check if it compiled properly, and if not, you should throw an exception which prints the compile log - If the shader code is wrong, it doesn't make sense to keep the game running.

I thought it best to write a function for this to avoid making the user import and use LibGDX's ShaderProgram stuff. This way they can just pass the shader files / shader strings directly rather than mixing BDX and LibGDX stuff (though they still use LibGDX to handle opening the shader files themselves). 

As I'm sure you know, my eventual goal is to further develop the filter system and move Filters into a custom class, as you suggested our classes wrap existing LibGDX ones. The filter system as it is would be like this only temporarily.

I don't know if it makes sense to stop the game if shaders don't compile correctly. On the one hand, yeah, I'd say so, but on the other, small differences between graphics cards can pose problems. I think it better that a bloom filter not to work than for your game to crash for a large portion of people because they have an NVIDIA card or something.

> I don't think you need Bdx.useFramebuffers: Whether or not you should render a scene to an offscreen buffer depends solely on Scene.renderedBufferPass. If it's null, that means there is no shader, and so, that scene should not render to an FBO.

I suppose so. Fixed.

GoranM:
Well, if we're going to internalize the filter concept, then there's no need to wait: Rename Scene.renderedBufferPass to Scene.filter, and make a basic Filter class that extends ShaderProgram.

> I don't know if it makes sense to stop the game if shaders don't compile correctly. On the one hand, yeah, I see the advantage, but on the other, small differences between graphics cards can pose problems for shaders. I think it better that a bloom filter not to work than for your game to crash for a large portion of people because they have an NVIDIA card or something.

From what I understand, it's not a hardware issue; If the card supports GLSL, it supports it fully - It has to, by specification. It's just that ATI and Intel drivers compile GLSL in strict mode, while NVIDIA drivers are more forgiving by default. There was actually a BA thread in which this was explained to me: [here](http://blenderartists.org/forum/showthread.php?302926-this-bloom-filter-didnt-work-for-me&p=2437907&viewfull=1#post2437907).

We should simply include "#pragma optionNV(strict on)" at the top of the shader source, as suggested at the end of that thread.

SolarLune:
OK, fixed!

Issue #199, opened 2015-03-21
Suggestion: Random.random() With Argument


SolarLune:
Hey, there. In line with the Vector changes I suggested, it'd be nice if the random function took a couple of arguments (min and max, again). The default would just be like it works currently, and give 0 - 1.

GoranM:
That sounds fine.

SolarLune:
PR submitted.

rafcolson:
Why would it be useful to have a method: ```direction(float min, float max)```, if I may ask? Isn't a random direction always  ```random(-1, 1)```

SolarLune:
A random direction would be that, yes, but you can easily set the length with this to a random length between the minimum and maximum. I mean, it's a pretty self-explanatory function; the usefulness is up to the user of the function.

You could use it for placing objects in a circle in a range, or picking a random direction and speed to move in, or whatever.

GoranM:
If you need a random direction vector with a random length between a given range, you could simply do: `Random.direction().mul(Random.random(min, max))`.

I guess one could make the argument that Random.direction(min, max) is more convenient, but I don't think that "random direction with random length between given range" is a common enough requirement to justify that addition.

Also, it's a bit confusing, because direction(min, max) doesn't really make it clear that min can't be negative, or that the method will do something sane (like clip to 0) if it is - You basically have to look at the implementation to be sure.

SolarLune:
OK, just the random functions, then. PR updated.

SolarLune:
Cool, this was merged. Issue closed.

Issue #198, opened 2015-03-20
Consolidation of Random.Vector and Random.Direction


SolarLune:
Hey, there. I was just using Random's vector randomization functions when I realized they could easily be consolidated and expanded into a single, more useful function. It would take a minimum and a maximum length and return a random vector within that range. 

That would afford you the ability to do what Direction does by just having both min and max be 1, and gives you more control than Vector does. The function call with no arguments would just be the same as vector() worked, and give you a random vector between 0 and 1 length.

GoranM:
The methods in question are Random.vector and Random.direction, not Random.Vector and Random.Direction.

> That would afford you the ability to do what Direction does by just having both min and max be 1, and gives you more control than Vector does. The function call with no arguments would just be the same as vector() worked, and give you a random vector between 0 and 1 length.

Random.vector is supposed to return a vector with random *components* in range [-1,1].

SolarLune:
> The methods in question are Random.vector and Random.direction, not Random.Vector and Random.Direction.

Yes? Does it matter? You know what I'm talking about, right?

> Random.vector is supposed to return a vector with random components in range [-1,1].

OK, so then the function call with no arguments can return that, and the argumented function will return one of those vectors, but normalized, with a set length between the minimum and maximum values.

GoranM:
> Does it matter?

I believe so: By using the correct names, you spare your readers the otherwise required double take ... It also implies that you took the time to really think about, and subsequently craft your post.

> OK, so then the function call with no arguments can return that, and the argumented function will return one of those vectors, but normalized, with a set length between the minimum and maximum values.

I don't see how that would be better.

SolarLune:
>I believe so: By using the correct names, you spare your readers the otherwise required double take ... It also implies that you took the time to really think about, and subsequently craft your post.

I hardly see how a change in capitalization would require a double take (it's the same word), or imply that I didn't take the time to think about or craft my post. Did you have trouble understanding my question?

> I don't see how that would be better.

I thought it'd be useful to be able to create a vector randomization function that allows one to specify the range of the randomness. I could alter the direction() function to account for this, but I don't think we want to have multiple functions that do very similar things.

rafcolson:
> I hardly see how a change in capitalization would require a double take (it's the same word), or imply that I didn't take the time to think about or craft my post.

[...]

PR #201, opened 2015-03-21
(#199) - Random.random() with arguments.

SolarLune:
Additional random() function with ability to provide minimum and maximum values.

### DIFF
--- a/src/com/nilunder/bdx/utils/Random.java
+++ b/src/com/nilunder/bdx/utils/Random.java
@@ -22,9 +22,15 @@ public class Random{
 		return R.nextFloat();
 	}
 
+	public static float random(float min, float max){
+		return (min + (random() * (max - min)));
+	}
+
 	public static Vector3f direction(){
 		Vector3f vec = vector();
 		vec.normalize();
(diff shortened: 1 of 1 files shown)


GoranM:
Why not mini - maxi, and not min - max? Is there some significance to the trailing 'i'?

SolarLune:
Fixed. Also, the random vector improvements are in this commit because they depend on random.random.

GoranM:
That doesn't seem like the right code - It touches Bullet.java.

SolarLune:
Huh, I thought I didn't commit that. Fixed, thanks.

GoranM:
Seems like something went wrong again: Random.vector(min, max) is missing, and you have a Random.direction(min, max), which is not supposed to be there.

SolarLune:
Huh? You told me to change direction() instead of vector() in issue #198.

GoranM:
I said:

> You should change Random.direction, so that it can only return vectors of non-zero length.

I didn't tell you to overload the method with min/max arguments.

SolarLune:
I guess I misunderstood you. Fixed.

Issue #207, opened 2015-03-24
Suggestion: Compound Physics Type


SolarLune:
Hey, there. I think I could also make use of compound physics, so I'm going to take a look into that.

UPDATE: I think I have it working fairly satisfactorily. It works with parenting both in Blender as well as in-game. Parenting and de-parenting updates the compound meshes as expected. Multiple compound objects can also be parented together to form a larger compound object with no problems (as far as a bit of testing shows).

A problem I do see, though, is that it seems like this doesn't adjust the center of mass for the object when multiple objects are parented. I'm not quite sure how to best approach this so that it reacts accordingly. I see that there's a center adjustment function on the GameObject's body (parent.body.setCenterOfMassTransform). EDIT: Ah, it doesn't seem to be what this function does. Seems like I might have to take an alternative approach.

Should the center of mass be the average center point between all objects? Should it take into account the actual objects' respective sizes? What about their mass amounts? Or is it not really worth it? It doesn't seem to me like it's updating, but maybe Bullet's doing things behind the scenes and my test is just skewed...? Any thoughts?

UPDATE 2: After doing some reading, it would appear that Bullet has no knowledge of the center of mass point for objects and doesn't do anything regarding it. So I think I'll have to update it.

GoranM:
I would assume that bullet has some facilities to create compound objects, which would account for all the relevant attributes (position, mass, etc).

I have yet to research any of this, so, at this point, your guess is as good as mine.

As long as the PR is not too involved, and it works, I'll merge it.

SolarLune:
OK. I think I managed to get everything but adjusting the center-of-mass to work. I'll submit a PR after the physics bug I just reported is resolved, as that solution will probably touch the same files in the same places I'm working with.

SolarLune:
Cool, that was merged. PR submitted.

SolarLune:
Thanks for the merge. Issue closed.

PR #209, opened 2015-03-25
(#207) - Compound physics support.

SolarLune:
Adds support for compound physics using childrens' collision shapes if compound is activated for the parent in Blender.
TODO at some point: Add function / capability to shift center of mass on parent.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -413,7 +413,8 @@ def srl_objects(objects):
                 "restitution": obj.active_material.physics.elasticity if obj.active_material else 0,
                 "ghost": obj.game.use_ghost,
                 "group": sum([2**i for i, v in enumerate(obj.game.collision_group) if v]),
-                "mask": sum([2**i for i, v in enumerate(obj.game.collision_mask) if v])
+                "mask": sum([2**i for i, v in enumerate(obj.game.collision_mask) if v]),
+                "compound": obj.game.use_collision_compound
             }
         }
 
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -18,6 +18,7 @@ import com.bulletphysics.collision.narrowphase.PersistentManifold;
 import com.bulletphysics.collision.narrowphase.ManifoldPoint;
 import com.bulletphysics.collision.dispatch.CollisionFlags;
 import com.bulletphysics.collision.shapes.CollisionShape;
+import com.bulletphysics.collision.shapes.CompoundShape;
 import com.bulletphysics.dynamics.RigidBody;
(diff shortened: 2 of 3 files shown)


GoranM:
I'm not sure about updateCompoundShape:

The fact that it's always called, in the parent method, implies that every parent is a compound object that needs to update its shape.

Additionally, using the children list, in the implementation, to determine whether a given shape should be added to, or removed from the compound shape ... It seems convoluted, given that you know, from the call site in the parent method, what should be done.

Instead of updateCompoundShape, you could have a compoundShape() method, and do something like:

```java
public void parent(GameObject p){
	if (parent != null){
		parent.children.remove(this);

		if (parent.compoundShape() != null){
			parent.compoundShape().removeChildShape(this.body.getCollisionShape());
		}
	}
	
	parent = p;

	if (parent != null){
		
		parent.children.add(this);
		
		updateLocalTransform();
		updateLocalScale();

		if (parent.compoundShape() != null){
			parent.compoundShape().addChildShape(new Transform(this.localTransform), this.body.getCollisionShape());
		}

		dynamic(false);
	}else{
		dynamic(true);
	}
	
}
```

That seems easier to follow, right?

Oh, and here:

```java
			return compShape;
		}
		else
			return shape;
```

The "else" is redundant - You can just return shape at the end:

```java
			return compShape;
		}
		return shape;
```

SolarLune:
It seems to be a bit more convoluted to me to add a function to check to see if a body is compound (you can just do `body.getCollisionShape()` or whatever; no reason to make a whole new function for it). Also, I tried to avoid doing things on the parent object when possible, instead just calling the parent's updateCompoundShape() function; I thought that would be easier to follow, since it all happens in there, regardless of whether it should add or remove a shape from the compound shape.

In any case, your version seems to be less lines and simpler than my implementation (I think?), so it's cool.

Fixed it.

GoranM:
> It seems to be a bit more convoluted to me to add a function to check to see if a body is compound (you can just do body.getCollisionShape() or whatever; no reason to make a whole new function for it).

Not that it matters, because it's not an accurate measure of overall complexity, but the number of methods is the same - Instead of updateCompoundShape, we simply have compoundShape.

> I thought that would be easier to follow, since it all happens in there, regardless of whether it should add or remove a shape from the compound shape.

As I said:

> The fact that it's always called, in the parent method, implies that every parent is a compound object that needs to update its shape.
>
> Additionally, using the children list, in the implementation, to determine whether a given shape should be added to, or removed from the compound shape ... It seems convoluted, given that you know, from the call site in the parent method, what should be done.

There's more to clarity than simply placing everything into a single function.

> your version seems to be less lines and simpler than my implementation (I think?)

I don't know if it's less lines ("line count" is not that important), but I do think it's simpler, for reasons already given, and I think simplicity is very important.

Issue #214, opened 2015-03-27
Suggestion: Additional Built-in Uniforms for Filters


SolarLune:
Hey, there. I know we should be careful to keep the engine light, but I think we could stand to have a couple of additional uniform values for Filters.

I think it'd be nice if there was a "u_bufferTexture" uniform property that links to the texture rendered from the previous frame. It'd be useful for blurring, fog, and ghosting effects, I think.

A "time" uniform that just links to BDX's time property would be nice as well, since a lot of GLSL shaders rely on passing time.

GoranM:
> I think it'd be nice if there was a "u_bufferTexture" uniform property that links to the texture rendered from the previous frame. It'd be useful for blurring, fog, and ghosting effects, I think.

I don't see how it would be useful for blurring, or fog. I'm not really sure what you mean by "ghosting effects".

> A "time" uniform that just links to BDX's time property would be nice as well, since a lot of GLSL shaders rely on passing time.

That sounds fine.

SolarLune:
> I don't see how it would be useful for blurring, or fog. I'm not really sure what you mean by "ghosting effects".

I meant like motion blur, or a soft fading effect. This GLSL sandbox is what gave me the idea, since it's not really something that the user would be able to easily patch into the filter: http://glslsandbox.com/e#23433.10

GoranM:
Ok, but I'm not really sure if "u_bufferTexture" is the best name - Why not "lastFrame"?

SolarLune:
"lastFrame" is fine; I thought "u_bufferTexture" because the default texture uniform name LibGDX provides for the first texture is "u_texture". So, I thought keeping with that convention would be a good idea.

GoranM:
If following the convention has clear, and significant benefits, then it would but a good idea, but not otherwise.

SolarLune:
Uh, it's just a name. There's no benefits other than being different; either one is clear and easy to understand, so I don't really mind either way. 

I don't really see the need to have a separation between "how we do things" and "how LibGDX does things". If the user's going to see "u_texture" in their shader, seems like "u_bufferTexture" tells you pretty much everything you need to know - it's the buffered version of the texture that you already have. "lastFrame" is equally descriptive, but it doesn't tell you implicitly that it's the same as "u_texture". But like I said, I don't really mind; it's up to you.

GoranM:
> Uh, it's just a name.

In programming, names matter.

[...]

PR #215, opened 2015-03-29
(#214) Support for additional built-in Filter uniforms.

SolarLune:
"time" - A float that increases by 1 per second.
"lastFrame" - A sampler2D (texture) that is the complete, last rendered frame for the scene the filter runs in.

___

Changes:

- A new Buffer class has been created to sufficiently house and handle the frequent buffer updates the last frame uniform necessitates. It allows for drawing to another buffer or the screen, if null is passed. This could be expanded to work with FrameBuffers as well, rather than just BDX Buffers.

- Bdx has been altered to use multiple frame buffers (one for the scene, one for the final output, and a temporary one for shuffling) to properly handle drawing to a framebuffer with a shader. This was done as it seems OpenGL doesn't like writing to and reading from the same framebuffer at the same time.

- Each Scene now has a lastFrameBuffer Buffer that houses the texture from the last frame for that scene.

### DIFF
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -73,7 +73,7 @@ public class Bdx{
 
 	private static ArrayList<Finger> allocatedFingers;
 	private static ModelBatch modelBatch;
-	private static FrameBuffer frameBuffer;
+	private static RenderBuffer frameBuffer;
 	private static SpriteBatch spriteBatch;
 
 	public static void init(){
@@ -99,8 +99,10 @@ public class Bdx{
 		modelBatch = new ModelBatch();
 
 		ShaderProgram.pedantic = false;
-		frameBuffer = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), true);
+		
 		spriteBatch = new SpriteBatch();
+
+		frameBuffer = new RenderBuffer(spriteBatch);
 
 	}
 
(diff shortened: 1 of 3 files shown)


GoranM:
Why couldn't you simply draw the existing frameBuffer to scene.lastFrameBuffer?

SolarLune:
Ah, I missed that. Thanks! Fixed. Got rid of the scene buffer one and renamed the "finalBuffer" to just "frameBuffer".

GoranM:
I meant overall:

Starting from the [code currently in master](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Bdx.java#L168), it seems like it should be possible to simply draw that frameBuffer into something like scene.lastFrameBuffer.

If so, there would be no need for a "tempBuffer", and basically no need for a custom FrameBuffer class.

SolarLune:
Ah, I thought I needed a tempBuffer because I couldn't get it to work without swapping between a temporary buffer and the frameBuffer, but that must have been one of my non-working tests.

Updated. However, I think we should keep the Buffer class in, as otherwise the drawing process for BDX would get pretty hairy (as I've already experienced). It's much cleaner with single-line function calls to draw buffers.

GoranM:
At the moment, we only need to draw the frameBuffer twice (to scene.lastFrameBuffer, and to the display) - It seems like it would be less code to simply use the spriteBatch.

If there's a need to draw back and forth between more buffers in the future, we can add it then, but for now, I don't think it's necessary.

SolarLune:
Actually, we're drawing three times. Once per filter loop to the framebuffer, again to the screen, and once to the lastFrameBuffer. We also set the texture filtering on the scene's lastFrameBuffer and on the frameBuffer that we would have to add to Scene.Init() and Bdx.init(). 

I urge you to reconsider.

GoranM:
> We also set the texture filtering on the scene's lastFrameBuffer and on the frameBuffer that we would have to add to Scene.Init() and Bdx.init().

Ok, but I think "Buffer" is far too generic - FrameBuffer seems like the proper name, but if you want to keep it different from libgdx, you can use RenderBuffer, or FBO.

There's no reason to have the "setBatch" method - it's not being used externally. If there's a need for that in the future, we can make the relevant setter/getter methods (assuming that we can't simply make the field public).

It might be a good idea to rename draw to drawTo, and drawTo could take a destination, and a optionally a filter.

For clearing the buffer, it would be best to have a clear method.

SolarLune:
OK, I updated it accordingly.

GoranM:
> Just to confirm my earlier findings, you can't name a class after another class if one extends the other. So I went with "RenderBuffer".

It should be possible: You would just need to refer to one of the two classes by its fully qualified name, so that the compiler can make the distinction ... It's probably better to use RenderBuffer, just so that we don't have to think about that.

Anyway: This looks good, but there seems to be an indentation error in RenderBuffer.java, on line 42.

SolarLune:
> It should be possible: You would just need to refer to one of the two classes by its fully qualified name, so that the compiler can make the distinction ... It's probably better to use RenderBuffer, just so that we don't have to think about that.

I tried that, referring to the extended class as the fully-qualified LibGDX FrameBuffer. Eclipse informed me of an error with that, so I don't think it'll work, unless I missed something.

> Anyway: This looks good, but there seems to be an indentation error in RenderBuffer.java, on line 42.

Fixed it.

GoranM:
> Eclipse informed me of an error with that, so I don't think it'll work, unless I missed something.

What's the actual arror?

Did you remove the line that imports libgdx FrameBuffer?

SolarLune:
Ah, I see. Removing the importing line does allow the fully qualified name to be extended.

Anyway, thanks. PR closed.

Issue #219, opened 2015-03-29
Bug: Sphere Collision Type Incorrect Size


SolarLune:
Hey, there. 

So the Sphere collision type gives an incorrect size; I think it's due to mesh.calculateRadius returning a wrong value in Bullet. It's returning a radius of 1.131 for a block that's 2 BU tall and .75 BU wide and deep. Maybe it's testing the wrong axis or something, I dunno.

A simple solution would be to use half of the largest value for Sphere (which is how it's visualized in Blender).

GoranM:
I can't reproduce.

Can you please make a demo that clearly shows the nature of the issue?

SolarLune:
I'm kinda surprised you can't reproduce it. Here's a demo blend file.

https://www.dropbox.com/s/0fhioebvr7yp92q/spheretestpacked.blend?dl=0

- In the demo, the cube with Sphere bounds type doesn't sit on the ground. 

- Printing out the object's modelInstance.model.meshes.first().calculateRadius() (which is what Bullet.java uses to calculate the radius of the sphere collision shape) returns 1.7320508, even though the cube is 2x2x2, and so the radius of its sphere collision object should just be 1.

GoranM:
> Basically, I'd expect a sphere inside of the cube in the demo; I think it's returning a sphere that can contain the cube in the demo.

Yea. It makes sense for a collision shape to contain the mesh.

SolarLune:
I don't know about it making sense; it's a way to go. Just as equally as a collision shape being inside the mesh.

The main problem seems to stem from the fact that the visualization in Blender shows the collision shape inside the mesh. That means you have no real indication of where the collision mesh is; even worse when the shape is complex (Blender shows the capsule inside the object when it would really be surrounding the object, if I were to make the capsule "correctly"). It's probably best to just use the maximum value of the width, depth, or height of the bounds of the object rather than to actually get the shape that should "contain" the object.

GoranM:
If you want to export and use the bound values, I think that would be fine.

SolarLune:
I don't think that's necessary - just getting the bounding box and using half of the maximum amount out of all of the axes should be fine, right?

PR Submitted.

GoranM:
> I don't think that's necessary - just getting the bounding box and using half of the maximum amount out of all of the axes should be fine, right?

That still has the potential to be confusing, because it's not guaranteed to match the bounds visualization in blender.

But, if that's fine with you, than I'm ok with it.

[...]

PR #224, opened 2015-03-30
Use bounding box to calculate sphere shape radius (Bug #219).

SolarLune:
Matches up with Blender visualization better.


GoranM:
getDimensions is deprecated.

SolarLune:
Tweaked to use the non-deprecated version.

GoranM:
I guess you made the change just before I commented.

Issue #235, opened 2015-04-10
About Light and Shadows


SolarLune:
Hey, there. 

So I figure some discussion on lighting would probably be a good idea. It'd be really nice to add in to the engine. As far as I can tell, there's support for some lighting in LibGDX already (Ambient, Point, Spot, and Directional), so we could work on adding those. I don't think the default shader does anything for spotlights at the moment, but there's a PR waiting in the LibGDX repo that would add it, so maybe the next version will have support that we could take advantage of.

Now, would the idea be to create a single class for lights and change an internal `type` variable depending on what kind of light it is, or have a different class for each kind? I think I prefer the different classes because different light types use different variables. However, most of those variables would be shared (position, color, intensity, rotation (except for PointLights)), so the overall gain from different classes might be outweighed by the additional files.

Also, what's the naming convention? `PointLight`, `DirectionalLight`, and `SpotLight`? Ambient lighting doesn't need a whole class; it can just be directly exported from Blender and set.

Haven't looked into shadowing too much yet; I guess it'd be better to get lighting in and working and then consider how shadows could be implemented.

Anyway, thoughts?

rafcolson:
I'm excited for this to get in. It all sounds good. Is it possible to change the lamp type in real-time? If so, maybe the first question to ask is, do we need this? If not, then a single class for each light-type sounds better.

SolarLune:
Nothing's done right now, I'm just asking how it should be, though I don't imagine it to be too difficult to begin work on? I might be able to handle a basic approach.

> If so, maybe the first question to ask is, do we need this? 

I dunno; most times lights are pretty much "static" in terms of type. I'd imagine adding a new object to stand in for a changed light would work fine as opposed to altering a type variable.

GoranM:
> However, most of those variables would be shared (position, color, intensity, rotation (except for PointLights)), so the overall gain from different classes might be outweighed by the additional files.

That's my feeling.

SolarLune:
OK, I made some progress on this, but it seems like I might have run into a wall. Like mist, I haven't been able to get this to work, and since I wasn't able to get mist to work for a mysterious reason, I'm gonna pause and ask for confirmation that I'm not doing something stupid rather than waste time. Here's the code I have so far, which is committed to a branch on my fork of BDX:

https://github.com/SolarLune/bdx/commit/226f3ee46f3f3725c9b52761d12832cb174ba94d

[...]

PR #238, opened 2015-04-13
(#235) - Basic lighting support.

SolarLune:
Support for point, directional, and ambient lights. Energy works pretty much correctly for point lights; it doesn't do anything for directional, though.

Ambient lighting can be changed through the scene's ambientLight() function, or by setting the scene's ambient color in the world panel.

Objects can be drawn as shadeless by enabling it in the material panel in Blender.

Note that I'm using Vector4f()s for colors to be consistent in setting colors, even though lights don't use the alpha components. Also, the Light() class's color() function has to take Vector4fs to overload the GameObject color() function correctly.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -427,6 +427,12 @@ def srl_objects(objects):
         elif obj.type == "FONT":
             d["font"] = obj.data.font.name
             d["text"] = obj.data.body
+            
+        elif obj.type == "LAMP":
+            d['lamp'] = {"type": obj.data.type, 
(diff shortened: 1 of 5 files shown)


GoranM:
In Light.makeLightData:

```java
PointLight l = new PointLight();			
lightData = l;
```

Why not simply do: `lightData = new PointLight();`?

SolarLune:
Ah, I missed that. Thanks, fixed.

SolarLune:
Updated to not fail if there's no world set for the scene in Blender (as then there's no ambient color).

Issue #234, opened 2015-04-08
Suggestion: Have objects in group inherit instance scale


SolarLune:
Hey, there. So if you make a group instance and then change the instance's scale, the objects in the group appear to be unscaled. It'd be nice if they changed scale as well.

GoranM:
I guess so.

SolarLune:
Cool. I got the root parent object to be scaled by scaling it in the clone() function, like so:
```java
private GameObject clone(GameObject gobj){
		String instance = gobj.json.get("instance").asString();

		GameObject inst = gobj;
		if (instance != null)
			gobj = templates.get(instance);

		GameObject g = cloneNoChildren(gobj);

		for (GameObject c : gobj.children){
			GameObject nc = clone(c);
			nc.parent(g);
		}

		if (instance != null){
			g.position(inst.position());
			Matrix3f ori = inst.orientation();
			ori.mul(g.orientation());
			g.orientation(ori);
			g.scale(inst.scale());
						
			g.props = new HashMap<String, JsonValue>(g.props);
			g.props.putAll(inst.props);

			for (GameObject c : inst.children){
				GameObject nc = clone(c);
				nc.parent(g);
			}
		}

		return g;

	}
```

However, children of the root object aren't offset correctly like they appear in Blender. They just position themselves like they would be if the object weren't scaled. Are objects positioned or transformed initially somewhere else after the clone function is run?

GoranM:
> Are objects positioned or transformed initially somewhere else after the clone function is run?

Not to my knowledge.

SolarLune:
Hm, I guess I was just mistaken the first time I tried to do this. Got it working. PR submitted.

SolarLune:
This was merged in, so issue closed. Thanks!

PR #240, opened 2015-04-15
(#234) - Objects in group inherit instance's scale.

SolarLune:


GoranM:
It doesn't work when the root object has more than one child.

Also, I don't really see the rationale for this:

```java
for (GameObject c : g.children) {
	Vector3f offset = c.position().minus(g.position());
	c.position(g.position());
	c.move(offset);
}
```

Mathematically, it should not change c.position: 

```java
c = 2
g = 1
offset = c - g   // 2 - 1 = 1
c = g  // c = 1
c = c + offset  // c = 1 + 1 = 2
```

If you print c.position(), before and after, you'll see that it stays the same.

I suspect the visible changes are due to how the transform chain is updated, and how scale is incorporated in that process ... It's possible that there's bug in GameObject.scale, relating to how it affects children, even for non-instanced objects.

SolarLune:
Ah, you're right. 

Hm, there doesn't seem to be any scaling issues when not dealing with groups. It seems like it might be the process of spawning in the objects after the parent and then re-parenting them that houses the problem? I'm unsure, though.

GoranM:
You should be able to see the problem when actually setting a new scale, in game, for a given parent.

I've made a commit that should fix that - test it out.

SolarLune:
OK, I think I saw some weird behavior using the scale() function in-game that's not there now. Thanks for the fix - simply applying the scale seems to work now. PR should be updated.

Issue #236, opened 2015-04-11
Suggestion: GameObject.components.get() Component Class


SolarLune:
Hey, there. I think it'd be useful to have a get() function in the components list that allows you to get a component by a specific class (i.e. components.get(Halo)). The only way it seems to be possible to do this currently would be to store references in the GameObject and pull them from there, but that's not always possible (like in my current case).

GoranM:
> but that's not always possible (like in my current case).

What case, specifically?

SolarLune:
I have a PlayerControl component (that extends from Component) that exists in a separate file from the Player. Its GameObject reference doesn't point to the Player, but rather a GameObject that is the Player (so I can't access Player-specific code). Basically, "g" is hardwired to be a GameObject, not an instance of the Player class, and I'd rather not do some "hack" to create a Player specific variable in the PlayerControl component. 

I thought the idea was for components to be re-useable - it doesn't work as well if you program them to work with one kind of object.

Some sort of messaging system might also help with this.

GoranM:
> I thought the idea was for components to be re-useable

They are, for objects that can actually use them.

If PlayerControl requires access to Player specific fields/methods, it is a Player specific component - You can't expect to use it on objects that don't actually support the requirements.

> Basically, "g" is hardwired to be a GameObject, not an instance of the Player class, and I'd rather not do some "hack" to create a Player specific variable in the PlayerControl component.

I don't see how a "messaging" system, or your original suggestion, would actually resolve that issue ... Seems like you're jumping to a completely different topic here.

If that's really your core issue - having to cast to a specific type  - then maybe you should change the topic title to something more relevant.

That said: We could employ generics to have `Component<T extends GameObject>`, where T could be specified on extension, to make `g` the proper type.

SolarLune:
> If PlayerControl requires access to Player specific fields/methods, it is a Player specific component - You can't expect to use it on objects that don't actually support the requirements.

It doesn't. It requires data from another component. Thinking about it now, I can pass that other component as a reference to the PlayerControl component.

 >I don't see how a "messaging" system, or your original suggestion, would actually resolve that issue ... Seems like you're jumping to a completely different topic here.

[...]

PR #243, opened 2015-04-17
(#236) - components.get() function.

SolarLune:
New ArrayListClass class to allow for this.


GoranM:
I don't think you need a new container class: You could just have Component implement the Named interface, like GameObject does, and then you can just use ArrayListNamed.

SolarLune:
That makes sense, yeah. PR updated!

PR #250, opened 2015-04-17
(#76) profilerScene

rafcolson:
```Bdx.profiler.show()``` will show the profiler. By default this is added to ```Sacky.init()```. Let me know what you think. (Edit: you can put it anywhere, it's just an example)

Well, finally some results after using ```TimeUtils.nanoTime()``` instead of ```TimeUtils.millis()```. So much better now.

I tried but failed to activate ```show()``` when ```game_settings.show_framerate_profile``` is activated. Could this be stored in ```bdx``` on export to get it in the ```Profiler``` later on?

```percents``` and ```nanos``` are private now.

### CODE REVIEW

**Comment by SolarLune** [comment]
> Bdx.profiler.show() will show the profiler. By default this is added to Sacky.init(). Let me know what you think.

That doesn't sound like a good choice. What was the issue with using show() when the game settings' show framerate boolean is activated? If it doesn't work otherwise, storing it in the Profiler as a static field would probably be fine.

**Comment by rafcolson** [comment]
I stored a boolean in the Profiler, which is ```visible```. But of course some initialization has to occur too, that's why there is a method ```show()```. What doesn't sound good?

**Comment by SolarLune** [comment]
Putting it in a GameObject? Seems like it should be something that can't be interfered by the user. Seems like it should be where whether the Profiler scene is initialized, updated, and rendered is all up to the Show Framerate and Profile checkmark.

**Comment by rafcolson** [comment]
You can put the line to call ```show``` wherever you want. As an example I just put it in ```Sacky.init()``` so the profiler starts automatically in the default blend.

**Comment by SolarLune** [comment]
But why have the show() call at all? Why not store the "Show Framerate And Profile" status from Blender somewhere, and then call show() yourself from the Scene's init() function, or wherever else behind the scenes if you have it checked?

**Comment by rafcolson** [comment]
I understand. But for now if you can help me to get the Show Framerate and Profile value in Bdx, then I could make it so. But I couldn't access it.

**Comment by SolarLune** [comment]
Yeah, okay, sure. Gimme a sec.

**Comment by rafcolson** [comment]
Thanks a lot. :+1:

**Comment by SolarLune** [comment]
OK, so:

```python
"framerateAndProfile": scene.game_settings.show_framerate_profile,
```
Goes in exporter.py in the bdx dictionary around line 642. You can then get that boolean using

```java
json.get("framerateAndProfile").asBoolean()
```

in Scene.java; if you're trying to set it up elsewhere, then I imagine you'll have to make a new JsonReader to parse the current scene file, like the Scene class already does. Does that help?

SolarLune:
OK, so it seems like you should be able to just make a new field in the Bdx class and then write to it from the scene's init() function?

rafcolson:
I'm sorry Solar, I don't understand. Which scene's init() function do you mean?

You see, if I would get it through the JsonReader than I would have to parse the "current" scene file. But what is the "current" scene file?

```java
		json = new JsonReader().parse(scene); // do I have to get the active scene here?
		name = json.get("name").asString();
```

SolarLune:
I meant the Scene.init() function - the Scene class's init() [function](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Scene.java#L100).

rafcolson:
But do I just get the first scene of the scenes list then? What if you have an object in your scene, which calls an overlay and a background scene at initialization. Then I'll have to get the middle one. You see that I can't seem to find a clean way to do this.

You're welcome to give it a try yourself. Probably I'm being an amateur. But it's to confusing to me. I'd which all those global attributes where global as part of Bdx, and not of one of the active scenes.

SolarLune:
> What if you have an object in your scene, which calls an overlay and a background scene at initialization. 

Those scenes will be added later, on the next game frame; it doesn't happen immediately, AFAIK. On start, there's only one scene - the one that BDX loads up with. All other scenes that can be added are added after that one.

I'll check your code out.

SolarLune:
Hey, so here's a few slight changes I made.
https://github.com/SolarLune/bdx/commit/72f7cac2ec29b46de2fd2124a3840b77ba2a1919

Note that it's still not ideal in that the "Show Framerate and Profile" setting can be true in other scenes, so you can de-activate it in the current scene and end up with it still on because a scene added later has it enabled. Ideally, you would basically "deactivate" the profiler if the Framerate and Profiler button is deactivated (so no other Scene can start it up).

rafcolson:
PR updated. Thanks, SolarLune.

rafcolson:
You know what's funny? If you run the game once and then remove the ```__Profiler``` scene, it will still work because it is now saved as ```/assets/scenes/__Profiler.bdx```.

GoranM:
Yeah, that's how it works - As long as the bdx file is in the right place, along with all the resources it requires, the scene itself doesn't have to be present in the .blend

I don't think it should be too difficult to leverage that, so we don't have to include the profiler scene in the .blend, but this is ok for now.

Good work!

rafcolson:
@GoranM I'm glad you like it. But I think I forgot to push the last changes to the commit. The only change was the output of function ```Bdx.profiler.stop("myFunction")```. Currently, as it was merged, this will return the ```deltaTime``` in Long nanoseconds. But shouldn't I have to change this into Float milliseconds? I think this is an easier format for the user to work with. So let me know if you think it's a good idea.

SolarLune:
No problem!

PR #256, opened 2015-04-20
(#76) profiler scene adjustments

rafcolson:
- the profiler scene is now separated from the game.blend file into
profiler.blend (this also made it unnecessary to remove the profiler
scene from the game.blend after project creation

- if the profiler was not exported to *.bdx yet, then this will happen
when the game runs, by appending the scene from the default.blend and
removing it afterwards

Edit:

- profiler text font changed to [Liberation Mono Bold](http://en.wikipedia.org/wiki/Liberation_fonts), which is free. If you don't like the font, I could search for a better one.

- added ```avgTicRate```, which is the average per second.

- semi-transparent and shorter bars.


GoranM:
Why is adding a whole new .blend better than simply adding the line to remove the scene after project creation? I think it would be better to have an additional line, instead of an additional file.

A lot of the code seems unnecessary - All you really need to do is just add the scene, and then remove the scene. There's no need to check if the .bdx file exists, or anything else.

Also, there's no need to use the addon_utils module, because we already have the required functions in our own utils: You can get the path to the gen directory via `ut.gen_root()`.

rafcolson:
>Why is adding a whole new .blend better than simply adding the line to remove the scene after project creation? I think it would be better to have an additional line, instead of an additional file.

Maybe you are misunderstanding what I'm talking about here. This has nothing to do with appending and removing the scene from the ```game.blend``` in the addons directory.

You see, if the default ```game.blend``` in the addons directory would contain the profiler scene, then also the start-up ```game.blend``` in the project directory would contain it, because it's a duplicate, right? But this scene would be useless whatsoever, so it would be required to check if it exists and if so, remove it and save the blend. Ideally this should happen when the project is created, but it's only possible from within the start-up blend itself, when it's opened.

So the only way to remove it is when BDX is run. Unfortunately this would also require it to exist as long as BDX is not run (for the first time). So after project creation, when the start-up blend is opened, it would have this ```__Profiler``` scene in the Outliner, which is useless. And then after you started the game for the first time, it would be gone. Needless to say this would be a little confusing to the user, not to say "buggy" in the eyes of many.

So the most straightforward solution would be to not have the profiler scene in the default ```game.blend```, from which the start-up blend is created. Assuming that I'm right about all of this.

>A lot of the code seems unnecessary - All you really need to do is just add the scene, and then remove the scene. There's no need to check if the .bdx file exists, or anything else.

But wouldn't this extra code weigh up against the extra processing time for appending and removing the profiler scene every time BDX is run?

>Also, there's no need to use the addon_utils module, because we already have the required functions in our own utils: You can get the path to the gen directory via ut.gen_root().

I could have sworn this path is pointing to the project directory, which I thought was strange, but it clearly isn't as I just checked it again. Silly mistake.

GoranM:
I don't really understand what you mean.

It should be possible to simply remove the profiler scene during project creation (createproj.py), just before you save:

```python
        # remove profiler scene
        bpy.ops.wm.save_mainfile()
```

> But wouldn't this extra code weigh out against the extra processing time for appending and removing the profiler scene every time BDX is run?

Is the processing time noticeable? If not, there's no need to complicate the code with unnecessary optimizations.

As for the other stuff: You shouldn't put more into this PR than originally intended.

Put the other stuff in a new PR.

rafcolson:
>I don't really understand what you mean.
>
>It should be possible to simply remove the profiler scene during project creation (createproj.py), just before you save:
```java
      # remove profiler scene
        bpy.ops.wm.save_mainfile()
```

I had tried this before, but Blender will crash.

But wouldn't this extra code weigh out against the extra processing time for appending and removing the profiler scene every time BDX is run?
>Is the processing time noticeable? If not, there's no need to complicate the code with unnecessary optimizations.

Okay, I changed it. But I ran into a problem. When BDX runs for the first time it doesn't fail. But after that it will when it clones the text object to add it to the profiler scene. I don't know why, so I'm leaving it as it is. Shortly said: if you remove (or overwrite) ```__Profiler.bdx``` it will give this exception:
```
Exception in thread "LWJGL Application" java.lang.NullPointerException
        at com.nilunder.bdx.Scene.clone(Scene.java:341)
        at com.nilunder.bdx.Scene.add(Scene.java:398)
        at com.nilunder.bdx.Scene.add(Scene.java:414)
        at com.nilunder.bdx.utils.Profiler.show(Profiler.java:40)
        at com.nilunder.bdx.Scene.init(Scene.java:139)
        at com.nilunder.bdx.Bdx$ArrayListScenes.add(Bdx.java:36)
        at com.raco.game.BdxApp.create(BdxApp.java:18)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplic
ation.java:143)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplicati
on.java:120)
```

>As for the other stuff: You shouldn't put more into this PR than originally intended.
>
>Put the other stuff in a new PR.
Okay.

PR #257, opened 2015-04-20
(#76) profiler scene internal

rafcolson:
As was mentioned [here](https://github.com/GoranM/bdx/pull/256#issuecomment-94565183) there are two issues to be tackled:

- removing (or overwriting) ```__Profiler.bdx``` will give an exception. So the alternative is to export to *.bdx only when the game runs for the first time after project creation.

- as been demonstrated below, Blender will crash at project creation when removing the redundant ```__Profiler``` scene from the start-up ```game.blend```. So the alternative would be to separate the ```__Profiler``` scene from the default ```game.blend``` into a second ```profiler.blend```, from which the scene can be appended.

@GoranM I can't seem to find another solution, so if you would have a look at it, it's well appreciated.


GoranM:
> removing (or overwriting) __Profiler.bdx will give an exception. So the alternative is to export to *.bdx only when the game runs for the first time after project creation.

You have: `prof_blend_name = "default.blend"` - it should be `"game.blend"`, right?

That said: It seems like removing the scene from the .blend doesn't unlink all the data immediately, so when the scene is added again, the original names that were used for objects in that scene are considered unavailable, so blender gives all objects (and even fonts) new names.

Given that, I would agree that checks are probably the best option, however, instead of `profiler_on_exprun`, you should use a more descriptive name, like `profiler_exported`.

> as been demonstrated below, Blender will crash at project creation when removing the redundant __Profiler scene from the start-up game.blend. So the alternative would be to separate the __Profiler scene from the default game.blend into a second profiler.blend, from which the scene can be appended.

There seems to be some issue with `bpy.data.scenes.remove(scene)`, when called in the same context that opened a new file.

We could do this:

```python
        @bpy.app.handlers.persistent
        def remove_prof_scene(dummy):
            bpy.data.scenes.remove(bpy.data.scenes["__Profiler"])
            bpy.app.handlers.load_post.remove(remove_prof_scene)
        bpy.app.handlers.load_post.append(remove_prof_scene)

        if ut.in_packed_bdx_blend():
        # ...
```

But, considering the other problem, I would agree that it's probably best to just have a separate `profiler.blend`.

rafcolson:
PR updated. Thanks for having a look at it.

PR #266, opened 2015-04-24
(#76) profiler improvements

rafcolson:
New init() function:
- proper initialization of variables

New remove() function:
- profiling should also be possible with objects while adding and removing them to the scene; calling this function ```onEnd()``` will ensure that.

Display changes:
- smaller space between subjects (tic info - internal subsystems - user
created subsystems)
- a bit smaller in size (almost 90% of the original)

Profiler Blend changes:
- Bar to a square one by one plane to give the Profiler class full
control over its dimensions

General changes (vars: spacing, TIC_RATE, exceptionMessage, ...)


GoranM:
bdx.zip should not be in the repo

BdxApp.java should not have any initialization code - If you need to initialize something, outside the Profiler constructor, you can do it in Bdx.init().

We didn't agree on having a small/large option for the profiler.

rafcolson:
Thanks, I didn't notice that BDX.zip was in it. Ai ai ai.

So PR updated.

GoranM:
It's better if BdxApp methods simply call on the relevant Bdx methods, and nothing else, because that defines a clear border between BDX code and libgdx application frame.

Just set visible in the scene, like you did before.

rafcolson:
Okay. But that would mean I ll have to remove initialization for the Profiler as it will have check every scene for "framerateProfile".

GoranM:
You could pass the boolean to init, and use an initialized flag to avoid subsequent runs:

```java
public void init(boolean framerateProfile){
    if (initialized)
        return;
    // init code
    visible = framerateProfile;
    initialized = true
}
```

rafcolson:
Okay, I understand you are basing this on the fact that the initial scene will also be the first scene that will run the ```init()```. I wasn't sure about this until now. Thanks. I should manage now.

GoranM:
The initial scene calls profiler init, and profiler init calls `new Scene("__Profiler")`, which calls profiler init.

That should explain the behavior you observed.

It's not something that should cause problems, if you simply do this:

```java
public void init(boolean framerateProfile){
    if (initialized)
        return;
    visible = framerateProfile;
    initialized = true
    // init code
}
```

With that, starting scene calls profiler init, which calls `new Scene("Profiler")`, which will also call profiler init, but since initialized is true at that point, it will just return, and the same with all subsequent calls.

About the PR:

It's ok for all of this to be in the same PR, but you need to have different commits for different changes (one for init method, one for remove method, one for display changes, etc).

rafcolson:
>The initial scene calls profiler init, and profiler init calls new Scene("__Profiler"), which calls profiler init.

>That should explain the behavior you observed.

You're right. That explains it. Pretty dumb of me.

>It's not something that should cause problems, if you simply do this: [...]

>With that, starting scene calls profiler init, which calls new Scene("Profiler"), which will also call profiler init, but since initialized is true at that point, it will just return, and the same with all subsequent calls.

I finally get what I was doing wrong. ```if (visible) show();``` had to be beneath ```initialized = true;``` of course. If not  ```initialized``` would still be false when the profiler scene is added by ```show()```.

Sorry again for probably wasting your time with my incompetence.

PR #269, opened 2015-04-27
(#76) Profiler improvements

rafcolson:
New init() function:
- proper initialization of variables

New remove() function:
- profiling should also be possible with objects while adding and removing them to the scene; calling this function ```onEnd()``` will ensure that.

Display changes:
- smaller space between subjects (tic info - internal subsystems - user
created subsystems)
- a bit smaller in size (almost 90% of the original)

Profiler Blend changes:
- Bar to a square one by one plane to give the Profiler class full
control over its dimensions

General changes (vars: spacing, TIC_RATE, exceptionMessage, ...)


GoranM:
Why are you passing the whole scene to init?

It seems like it would be better to just pass the framerateProfile boolean, because that's the only thing you actually need, and it helps to make that clear.

rafcolson:
Done.

Issue #275, opened 2015-04-29
Suggestion: changes to Scene (adding & removing)


rafcolson:
>You should be able to "end" a scene by just removing it from the scenes list, right? No need for a specific function? Something to end the game sounds good, though.

Yes, it seems like it. But just for convenience sake I could add ```remove(String sceneName)```, because currently you have to get the scene or give the index: ```remove(int index)``` and ```remove(Scene scene)```.

Wouldn't it also be nice to have ```add(String sceneName)``` instead of ```add(new Scene(String sceneName))```?

In addition to that it would be really convenient to have ```add()``` return the scene itself.

Oh, right, also ```add(int index, String sceneName)```.

What do you guys think?

GoranM:
The scenes list can have multiple scenes that share the same name, so `remove(String sceneName)` doesn't really make sense, because there's no way to specify which scene to actually remove in that case.

We discussed add(String) in #148.

I don't think the gains (in terms of convenience) are worth the additional methods, but if you really want it, I'll merge it.

It's your decision.

rafcolson:
@GoranM 

When looking at ```add(int index, Scene scene)``` I noticed that it returns a boolean.
```java
		public boolean add(Scene scene){
			boolean ret = super.add(scene);
			if (scene.objects == null)
				scene.init();
			return ret;
		}
```
Could this return the scene instead? I tried but it really can only return a boolean. Is there no way to replace ```add```? Because returning the scene works when for example changing it to ```addScene```.

The same with ```add(int index, Scene scene)```. This can only return ```void```, so it seems.

Also ```set()``` is modified to be able to return the "old" scene. Could you maybe explain why it's like that?
```java
		public Scene set(int index, Scene scene){
			Scene old = remove(index);
			add(index, scene);
			return old;
		}
```
>because there's no way to specify which scene to actually remove in that case.

Thinking about it, wouldn't it be only the case if the same scene is added more than once? Maybe that would be useful in some cases, but it would not be regular practice. Doing this also requires to store the scene reference somewhere outside Bdx.scenes in order to have access to it.

GoranM:
> Could this return the scene instead?

Why would you want/need that? You already have the relevant scene reference.

> The same with `add(int index, Scene scene)`. This can only return void, so it seems.

Don't you mean `boolean`? You're referring to the same method as above.

> Also `set()` is modified to be able to return the "old" scene. Could you maybe explain why it's like that?

[...]

PR #278, opened 2015-05-02
(#275) add, set, remove scene by name and return scene

rafcolson:


GoranM:
Shouldn't set create a new scene, and place that into given position?

rafcolson:
@GoranM Yes, but it already does. Let's say you want to replace the HUD with a Menu, then you would do this, for example:
```Java
Bdx.scenes.set(1, new Scene("Menu"));
```
You created a new scene, named "Menu", and added it to the list at the given position.

The built-in method ```set(int index, E element)``` (with E being Scene in this case) is exactly doing the latter. It **"Replaces the element at the specified position in this list with the specified element."**

So I think ```ArrayListScenes.set(int index, Scene scene)``` could be omitted. If you think so too, I can remove the code in this commit.

GoranM:
Unlike ArrayListScenes.set, the built-in set method doesn't actually call add, so it's not going to initialize the scene as we require. 

Anyway, my question relates to the set method that you added (`set(int, String)`), not the one that was already there.

rafcolson:
Sorry, I totally missed those things. Thanks. PR updated.

rafcolson:
@GoranM Maybe better to first merge this one so I can rebase ```replaceModel()```.

GoranM:
The commits don't depend on each other, so there's no need to rebase.

Issue #279, opened 2015-05-03
Suggestion: GameObject.replaceModel()


rafcolson:
Well, I think it would be nice to be able to replace the mesh of the object. Similar to the BGE: ```replaceMesh(String meshName, boolean useDisplayMesh, boolean usePhysicsMesh)``` and ```replaceMesh(String meshName)``` with ```useDisplayMesh = true``` and ```usePhysicsMesh = false``` as the default.

Only difference is that the mesh is replaced by a unique mesh. So we're not replacing one of the ```models```, we're adding a new model, being a copy of the specified model.

What do you think? Any suggestions? Improvements?

GoranM:
We don't actually have a Mesh concept in BDX - We use libgdx models, and representative model instances, so the name "replaceMesh" doesn't really cover existing semantics. Something like "replaceModel" would make more sense, in my view.

That said: This seems like it would be a useful feature, so you can make the PR, and I'll review the code there.

rafcolson:
@GoranM 
@SolarLune 

I think you're right. A ```Model``` is described as an owner of meshes, materials and textures, as been documented [here](http://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/g3d/Model.html). And currently ```replaceMesh()``` is replacing the entire Model.

Maybe we could change ```useUniqueMesh()``` to ```useUniqueModel()``` or ```cloneModel``` too? I'll make a separate PR for this and I'll try to replace ```mesh``` with ```model``` were needed. But first let me know what you think.

GoranM:
Yes, "useUniqueModel" would probably make more sense.

If you're going to change it, be sure to do it in all places where it's used.

rafcolson:
Okay, I'll do.

rafcolson:
PR's merged. Thanks!

PR #280, opened 2015-05-03
(#279) replaceModel (and angularVelocity)

rafcolson:
I thought it was cleaner to first add a ```angularVelocity()``` method. In order to maintain physics velocities it was required to get both linear and angular velocity first, before destroying the body.

I was not very sure about ```body.updateInertiaTensor()```, so if you could let me know if this line could be removed or not, that would be nice.


GoranM:
I'm not really sure if `updateInertiaTensor` is required; It's something best confirmed through testing, or if you don't feel like doing that, it's ok to just leave it as is.

Issue #277, opened 2015-04-30
Does BDX support Java 1.8?


skypatcher:
Hi:

My System is Win 8.1 (Intel 64 bit). The Android SDK is at C:\AndroidSdk. I just installed the BDX plugin 0.1.8 to Blender 2.74. When I click 'Create BDX project'. I got the following message in Blender's system console:
![bdx-error](https://cloud.githubusercontent.com/assets/6044891/7420481/d2794d0a-ef4a-11e4-8e1c-03c5a178d75b.png)

SolarLune:
BDX does support Java 1.8 - it's what I'm running on my machine. Do you only have Java 1.8 installed? Can you make a project without the Android SDK? What happens if you type `java -version` into a command prompt?

rafcolson:
JDK is installed, but not JRE?

GoranM:
Assuming that `javac -version` prints something higher than or equal to 1.7, it should work.

Looking at your error, I'm reminded of [this post](https://github.com/GoranM/bdx/issues/182#issuecomment-81018425). As you can read there, it seems like the problem stems from having both java 7 and java 8 installed.

You can confirm that by running this python code:

```python
import subprocess
subprocess.check_call(["javac", "-version"])
```

If that prints 1.7, while `javac -version` prints 1.8, that's the problem.

skypatcher:
Thanks for everyone's suggestion. 

When I type java -version, it tells me 1.8, but when I type javac -version, it said unrecognized command.  Turns out JAVA_HOME was pointing to an old JDK folder (Java 7, which I have already uninstalled a while back).

After I fix the JAVA_HOME, I can now get version 1.8.0_31 from both java -version, and javac -version prompt.

But when I restart Blender and create a new project, I get the following error. I saw something about -source 1.6. Don't know if it is because my JDK installation is still corrupted?

![bdx-1](https://cloud.githubusercontent.com/assets/6044891/7433023/30bf96de-effb-11e4-89f0-da2e2afcec8f.png)

![bdx-2](https://cloud.githubusercontent.com/assets/6044891/7433031/3aca91b0-effb-11e4-8016-1af5406ffe21.png)

rafcolson:
I think you didn't specify the correct SDK path. On my system for example it's:  ```C:\Users\Acer\AppData\Local\Android\sdk```. Something like ```\android-sdk``` to me doesn't seem valid.

skypatcher:
[...]

PR #288, opened 2015-05-05
(#277) SDK buildtools sort dir fix

rafcolson:
It's not perfect, but at least if in the ```sdk\build-tools``` folder there are directory names that differ from the regular format (e.g. ```20.0.0```), it will put these on the bottom of the list. Only if all directory names differ from the regular format I expect the ```build.gradle``` not to be set properly. I could make another exception for when the name contains ```build-tools```, but I have to be sure that ```build-tools-21.1.2``` is a name produced by the SDK Manager.

Any advise is appreciated.


rafcolson:
@GoranM I'll leave it up to you if you think it's necessary to merge this as @skypatcher explains there are no problems with current SDK versions.

GoranM:
If the old code works fine, with the latest sdk (seems to be the case), there's no need for this, so I think it's best if we close for now.

If we encounter similar issues in the future, we can revisit this.

rafcolson:
Okay, thanks.

Issue #289, opened 2015-05-06
Bug: parent(null)


rafcolson:
Currently, ```parent(null)```
- breaks NO_COLLISION. (fixed)
- breaks RIGID_BODY: for example, an unparented cube that falls down will bounce, but keeps rotating slowly in one random direction, until it gets stationary. Edit: it turns dynamic only.

GoranM:
> breaks RIGID_BODY: for example, an unparented cube that falls down will bounce, but keeps rotating slowly in one random direction, until it gets stationary.

I can't seem to reproduce this. Can you make a simple example?

rafcolson:
Here's an [example](https://www.dropbox.com/s/nd6u69j2sfhmksg/physics_unparent_rigidbody_bug.blend?dl=0).

I think the reason why in my other file the box momentarily turned was because it collided at the moment it was unparented. In this example it's just a cube falling down and you'll notice it turns dynamic (only).

GoranM:
I just see a box that falls to the ground ... It doesn't turn.

Are you on the master branch? Did you recompile?

rafcolson:
>I just see a box that falls to the ground ... It doesn't turn.

That's exactly the problem. I'm working on a fix right now.

>Are you on the master branch? Did you recompile?

Yes.

rafcolson:
PR submitted.

rafcolson:
PR merged.

PR #290, opened 2015-05-06
(#289) parent(null) fix for NO_COLLISION

rafcolson:


GoranM:
It might be better to modify the dynamic method:

```java
	public void dynamic(boolean dynamic){
		if (!body.isInWorld())
			return;
		...
	}
```

rafcolson:
I wouldn't do that. Because then you loose the option to make a NO_COLLISION object DYNAMIC. By modifying the ```parent()``` method, that option would be preserved.

Issue #300, opened 2015-05-09
Bug: Scale() doesn't activate body.


rafcolson:
Let's say you want to scale a (sleeping) cube with Dynamic body type to half its size, while resting on a surface. Then it will stay a sleep and float above the surface.

I'll look into to this. It should be a quick fix.

rafcolson:
PR submitted.

rafcolson:
PR merged.

PR #302, opened 2015-05-09
(#300) Bug: scale() now activates body

rafcolson:


GoranM:
Couldn't you simply simply add `body.activate(true)` to the beginning of the scale method?

rafcolson:
>Couldn't you simply simply add ```body.activate(true)``` to the beginning of the scale method?

Yes. But there's no need to activate the body if it's a child, right? And the method already checks if it's a child so better take advantage of that. The same goes for the ```transform``` method. 

But if you want me to just add ```body.activate(true)```, I don't mind changing it. I don't think redundant activation would cause any problems.

GoranM:
I'm not exactly sure on the complete bullet requirements, but I've experienced bugs when moving static objects without activating them first.

So, I think it's best to just add `body.activate(true)`.

rafcolson:
Okay. Done.

Issue #306, opened 2015-05-10
Suggestion: GameObject.mass setter and getter


rafcolson:
Getter and setter for mass, with private variables ```mass``` and ```inertia```. I've added inertia so this wouldn't have to be calculated with every body change.

GoranM:
If you can simply recalculate inertia, that would be better than having an additional field on GameObject. Same with mass: If you can get the current mass from the body, you should just do that, instead of having a redundant field.

Also, is there a situation in which you want to change mass, but not calculate local inertia? If not, there's no need for an additional method - You can just have `mass(float)` which will always calculate local intertia.

rafcolson:
Oh, yes, you are right. I missed that! I could remove ```inertia``` then. But I'll better keep ```mass``` for this will become necessary later on. We like to change mass and preserve the value without changing the json value, don't we? We can't always access the body for this, as sometimes objects will lose their mass when parented to other objects. When unparenting we have to get the old mass somewhere.

GoranM:
I didn't say that we should change the json value - json data should be treated as read only.

For the mass getter, it should be possible to use the information already on the body, so, in general, that shouldn't require a mass field on GameObject.

For situations where you need to remember the last mass, it would make sense to have a `lastNonZeroMass` field.

rafcolson:
PR submitted.

rafcolson:
Thanks for the merge!

PR #307, opened 2015-05-10
(#306) GameObject.mass setter and getter

rafcolson:


GoranM:
It may be better to call the field `lastNonZeroMass`, to hint that we don't actually use this to store current mass.

rafcolson:
Okay. Done. I also could reduce some lines by using the mass setter.

GoranM:
The commit name doesn't make sense:

You're not "implementing lastNonZeroMass - That was simply a field you added as part of the setter getter commit.

The second commit should be named something that actually describes the changes.

In this case: "GameObject.dynamic now uses newly added mass methods"

rafcolson:
Done.

PR #312, opened 2015-05-11
Builtin screen filters

SolarLune:


GoranM:
There seem to be some formatting problems with your code (comments are misaligned).

Also, it might be better to have all the shaders in `gen/filters`.

SolarLune:
OK, I'll create another PR.

Issue #305, opened 2015-05-09
Suggestion: GameObject.replaceBodyType()


rafcolson:
Currently there's only ```dynamic(boolean)``` to change body types. And it isn't exactly perfect. Because if you would want to change RIGID_BODY to DYNAMIC, this method would do the trick. Actually, this method could be better described by ```suspendDynamics()``` and ```restoreDynamics()```, which I think should be private methods.

Then there could be methods to actually change the body type:
-```noCollision(boolean) to switch between "NO_COLLISION" and "STATIC"
- ```dynamic(boolean)``` (to replace the old one) to switch between "STATIC" and "DYNAMIC"
- ```rigidBody(boolean)``` to switch between "DYNAMIC" and "RIGID_BODY"

Of course "SENSOR" should be treated as "STATIC" and so I propose to change this in a new ```bodyType``` variable at initialization, which will be derived from ```json.get("physics").get("body_type")```. Also I think it would be better to change to small caps.

These three methods would be mere convenience functions. I think we would benefit more from one method to replace the body type, so: ```replaceBodyType(String)```.

GoranM:
A `bodyType(String)` would be fine. With that, there should be no need for the other methods you mentioned.

> Of course "SENSOR" should be treated as "STATIC" and so I propose to change this in a new bodyType variable at initialization, which will be derived from json.get("physics").get("body_type"). Also I think it would be better to change to small caps.

Well, 'SENSOR' indicates a 'STATIC' object set to ghost, so it's a relevant distinction. In general, I think we should store blender data as exported, and then use it in BDX according to our actual needs.

It also makes sense to keep blender notation for body types, which is capital case - It also hints that those strings are used as flags, rather than just arbitrary data.

rafcolson:
Alright, so I'll have a go at this. I'll submit a PR after #306 and another PR that would add support for joined masses with compound objects. It would be nice if this first got in so I immediately could use the ```mass()``` method in some applications. To be honest, I already have a rough commit with changes to mass, dynamics and parent.

rafcolson:
PR submitted.

rafcolson:
I'll have another go at this and hope I'll get it right the first time.

rafcolson:
PR submitted.

rafcolson:
PR merged.

PR #314, opened 2015-05-14
(#305) GameObject.bodyType() and Bullet update for mass and ghost

rafcolson:
I split this up into several commits, just to keep it organized. Bullet needed some updates with mass and ghost, so I hope it's okay to have this in this PR too.

Notice that also the ```dynamic()``` methods are removed since there's no use for those anymore.

Any suggestions/remarks are welcome.

Edit: Now ```parent(null)``` doesn't just convert the child to "dynamic", but to the last body type.


GoranM:
I don't think there's any reason for Bullet.makeBody to take additional arguments - Why can't you simply let it make the body from json data, as before?

I think we should change the implementation of dynamic, not remove it, so it would probably be better to first "Add bodyType getter and setter", and then "Change implementation of dynamic() to use bodyType()".

With that, the last commit would be unnecessary, because dynamic(true) would go back to the last proper mass and body type.

rafcolson:
>I don't think there's any reason for Bullet.makeBody to take additional arguments - Why can't you simply let it make the body from json data, as before?

```Bullet.makeBody()``` is not only used for generating templates, but also for replacing models. See: ```GameObject.replaceModel()```. The ```ghost``` state, the ```mass``` and/or the ```bodyType``` of an object could be changed. That's why it is necessary to have additional arguments for these three attributes when replacing the model, because ```makeBody()``` needs those. Else, replacing the model would discard all changes to ghost, mass or bodyType.

>I think we should change the implementation of dynamic, not remove it, so it would probably be better to first "Add bodyType getter and setter", and then "Change implementation of dynamic() to use bodyType()".

The reason why I completely removed it is because privately this method would only be useful for the ```parent()```method. It also would be a mistake to consider the initial ```dynamic()``` method to be as was described in the API:
><a name="dynamic-bool" href="#dynamic-bool">#</a> **dynamic**(boolean dynamic)
>
>Set whether the object is dynamic or not.

It didn't set whether the object is dynamic or not. This would be true however, if also the ```angularFactor()``` would be set to zero. A better description of what this method does is "suspend or restore dynamics".

For public use this could be convenience methods, but I strongly suggest against ```dynamic(boolean)```, so it wouldn't be confused with actually setting the body type to "DYNAMIC". I think ```suspendDynamics()``` and ```restoreDynamics()``` would be more descriptive.

And I think we can leave out the getter too, as ```bodyType().equals("DYNAMIC")``` does the same thing.

><a name="dynamic" href="#dynamic">#</a> **dynamic**() - boolean
>
>Whether this object is dynamic or not.

Finally, removing the parent with the ```parent(null)``` method doesn't always require to restore dynamics on the child. If doing so with "NO_COLLISION", "STATIC" or "SENSOR" objects, this would cause problems, especially with "NO_COLLISION". (With the other two mass it would be a null operation). That's why I thought setting to last body type would be more straightforward. I'm also thinking on how potential other people would read the code and try to figure out what actually needs to happen with the child when the parent is removed.

>With that, the last commit would be unnecessary, because dynamic(true) would go back to the last proper mass and body type.

Is it okay with you that I replace ```dynamic(boolean)``` with methods ```suspendDynamics()``` and ```restoreDynamics()```? And should these be public?

GoranM:
Couldn't you simply set ghost, mass, and bodyType on the GameObject itself (where that is necessary)? It seems like doing that would be cleaner than stuffing makeBody with additional arguments.

> Is it okay with you that I replace dynamic(boolean) with methods suspendDynamics() and restoreDynamics()? And should these be public?

Yes.

rafcolson:
>Couldn't you simply set ghost, mass, and bodyType on the GameObject itself (where that is necessary)? It seems like doing that would be cleaner than stuffing makeBody with additional arguments.

I don't understand why you say it would be cleaner. I think it's just the opposite. Because ```makeBody()``` uses ```mass``` to create ```RigidBodyConstructionInfo``` and ```ghost``` and ```bodyType``` to set ```CollisionFlags```. So if ```ghost```, ```mass``` and ```bodyType``` are also set "outside" ```makeBody()``` it would be doing things twice, which is useless.

But how about instead of these arguments passing the GameObject as an extra argument? Would that be a better idea? (Note that the mesh has to stay as an argument for the templates).

Issue #317, opened 2015-05-15
Bug:While switching dynamics, mass is only restored when set manually


moshme:
When i have an GameObject with dynamics and then turn them off and afterwards on again, i loose the mass of the object. As long as i see, it is because the variable lastNonZeroMass of the GameObject class is only set correctly when you invoke the mass setter method manually. But not on creation of the object.

rafcolson:
This is a bug. The ```lastNonZeroMass``` is not set correctly at Game creation, because the variable is not set in ```cloneBody()```. It would be fixed with the ```bodyType``` PR, but maybe it could take some time for this to get merged. So here's the fix:

rafcolson:
PR submitted.

rafcolson:
PR merged.

PR #318, opened 2015-05-15
(#317) Bugfix for clone() with lastNonZeroMass

rafcolson:


GoranM:
The lastNonZeroMass field was created specifically for the dynamic method, to serve as a temporary storage variable. It's not a part of the actual physics state, so it doesn't need to be copied, and it shouldn't be used to store current mass.

If you need to get the current mass, in cloneBody, you should be able to simply use the mass getter:

```java
float mass = gobj.mass();
```

rafcolson:
Okay, done.

GoranM:
You're still copying lastNonZeroMass.

As I said: The lastNonZeroMass field was created specifically for the dynamic method, to serve as a temporary storage variable. It's not a part of the actual physics state, so it doesn't need to be copied.

rafcolson:
If ```lastNonZeroMass``` is not set in ```cloneNoChildren()``` it will be ```null```. Imagine you would suspend dynamics on any object (because every object is a clone of a template), how will you get the ```lastNonZeroMass``` of that object in order to restore dynamics?

GoranM:
The dynamic method should set mass to 0 via a `gobj.mass(0)` call (which will set lastNonZeroMass), not `body.setMassProps(0, new Vector3f())`.

rafcolson:
By setting ```lastNonZeroMass``` to zero when suspending dynamics, you will never be able to restore dynamics for that same object, because you now erased the ```lastNonZeroMass```. Let me show you:

- suspending dynamics on an object with a mass of ```1```:
```java
mass(0); //sets lastNonZeroMass (which was 1) to 0
```
- restoring dynamics
```
...
mass(lastNonZeroMass); //sets lastNonZeroMass to 0 because lastNonZeroMass is 0
...
```
Result: both suspending and restoring dynamics will suspend dynamics.

GoranM:
Oh, right ... I confused myself there.

Anyway: I don't think that the mass setter should set lastNonZeroMass, because lastNonZeroMass is solely for the dynamic method to use. As I said in my first post: it shouldn't be used to store current mass.

So, in the dynamic method, it would probably be best to do the following, on suspension:

```java
lastNonZeroMass = mass();
mass(0);
```

Right?

rafcolson:
Yes, I was also thinking this. This is better.

PR #324, opened 2015-05-21
Frustum Culling added

moshme:
The vectors needed in the function are created on object creation once and not every function call to do not stress the garbage collector to much, because the function is called every frame..

### DIFF
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -183,7 +183,7 @@ public class Bdx{
 			Gdx.gl.glClear(GL20.GL_DEPTH_BUFFER_BIT);
 			modelBatch.begin(scene.cam);
 			for (GameObject g : scene.objects){
-				if (g.visible()){
+				if (g.visible() && g.insideFrustum()){
 					modelBatch.render(g.modelInstance, scene.environment);
 				}
 			}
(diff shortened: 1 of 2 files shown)


GoranM:
It seems that you added an empty file called "git". I assume this was by accident. If so, please remove it. If not, please explain why it was added.

Vector3f methods like plus and minus allocate new vectors ... In general, I don't think we should worry about "stressing the garbage collector", until that actually becomes a problem. So, for now, it would be ok to simply allocate the required vectors locally, in the method.

The comment seems unnecessary; The method name is pretty clear.

rafcolson:
Also note that ```Vector3f(0, 0, 0)``` could simply be ```Vector3f()```. See [docs](http://download.java.net/media/java3d/javadoc/1.5.1/javax/vecmath/Vector3f.html#Vector3f()).

moshme:
Ok i have added the proposed changes.

The reason i tried to avoid allocating vectors, is because developping on android im need to be interested in Performance. And it seems that this a common way to optimize things on mobile. As could be read [here](http://developer.android.com/training/articles/perf-tips.html#ObjectCreation) in under the header "Avoid Creating Unnecessary Objects". 

Also if have seen this practice done by Bullet or Libgdx.

GoranM:
Please make it a single commit.

> The reason i tried to avoid allocating vectors, is because i am developping on android i am interested in Performance.

I understand, but as I said, until it actually becomes a problem, I don't think we should worry about it; For the most part, the performance bottleneck on android is in the rendering subsystem.

rafcolson:
@moshme A little tip: I'm not sure if it will be noticeable to others that you updated your PR. So if your work is finished, you could always write a new message to say the PR was updated. Cheers.

GoranM:
Thanks.

PR #325, opened 2015-05-23
(#305) GameObject.bodyType & GameObject.dynamics

rafcolson:
Overview:
- removal of redundant mass settings
- replaceModel update for ghost and mass
- dynamic -> dynamics & various fixes (details in description)
- GameObject.bodyType getter and setter (details in description)

I hope I documented this enough so it will be easier to overlook the code. If something isn't clear, let me know.

Edit: The ```bodyType``` variable is public because it has to be set in ```Scene.init()```. Would it be better to change the name to something less common? ```bodyTypeCurrent``` perhaps?

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -34,6 +34,7 @@ public class GameObject implements Named{
 	public ArrayList<PersistentManifold> contactManifolds;
 	public ModelInstance modelInstance;
 	public RigidBody body;
+	public String currBodyType;
 	
 	public HashMap<String, JsonValue> props;
 	
@@ -49,7 +50,7 @@ public class GameObject implements Named{
 	private boolean visible;
 	private boolean valid;
 	private Model uniqueModel;
-	private float lastNonZeroMass;
+	private String lastBodyType;
 
 	
 	public GameObject() {
@@ -59,6 +60,7 @@ public class GameObject implements Named{
 		components = new ArrayListNamed<Component>();
 		children = new ArrayListNamed<GameObject>();
 		valid = true;
+		lastBodyType = "DYNAMIC";
 	}
 
 
@@ -100,9 +102,9 @@ public class GameObject implements Named{
 			if (parent.compoundShape() != null)
 				parent.compoundShape().addChildShape(new Transform(localTransform), body.getCollisionShape());
 
-			dynamic(false);
-		}else if (!json.get("physics").get("body_type").asString().equals("NO_COLLISION")){
-			dynamic(true);
+			dynamics(false);
+		}else{
+			dynamics(true);
 		}
 		
 	}
@@ -638,19 +640,12 @@ public class GameObject implements Named{
 			modelInstance = mi;
 		}
 		if (updatePhysics && body.isInWorld()){
-			Vector3f vel = velocity();
-			Vector3f angVel = angularVelocity();
+			String boundsType = json.get("physics").get("bounds_type").asString();
+			boolean compound = json.get("physics").get("compound").asBoolean();
 			scene.world.removeRigidBody(body);
-			body.destroy();
-			JsonValue physics = json.get("physics");
-			body = Bullet.makeBody(model.meshes.first(), trans.getValues(), physics);
-			body.setUserPointer(this);
-			mass(lastNonZeroMass);
+			body.setCollisionShape(Bullet.makeShape(model.meshes.first(), boundsType, compound));
+			mass(mass());
 			body.updateInertiaTensor();
-			if (dynamic()){
-				velocity(vel);
-				angularVelocity(angVel);
-			}
 			scene.world.addRigidBody(body);
 			scene.world.updateSingleAabb(body);
 		}
@@ -666,20 +661,20 @@ public class GameObject implements Named{
 
 	}
 
-	public void dynamic(boolean dynamic){
-		if (dynamic){
-			scene.world.removeRigidBody(body);
-			mass(lastNonZeroMass);
-			scene.world.addRigidBody(body);
-			body.activate();
+	public void dynamics(boolean enable){
+		if (currBodyType.equals("NO_COLLISION") || enable && dynamics() || !enable && !dynamics()){
+			return;
+		}
+		if (enable){
+			bodyType(lastBodyType);
 		}else{
-			lastNonZeroMass = mass();
-			mass(0);
+			lastBodyType = currBodyType;
+			bodyType("STATIC");
 		}
 	}
 
-	public boolean dynamic(){
-		return !body.isStaticOrKinematicObject();
+	public boolean dynamics(){
+		return !body.isKinematicObject();
 	}
 	
 	public float mass(){
@@ -687,10 +682,61 @@ public class GameObject implements Named{
 	}
 	
 	public void mass(float mass){
+		if (mass == 0){
+			throw new RuntimeException("no zero value allowed: use 'dynamics(false)' instead");
+		}
 		Vector3f inertia = new Vector3f();
 		body.getCollisionShape().calculateLocalInertia(mass, inertia);
 		body.setMassProps(mass, inertia);
 	}
+	
+	public String bodyType(){
+		return currBodyType;
+	}
+	
+	public void bodyType(String s){
+		int flags = body.getCollisionFlags();
+		if (s.equals("NO_COLLISION")){
+			scene.world.removeRigidBody(body);
+			flags &= ~CollisionFlags.KINEMATIC_OBJECT;
+		}else{
+			if (s.equals("STATIC")){
+				flags |= CollisionFlags.KINEMATIC_OBJECT;
+			}else if (s.equals("SENSOR")){
+				flags |= CollisionFlags.KINEMATIC_OBJECT;
+				flags |= CollisionFlags.NO_CONTACT_RESPONSE;
+			}else{
+				
+				// kinematic initialization hack
+				if (mass() == Float.POSITIVE_INFINITY){
+					mass(1); // Blender default
+					scene.world.removeRigidBody(body);
+					flags &= ~CollisionFlags.KINEMATIC_OBJECT;
+					body.setCollisionFlags(flags);
+					scene.world.addRigidBody(body);
+				}
+				
+				flags &= ~CollisionFlags.KINEMATIC_OBJECT;
+				if (s.equals("DYNAMIC")){
+					body.setAngularVelocity(new Vector3f());
+					body.setAngularFactor(0);
+				}else if (s.equals("RIGID_BODY")){
+					body.setAngularFactor(1);
+				}else{
+					throw new RuntimeException(s + " is no valid bodyType name.");
+				}
+			}
+			if (!body.isInWorld()){
+				body.clearForces();
+				body.setLinearVelocity(new Vector3f());
+				scene.world.addRigidBody(body);
+			}
+			body.activate(true);
+		}
+		body.setCollisionFlags(flags);
+		currBodyType = s;
+	}
+
 	public boolean insideFrustum() {
 
 	    Vector3f min = new Vector3f();
--- a/src/com/nilunder/bdx/Scene.java
+++ b/src/com/nilunder/bdx/Scene.java
@@ -213,12 +213,10 @@ public class Scene implements Named{
 				g.visibleNoChildren(false);
 				g.modelInstance = new ModelInstance(defaultModel);
 			}
-			Mesh mesh = g.modelInstance.model.meshes.first();
-
 			float[] trans = gobj.get("transform").asFloatArray();
-			JsonValue physics = gobj.get("physics");
-			g.body = Bullet.makeBody(mesh, trans, physics);
-			g.mass(physics.get("mass").asFloat());
+			
+			g.body = Bullet.makeBody(g.modelInstance.model.meshes.first(), trans, gobj.get("physics"));
+			g.currBodyType = gobj.get("physics").get("body_type").asString();
 			g.body.setUserPointer(g);
 			
 			g.props = new HashMap<String, JsonValue>();
@@ -310,7 +308,7 @@ public class Scene implements Named{
 		g.modelInstance = new ModelInstance(gobj.modelInstance);
 		
 		g.body = Bullet.cloneBody(gobj.body);
-		g.mass(gobj.mass());
+		g.currBodyType = gobj.currBodyType;
 		g.body.setUserPointer(g);
 		g.scale(gobj.scale());
 		
@@ -386,8 +384,7 @@ public class Scene implements Named{
 
 
 	private void addToWorld(GameObject gobj){
-		boolean collisionEnabled = !gobj.json.get("physics").get("body_type").asString().equals("NO_COLLISION");
-		if (collisionEnabled)
+		if (!gobj.currBodyType.equals("NO_COLLISION"))
 			world.addRigidBody(gobj.body, gobj.json.get("physics").get("group").asShort(), gobj.json.get("physics").get("mask").asShort());
 		
 		toBeAdded.add(gobj);
--- a/src/com/nilunder/bdx/utils/Bullet.java
+++ b/src/com/nilunder/bdx/utils/Bullet.java
@@ -161,16 +161,20 @@ public static class DebugDrawer extends IDebugDraw{
 		
 		RigidBody body = new RigidBody(ci);
 		
-		if (bodyType.equals("STATIC")){
-			body.setCollisionFlags(CollisionFlags.KINEMATIC_OBJECT);
-			//body.setActivationState(CollisionObject.DISABLE_DEACTIVATION);
-		}else if (bodyType.equals("DYNAMIC")){
-			body.setAngularFactor(0f);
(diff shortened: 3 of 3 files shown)


GoranM:
Wouldn't you need to set mass in cloneNoChildren, so that, when you're passing a GameObject to scene.add, the newly created copy would have the mass of the given object?

I'm not really sure if the dynamics method needs to be that complicated, given the bodyType method, but, in either case, there should be no preconditions for calling dynamics - If you're not supposed to do anything for NO_COLLISION objects, the logic for that should be in the method.

The bodyType method should not rely on the dynamics method, as I think is made obvious by this code:

```java
		if (s.equals("NO_COLLISION")){
			dynamics(true);
			scene.world.removeRigidBody(body);
```

You're enabling dynamics for a NO_COLLISION object ... It doesn't really make sense.

I think the bodyType method should do everything required to set the body type, and then maybe the dynamics method can simply use that:

```java
public void dynamics(boolean enable){
    if (enable && dynamics() || !enable && !dynamics())
        return;

    if (enable){
        bodyType(lastBodyType);
        mass(lastNonZeroMass);
    }else{
        lastBodyType = bodyType();
        lastNonZeroMass = mass();
        bodyType("STATIC");
    }
}
```

rafcolson:
>Wouldn't you need to set mass in cloneNoChildren, so that, when you're passing a GameObject to scene.add, the newly created copy would have the mass of the given object?

No, because ```bullet.cloneBody()``` already takes the mass of the given object to clone the body.
```java
float mass = gobj.mass();
```
Your proposed changes makes it definitely better, but I can't avoid initialization of both ```lastNonZeroMass``` and ```lastBodyType```. Hopefully that's not a problem for you.

To be honest, I would still prefer those to be extra arguments of ```bullet.makeBody()``` so ```GameObject.replaceModel()``` wouldn't have to do redundant actions, such as remove and add the body and set ```mass```, ```collisionFlags``` and ```angularFactor```.

rafcolson:
Maybe the changes weren't that good after all. ```Dynamics()``` should not depend on ```bodyType()```, because the latter should rename the bodyType reference name, which is ```lastBodyType```. :disappointed:

GoranM:
> No, because bullet.cloneBody() already takes the mass of the given object to clone the body.

Oh, that's right.

> I can't avoid initialization of both lastNonZeroMass and lastBodyType.

I don't see why; Both fields should only be used as temporary storage by the dynamics method, and there's no need to initialize temporary storage.

> To be honest, I would still prefer those to be extra arguments of bullet.makeBody() so GameObject.replaceModel() wouldn't have to do redundant actions, such as remove and add the body and set mass, collisionFlags and angularFactor.

Simply calling the relevant methods seems cleaner than having an additional method that takes an extended number of arguments. If Java had something like keyword arguments, then it would work well, but as things stand, a function with more than 3 arguments is usually a bad idea, because you have to remember what they are, and in which order they're passed - this has the potential to be very confusing.

The redundant work is negligible, in terms of performance, so optimizing for ease of use, and overall conceptual clarity, seems more valuable in this case.

Ideally, there would be a simple way to make a deep copy of the "physics" JsonValue node, modify relevant children values, and then pass that copy to makeBody (instead of the original). Unfortunately, all the ways of doing that (that I am aware of) are more complicated than our current approach.

> Dynamics() should not depend on bodyType(), because the latter should rename the bodyType reference name, which is lastBodyType. 

The `lastBodyType` field should not be used to store the current body type. It's basically the same issue we had with the mass method, where you originally had it set `lastNonZeroMass`, remember?

If you need to store the current body type, having a `bodyType` field for that is perfectly fine.

> Any chance you would like to reconsider?

Reconsider what, and why?

rafcolson:
>I don't see why; Both fields should only be used as temporary storage by the dynamics method,

I don't see it that way. Both ```lastBodyType``` and ```lastNonZeroMass``` are called in ```bodyType(String)``` too. It's necessary because of the NO_COLLISION issue I explained in the commit description. And since ```bodyType(String)``` now isn't using the ```dynamics(boolean)``` method, it has to store both fields (something I forgot about when making your proposed changes). But if they were never stored, how would you be able to switch to "DYNAMIC" or "RIGID_BODY"?

We have a similar issue with the ```mass(boolean)``` method. If this method doesn't update ```lastNonZeroMass```, then things don't always work as intended.

Examples:

I have a STATIC object "bucket" in my scene. The player runs into the bucket. Now the bucket should become DYNAMIC. We can't use the dynamics(true) method with the proposed changes. Because ```lastBodyType``` and ```lastNonZeroMass``` weren't initialized. We also can't use the ```bodyType("DYNAMIC")``` method for the same reasons.

Let's start out with a DYNAMIC "bucket". In contrary to the first case the bucket now has to be initialized when added to the scene, to avoid redundant physics calculation. The bucket could be suspended with ```dynamics(false)```, or users might want to use ```bodyType("SENSOR")```. Now the player runs into the bucket. The bucket should become DYNAMIC. In this case we can use the ```dynamics(true)``` method because ```lastBodyType``` and ```lastNonZeroMass``` were initialized. The same applies for ```bodyType("DYNAMIC")```.

But now, the player has to pick up the bucket, take it to the well and fill it with water. While it is being filled with water the player shouldn't be able to move the bucket, so dynamics are suspended. When it's filled and ready to be picked up again the bucket will be heavier, so we'll change the mass to a higher value. But if ```lastNonZeroMass``` is not set in ```mass(float)``` too, you'd have to update the ```lastNonZeroMass``` through a hack, by suspending dynamics (while being suspended already) and restoring again. This won't work as the ```dynamics(false)``` will only suspend when dynamics are enabled, but our bucket becomes heavier while being in a suspended state.

>and there's no need to initialize temporary storage.

I don't think you can say this is a general rule for temporary storage. Temporary storage in some cases, as in this case, is ALSO used as a reference to actual properties. The fact that ```json.get("physics").get("mass")``` and ```json.get("physics").get("body_type")``` have become useless because of ```lastNonZeroMass``` and ```lastBodyType```, prove this to be true.

>The lastBodyType field should not be used to store the current body type. It's basically the same issue we had with the mass method, where you originally had it set lastNonZeroMass, remember?

As I explained above, I still think it would be wiser to update ```lastNonZeroMass``` in ```mass(float)```. In the same way, also the body type should be stored when changing the body type, as you are suggesting:

>If you need to store the current body type, having a bodyType field for that is perfectly fine.

This wouldn't solve the issue with ```dynamics(true)```. Let's say you have a DYNAMIC object and you change it to STATIC with ```bodyType("STATIC")```. If you would then want to enable dynamics you couldn't use ```dynamics(true)``` because ```lastBodyType``` was never initialized.

>Ideally, there would be a simple way to make a deep copy of the "physics" JsonValue node, modify relevant children values, and then pass that copy to makeBody (instead of the original). Unfortunately, all the ways of doing that (that I am aware of) are more complicated than our current approach.

I think not. We could just have an extra ```GameObject.physics``` hashMap, as been suggested in #314. I think this would solve all initialization issues described above, so I'll try this approach.

GoranM:
> I have a STATIC object "bucket" in my scene. The player runs into the bucket. Now the bucket should become DYNAMIC. We can't use the dynamics(true) method with the proposed changes. Because lastBodyType and lastNonZeroMass weren't initialized.

Couldn't you just handle that in the method?

Something like this:

```java
 if (enable){
        if (lastBodyType == null){
            lastBodyType = "DYNAMIC";
            lastNonZeroMass = 1;
        }
        bodyType(lastBodyType);
        mass(lastNonZeroMass);
    }else{
```

> When it's filled and ready to be picked up again the bucket will be heavier, so we'll change the mass to a higher value. But if lastNonZeroMass is not set in mass(float) too, you'd have to update the lastNonZeroMass through a hack, by suspending dynamics (while being suspended already) and restoring again. 

I think you could just do the following, in dynamics:

```java
        mass(mass() == 0 ? lastNonZeroMass : mass());
```

> I don't think you can say this is a general rule for temporary storage. Temporary storage in some cases, as in this case, is ALSO used as a reference to actual properties.

Fundamentally, `dynamics(true)` is supposed to remember last body type and mass; `dynamics(false)` is supposed to set the last remembered values. That's the clean conceptual model in which `lastBodyType` and `lastNonZeroMass` can simply serve as temporary storage for the dynamics method.

That's the basis of a simple system.

The special cases that you brought up (calling `dynamics(true)` on what was originally a static object, and setting a non zero mass on a suspended object) can all be handled locally, in the method (as I hope my examples demonstrate), so I don't think it's necessary to complicate the system by making `lastBodyType` and `lastNonZeroMass` serve dual roles across multiple methods.

> The fact that json.get("physics").get("mass") and json.get("physics").get("body_type") have become useless because of lastNonZeroMass and lastBodyType, prove this to be true.

The json properties carry data for original templates, as they still do. I don't understand how temporary storage fields, for the dynamic method, could make them useless.

rafcolson:
I think I finally got something you might approve of.

As it turned out, it was never necessary to change the mass. Changing collision flags was sufficient enough. I think I was a bit misled by the code in the old ```dynamic(boolean``` method. And so we don't need ```lastNonZeroMass``` as the mass will not change when changing body types. If you really want to know if an object is dynamic or not, you can use ```dynamics()``` for that. And if you want to suspend dynamics you should use ```dynamics(boolean)```, not change the mass to zero.

You can find the details in the description, but I'll post those here anyway:

(#305) GameObject.bodyType & GameObject.dynamics

bodyType:

- new hashMap variable "physics" to be able to change values "mass" and "body_type".
- RIGID_BODY to DYNAMIC: angular velocity is not canceled, therefore disable angular velocity.
- NO_COLLISION to DYNAMIC: all forces and velocities applied while NO_COLLISION will be executed when DYNAMIC. You can do a simple test with a NO_COLLISION Sacky. Run the game and jump. Of course, nothing will happen. But when enabling dynamics without clearing forces, the jump is executed, which is unwanted behavior. Therefore, clear forces and disable velocities.
- kinematic initialization hack: objects that are initially STATIC or SENSOR will have infinite mass. Therefore set mass to 1 (Blender default) when change to DYNAMIC or RIGID_BODY. A second issue is a weird bug with collisionflags for objects that are initially STATIC or SENSOR. When changed to DYNAMIC or RIGID_BODY the body is not activated properly. Anyway, changing to NO_COLLISION first, solves this issue. Luckily this will happen only once during the game, and only with objects that are initially STATIC or SENSOR.
- fix in Bullet.makeBody() with CollisionFlags: STATIC objects were set to kinematic, whilst STATIC GHOST and SENSOR to static. Now those are set to kinematic too. (So CollisionFlags vary between 0, 2, 4 and 6. See Constant Field Values in jbullet docs).

dynamics:

- new variable "lastBodyType" for temporal storage of the body type when disabling dynamics. By default "DYNAMIC" to be able to enable dynamics on objects that are initially STATIC or SENSOR.
- dynamic -> dynamics name change: to avoid confusion with changing physics type to DYNAMIC; dynamics(true) will enable dynamics as dynamics(false) will disable dynamics. Naturally this won't work for NO_COLLISION objects, because those are not in the dynamics world.
- dynamics(boolean) now uses bodyType getter and setter
- dynamics() will only check for objects that are not kinematic, because BDX now doesn't have any static objects.

rafcolson:
PR updated.

GoranM:
The "physics" HashMap, and all the new code it requires, is more complex than simply calling the relevant methods after makeBody, as we did before.

rafcolson:
>The "physics" HashMap, and all the new code it requires, is more complex than simply calling the relevant methods after makeBody, as we did before.

Really, there's no new code except this:
```java
			g.physics = new HashMap<String, JsonValue>();
			for (JsonValue item : gobj.get("physics")){
				g.physics.put(item.name, item);
			}
```
But this saves at least the triple amount of lines at other places. All other new code doesn't have anything to do with this. Those are mere optimizations which would be necessary anyway.

The second, and most important advantage of this is that ```replaceModel()``` now doesn't have to do things twice, which I think would be hacky because of bad programming.

The third advantage is that now we don't need a public variable ```bodyType```. So no risk for users to overwrite such variable without actually changing the body type.

A forth advantage is that lines are a bit shorter when getting physics properties.

Another advantage could be that in the future someone might want to add support for changing other physics properties.

I definitely think all of this makes it less complex, but if you really want me to, I'll take your approach. But I won't like it.

>I'm not sure, but I think bullet expects static objects to have 0 mass - Maybe that's the source of your kinematic bug ... ?

The bug happens when enabling dynamics, not when suspending dynamics. I was having this issue before I removed those lines. And initially ```bullet.makeBody()``` still uses zero mass for STATIC objects. So this can't be the reason.

GoranM:
It's not just the three lines in the loop: There's a new data structure on the GameObject, and additional lines where you reference it in other functions.

> The second, and most important advantage of this is that replaceModel() now doesn't have to do things twice, which I think would be hacky because of bad programming.

I don't know how you define "bad programming", but in my experience, good programming is mostly about focusing on complexity management, especially when performance impact is negligible.

If we can avoid adding and managing a whole new data structure, by simply doing a bit of extra work, isn't that simpler?

> The third advantage is that now we don't need a public variable bodyType. So no risk for users to overwrite such variable without actually changing the body type.

The physics hash is public: Now they can overwrite all physics properties.

> A forth advantage is that lines are a bit shorter when getting physics properties.

I'm not really sure what you mean: `gobj.mass()` is shorter than `gobj.physics.get("mass").asFloat()`.

> Another advantage could be that in the future someone might want to add support for changing other physics properties.

And they would have to remember to update those properties in the physics hash, whenever they change them on the actual body.

With this new level of indirection, a desync between the physics hash and the state of the actual body could occur (a developer can just forget to update it, assuming they're even aware of it in the first place), and that could cause a number of different bugs.

> I definitely think all of this makes it less complex, but if you really want me to, I'll take your approach. But I won't like it.

I want the simple solution.

I hope our discussions will lead us to that.

rafcolson:
>It's not just those lines: There's a new data structure on the GameObject, and additional lines where you reference it in other functions.
>...
>If we can avoid adding and managing a whole new data structure, by simply doing a bit of extra work, isn't that simpler?

The same applies with having a new ```bodyType``` variable. This also is a new data structure on the GameObject, and also this is referenced in other functions to replace ```json.get("physics").get("body_type").asString()```.

I think we have a different understanding about what complexity is. If the existing code is complex and I would have a slightly complex commit in order to make it less complex, then that's making things less complex, isn't it? But if I would have a simple commit making things even more complex ... Well I think I would be doing that by calling ```ghost()```, ```bodyType()``` and ```mass()``` functions in ```replaceModel()```, while ```makeBody()``` is already handling those physics properties, only not as efficient is it should.

>I don't know how you define "bad programming", but in my experience, good programming is mostly about focusing on complexity management, especially when performance impact is negligible.

I agree. Let me be clear on this, to me it's not about performance, that's just the extra gain. To me it's about clarity. It's just confusing for me and I'm sure also to other programmers if there's seemingly redundant code in the code base.

>The physics hash is public: Now they can overwrite all physics properties.

My point is that there's much greater risk in having someone declaring ```bodyType``` as a variable than doing the same for ```physics.get("body_type")```. In a scale of one to ten I would give the latter practically zero, wouldn't you too?

>I'm not really sure what you mean: ```gobj.mass()``` is shorter than ```gobj.physics.get("mass").asFloat()```.

I was talking about the code base. Compare ```gobj.json.get("physics").get("mass").asFloat()``` with ```physics.get("mass").asFloat()```. Example is the arguments of ```addToWorld()```. Makes it more readable to me at least.

>And they would have to remember to update those properties in the physics hash, whenever they change them on the actual body.

That's not hard, is it? If anyone would like to add support for changing "compound" for example, he would look at how this property is get in GameObject. He'll see ```physics.get("compound")``` and looking at the ```physics``` variable he will see it's a container which can be altered.

The alternative is that he would find ```json.get("physics").get("compound")```, which at first sight looks like it could be altered too, but it can't. For not so experienced programmers as myself, it takes a while searching the docs, coming to that conclusion.

>With this new level of indirection, a desync between the physics hash and the state of the actual body could occur (a developer can just forget to update it, assuming they're even aware of it in the first place), and that could cause a number of different bugs.

I think your conclusion is far-reaching. You assume that developers would forget about something like that. But to me it's just the same as creating a new variable for the job and forget to update that.

That said, a higher level of indirection is already present in the code-base, which would be removed by using the ```physics``` hashMap. Right now, it's very easy for the developer to forget about updating ```replaceModel()``` and ```cloneNoChildren()``` if he would be adding for instance support for changing the "compound" physics property. You don't have to worry about updating those if they use the updated ```physics``` hashMap. Instead of updating three things, you now only have to update one thing.

>I want a simple solution.
>
>I hope our discussions will lead us to that.

Discussion is good, only if it's not a waste of time. I sincerely hope I didn't waste yours. I think my arguments are valid, but so do you about your own. That's why I have no problem doing as you suggested, considering how experienced you are in the field of development. I'll make the changes, but I'll keep this commit, in case you'll change your mind.

rafcolson:
PR updated. I was wrong about the amount of lines. This commit has less lines than the previous one.

rafcolson:
PR updated (hopefully it's finished now).

GoranM:
It's a bit too much for one commit, and the commit message is too long, but technically, this is ok, so I'll merge it.

Thanks.

rafcolson:
Yes! Thanks for the merge.

Issue #328, opened 2015-05-25
Suggestion: support for Blender Mesh Origin


rafcolson:
Currently there is no support for objects that have an origin that differs from the center of mass. The result is that the body will not match up with the model of the BDX GameObject.

I'm not exactly sure how to approach this, but I would like to have a go.

GoranM:
The BGE doesn't support this either.

Actually, the outline of the hull, visible in blender, is drawn around the origin, not the mesh, so the user is guided to expect that.

rafcolson:
Okay, but it's fairly easy to get the local origin of the mesh. I managed to get it, so it's only a matter of adding it to the motion state. If I'm finished with this I'll reference it, so you can have a look.

GoranM:
It's not about the implementation; I'm not sure it's wise to do something that goes counter to what the editing environment implies (in blender, the bounds outline is drawn around origin, so if it's different in BDX, at runtime, it could be confusing).

I wonder what @SolarLune thinks.

rafcolson:
I think you're right. That would be confusing with simple shapes. But with Convex Hull and Triangle Meshes this is not the case. That could be the reason why the BGE only supports origin offset with Convex Hull and Triangle Meshes. We could do the same. So if anyone's interested, I've got it as far as giving each GameObject a new variable "origin", which is the offset Vector.

I'll wait for SolarLune before closing this.

SolarLune:
Hello! It should probably follow the GUI as closely as possible. So, Convex Hull and Triangle Mesh shapes can be offset fine, while the others can be locked to origin.

Having an origin vector that indicates how offset the mesh is from the center would be ridiculously helpful, in my opinion. Please feel free to add that in, haha.

rafcolson:
Well, then the question is: should we implement origin in orientation and scale methods?

SolarLune:
I'm not sure what you mean - could you explain a bit?

rafcolson:
At first sight it might seem that the extra argument [```centerOfMassWorldTrans```](http://jbullet.advel.cz/javadoc/com/bulletphysics/linearmath/DefaultMotionState.html) will actually shift the center of mass, but it will only transform the initial ```startTrans```, if I'm correct.

So orientation and scale methods will still use the bounding box center. And so we should implement origin manually.

GoranM:
> It should probably follow the GUI as closely as possible. So, Convex Hull and Triangle Mesh shapes can be offset fine, while the others can be locked to origin.

[...]

Issue #321, opened 2015-05-19
Suggestion: Adding frustum culling 


moshme:
Hi there,
I discovered that there is no frustum culling implemented yet. Neither in bdx or libgdx.
I made some tests with a lot of big objects at the edge of the scene (to be sure they are not inside the camera frustum). Some simple sysouts showed me that they were sent to the modelbatch to be rendered. And also i recognized that the fps dropped signicantly. So i implemented some simple frustum culling with the help of this [tutorial] (http://xoppa.github.io/blog/3d-frustum-culling-with-libgdx/). I took the approach with a sphere to include all the possible rotations of the object.

In my testcases  it increases the the performance a lot. 
Should i make a PR with the solution i came up with and we can discuss my approach to the problem?

SolarLune:
Frustum culling sounds good, yeah. I think using the bounding box of the object is how most engines handle it (so rotating the object can make it appear or disappear, rather than the orientation not affecting it).

GoranM:
The libgdx Camera has a [Frustum object](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/Camera.java#L59), which we update on every frame, as can be seen [here](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/Scene.java#L670). I assumed that libgdx used that to do frustum culling, but if that's not the case, then we should definitely add that feature.

So, yes, you can make that PR.

> I think using the bounding box of the object is how most engines handle it

Yes, that seems like the best way, and since bullet already recalculates the AABB for each object, on every frame, we should be able to simply use that information.

moshme:
Ok, i tried to to implement it with AABB, but this works only for Objects which are added to the bullet world and failed for Objects set to "no collosion". I will have a closer look at this tommorow.

rafcolson:
>Ok, i tried to to implement it with AABB, but this works only for Objects which are added to the bullet world and failed for Objects set to "no collosion". I will have a closer look at this tommorow.

Checkout ```GameObject.dimensions()```. This method uses ```getAabb()``` and doesn't fail with "NO_COLLISION" objects.

moshme:
Ok thanks for hint. Yes there are calculations for those objects, which are correct most of the time.

But yesterday i tested it with objects where the center of the model is translated in blender to the edge of the model and so does not match the center of the aabb. It is only this special case which let it fail.

[...]

PR #337, opened 2015-05-27
(#328) Support for Blender Mesh Origin

rafcolson:
- only for Convex Hull
- all objects have the 'origin' attribute, which represents the offset vector from the bounding box center

This also fixes the ```insideFrustum()``` issue described [here](https://github.com/GoranM/bdx/issues/321#issuecomment-104385141) by @moshme. This issue was not limited to objects with no collision.

Edit: updated: check in ```insideFrustum()``` to prevent calculation when ```origin``` is an empty vector.
Edit: rebased.
Edit: forgot to update the ```centerOfMassTransform``` in ```replaceModel()```. This will be updated soon. Sorry about that.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -122,6 +122,18 @@ def srl_models(meshes):
 
     return name_model
 
+def srl_origins(objects):
+    name_origin = {}
+    
+    for o in objects:
+        if o.type == "MESH":
+            mesh_name = o.data.name
+            if mesh_name not in name_origin:
+                origin = sum([mt.Vector(fv) for fv in o.bound_box], mt.Vector()) / len(o.bound_box)
+                name_origin[mesh_name] = list(origin)
+    
+    return name_origin
+
 def char_uvs(char, angel_code):
     """
     Return a list of uv coordinates (for a quad)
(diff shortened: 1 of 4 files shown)


GoranM:
Instead of having srl_models return two objects, it may be cleaner to make an srl_origins function.

In insideFrustum, it seems like you should use `boundsType()`, instead of referencing the read-only json structure.

rafcolson:
>Instead of having srl_models return two objects, it may be cleaner to make an srl_origins function.

Okay.

>In insideFrustum, it seems like you should use boundsType(), instead of referencing the read-only json structure.

Did you mean ```bodyType()```, because there's no method for getting the json physics' "bounds_type"?

Thanks for the help.

GoranM:
> Did you mean bodyType(), because there's no method for getting the json physics' "bounds_type"?

Oh, yea, we don't have boundsType. I confused myself - you can just ignore that.

>  Oh, but wait. About a separate srl_origins function, wouldn't that cost much with large meshes?

Blender objects have a `bound_box` collection of 8 float arrays that represent 8 vertices of the bounding box in object space, so it could be possible to just do:

```python
origin = sum([Vector(fv) for fv in object.bound_box], Vector()) / len(object.bound_box)
```

rafcolson:
>```python
origin = sum([Vector(fv) for fv in obj.bound_box], Vector()) / len(obj.bound_box)
```
That works. Thanks a lot.

rafcolson:
PR updated. I finally got something working for ```replaceModel()```. But it wouldn't hurt to check the code if things could be done better. Especially, I'm not sure about scaling and transforming the new shape. I was assuming that just copying the transform would also copy the scale, but that's not so. Anyway, it works like a charm. Here's a [test blend](https://www.dropbox.com/s/tf8p2vjy5mhswf0/Test_origin_replaceModel.blend?dl=0) to play with.

Issue #343, opened 2015-05-29
Bugs: parent(compoundObject)


rafcolson:
Previous to #341 I found also bugs with parenting to compound objects.

A first is when an object is parented to a compound object, BDX will crash when unparenting. This [blend](https://www.dropbox.com/s/0urqszyt1erdgh0/unparent_compound_bug.blend?dl=0) will demonstrate that.

Edit: Can't seem to find a solution. It's the known ```"PHYSICS ERROR: Detected collision between Static objects set to Ghost, with Triangle Mesh bounds: Keep them seperated, or use different bounds."``` in ```Scene.update()```. However, also without ghost objects or triangle mesh bounds.

Edit: I found the fix. Quite happy about this.

rafcolson:
**BugFix: Set margin of compound to zero**

- The total collision margin of a compound object (and its children) is the margin of the compound object plus the margin if the children.

rafcolson:
**Fix: Remove and add the parent body in Parent(null)**

- Bug: crash when unparenting a compound parent.

**Fix: Set compound margin to zero**

- Bug: The total collision margin of a compound object (and its children) turns out to be higher as the margin in the Blender Collision tab. It's the total of two margins. The first is the margin of the compound object itself, the second is the margin of the children. If not set, the margin value will be 0.04, so it's necessary to set the margin of the compound object to zero.

**Fix: Remove bodies of children when compound**

- When a collision object is parented to a compound object, the shape of that object is added to the compound shape of the parent, leaving it obsolete. This causes collision interference when the child is not ghost and redundant physics calculation. Solution is to remove their bodies from the world when compound parenting.
- At initialization the children are hooked to their parents before they are added to the world for the first time, initializing collision groups and masks also. A private addToWorldNoChildren() enables hookParentChild() to happen after this. This changes nothing else. Initialization of collision groups and masks would get complicated if not doing this while adding to the world, so this seems to be the best solution.

**Fix: Compound only for collision objects**

- Bug: A child with NO_COLLISION affects the compound shape, which could be undesireable. And it doesn't make sense really for those to be part of any collision shape.
- To be able to do this anyway, the bodyType could always be changed to anything but NO_COLLISION before parenting to a compound object.

rafcolson:
PR submitted.

rafcolson:
[...]

PR #346, opened 2015-06-03
(#343) Compound parent bugs & parent(GameObject g, boolean compound)

rafcolson:
**Fix: Compound Parent(null) crash -> remove/add the parent body**

- Bug: crash when unparenting a compound parent.

**Fix: excessive Compound margin -> set to zero**

- Bug: The total collision margin of a compound object (and its children) turns out to be higher as the margin in the Blender Collision tab. It's the total of two margins. The first is the margin of the compound object itself, the second is the margin of the children. If not set, the margin value will be 0.04, so it's necessary to set the margin of the compound object to zero.

**Fix: Compound child collision detection -> remove body**

- When a collision object is parented to a compound object, the shape of that object is added to the compound shape of the parent, leaving it obsolete. This causes collision interference when the child is not ghost and redundant physics calculation. Solution is to remove their bodies from the world when compound parenting.
- At initialization the children are hooked to their parents before they are added to the world for the first time, initializing collision groups and masks also. A private addToWorldNoChildren() enables hookParentChild() to happen after this. This changes nothing else. Initialization of collision groups and masks would get complicated if not doing this while adding to the world, so this seems to be the best solution.

**Fix: replaceModel broken with Compound child -> unparent**

- Bug: replaceModel didn't support physics update if compound parented.

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -658,6 +658,12 @@ public class GameObject implements Named{
 		}
 		
 		if (updatePhysics){
+			GameObject compParent = parent != null && parent.body.getCollisionShape().isCompound() ? parent : null;
+			boolean isCompChild = compParent != null && !(currBodyType.equals("NO_COLLISION") || currBodyType.equals("SENSOR"));
+			if (isCompChild){
+				parent(null);
+			}
+			
 			Matrix4f transform = transform();
 			Vector3f scale = scale();
 			String boundsType = json.get("physics").get("bounds_type").asString();
@@ -681,10 +687,14 @@ public class GameObject implements Named{
 
 			if (body.isInWorld()){
 				scene.world.updateSingleAabb(body);
-			}else{ // update Aabb hack for NO_COLLISION
+			}else{ // update Aabb hack for when not in world
 				scene.world.addRigidBody(body);
 				scene.world.updateSingleAabb(body);
 				scene.world.removeRigidBody(body);
+			}
+
+			if (isCompChild){
+				parent(compParent);
 			}
 		}
 	}
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -81,16 +81,22 @@ public class GameObject implements Named{
 	}
 	
 	public void parent(GameObject p){
+		parent(p, true);
+	}
+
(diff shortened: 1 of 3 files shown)


GoranM:
> Initialization of collision groups and masks would get complicated if not doing this while adding to the world, so this seems to be the best solution.

In what way? Please describe the actual problem that your changes are intended to solve.

Also, please remember that commit messages should be concise. Try to condense things down to a single sentence: What you did, and why.

Don't write paragraphs - If you really need to elaborate on something, you can do it in the PR.

rafcolson:
>In what way? Please describe the actual problem that your changes are intended to solve.

As far as I know it's not possible to set the group or mask unless objects are added to the world already. Unless I'm wrong and it is possible, but that would be making things quite complicated, by manually constructing the BroadfaseProxy's. I've looked into it a couple of days, but couldn't get it to work.

But all of this is aside of what's the actual intention of the commit to solve the problem, which is: *At initialization the children are hooked to their parents before they are added to the world for the first time, initializing collision groups and masks also. A private addToWorldNoChildren() enables hookParentChild() to happen after this.* In other words, in order to have compound parenting work at initialization, this has to happen *after* all instances are added to the scene. Else, objects with collision (and not ghost) would be in the world, while they should remain out of the world.

rafcolson:
>Also, please remember that commit messages should be concise. Try to condense things down to a single sentence: What you did, and why.

I'll try to find more descriptive titles for those commits. Those paragraphs where intended to describe the commit content in detail. But if it's not really helpful, I can leave those out the next time.

rafcolson:
PR updated.

rafcolson:
I would like to put this on hold until #347 has been discussed.

rafcolson:
PR updated. Note that I removed **Fix: NO_COLLISION or SENSOR Compound child -> only if not** and added method ```GameObject.parent(GameObject p, boolean compound)``` to open the possibility to disable compound parenting if wanted. Not really a fix, but anyhow, I think I can add it to the PR.

GoranM:
> Do you know how I could change those titles directly? I'm always doing a git reset HEAD^ to get to the previous commits and do git commit --amend. This requires me to save every commit in a text file in order to restore. If you know a better way, that would be really helpful.

https://help.github.com/articles/changing-a-commit-message/

rafcolson:
Thanks, that's very useful. I must have looked over this before.

Issue #320, opened 2015-05-17
Suggestion: Local dimensions() function


SolarLune:
Hey, there. So, the Gameobject.dimensions() function returns the bounding box dimensions of the object, which can change if the object rotates. It'd be nice if there were a "dimensionsBase()" function, which would return the dimensions of the object without any orientation or scaling changes. Any ideas? 

I'm not good with matrices and Vectors, so an explanation of how to create this function if it's OK'd would be appreciated.

EDIT: OK, I think I got it. Multiply the vector by the inverse of the orientation matrix, then call absolute() on the vector. Anyway, does anyone like the sound of this local dimensions function?

rafcolson:
Yes, I do. But wouldn't "dimensionsLocal" be a better name? Because we already have "velocityLocal" and so fort.

SolarLune:
True, and I do agree. I guess disregarding scaling isn't really necessary, so just having it be local in terms of orientation would be helpful enough.

GoranM:
Perhaps dimensions() should simply return those of the object oriented bounding box?

I don't think we really need axis aligned, do we?

> Multiply the vector by the inverse of the orientation matrix, then call absolute() on the vector.

Wouldn't that simply rotate the vector? 

Imagine an object with an orientation that describes a rotation of pi/4 (45 degrees) around the Z axis: If `d = dimensions()`, and `d.x == d.y`, `inverse_orientation.mult(d)` would produce a vector v where `v.y == 0`.

PS: In the future, it would be best to justify suggestions with something more concrete than "It'd be nice". Why would it be nice? Why do you need it? Answers to those questions tend to be fairly convincing.

SolarLune:
> Perhaps dimensions() should simply return those of the object oriented bounding box?
I don't think we really need axis aligned, do we?

Not to my knowledge, no. Having dimensions simply return the globally aligned dimensions vector would be fine.

> Wouldn't that simply rotate the vector?

It should rotate it back to "identity", so to speak, so it's the normal version of the matrix again.

> PS: In the future, it would be best to justify suggestions with something more concrete than "It'd be nice". Why would it be nice? Why do you need it? Answers to those questions tend to be fairly convincing.

I didn't think I needed to go into deep detail about this, but okay. It's generally undesirable for dimensions() to return a rotated box as it changes the dimensions of the shape, making processes like positioning something one object-width away basically unreliable.

rafcolson:
[...]

PR #353, opened 2015-06-06
(#320) GameObject.dimensions() now from Blender dimensions

rafcolson:
Note: I thought it would be best to keep the old method for Text objects exclusively.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -134,6 +134,17 @@ def srl_origins(objects):
     
     return name_origin
 
+def srl_dimensions(objects):
+    name_dimensions = {}
+    
+    for o in objects:
+        if o.type == "MESH" or "FONT":
+            data_name = "__FNT_" + o.data.name if o.type == "FONT" else o.data.name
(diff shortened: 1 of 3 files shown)


GoranM:
> I thought it would be best to keep the old method for Text objects exclusively.

Why?

rafcolson:
Sorry, I guess I should just update ```aabb``` and ```dimensionsNoScale``` in ```Text.set()```. Then there's no need for an override. Working on it.

rafcolson:
PR updated. I'll leave the update of ```dimensionsNoScale``` in ```Text.set()``` for another time as I still have to figure out how to calculate it. Also, the dimensions in Blender are not exactly the same as those in BDX, which I figure is due to the font conversion. So I'm not sure how to tackle this. Should I maybe set ```dimensionsNoScale``` right after conversion at initialization?

Anyway I guess you could just merge this, if no one cares about this update right now.

GoranM:
Dimensions for Text object will have to incorporate character width/height - It's something that should be done in a different PR.

PR #363, opened 2015-06-09
Fix: ensure parent(null) will not add bodies if already null

rafcolson:
In some cases (like in the code for ```endNoChildren()```) we would like to call ```parent(null)``` to a list of objects, while some of those may already be without a parent. In such case, calling would add the body again, which is not what we want.

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -99,6 +99,9 @@ public class GameObject implements Named{
 					scene.world.addRigidBody(parent.body);
 				}
 			}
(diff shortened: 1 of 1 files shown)


GoranM:
There's no need to prefix your commit messages with "Fix:".

rafcolson:
Okay, shouldn't I do it in general or only in this particular case? I mean, next time I'll commit a bug fix, should I prefix it with "Fix" or would that be unnecessary?

GoranM:
It's not necessary in general - the issue reference is there to provide more detail, if required.

Issue #364, opened 2015-06-09
Bug: srl_dimensions() doesn't work with groups.


SolarLune:
Hey, there. So the srl_dimensions() function in exporter.py doesn't work with group objects, so BDX can't export and run if there's a group instance in the scene.

A proposed solution would be to get the "first, non-parent" object / group's "instanced object" (recursively, if it's also a group), and use its dimensions instead. The instance() function would probably be key to help make this easier, though I don't see any recursiveness in this in case you create groups of groups (not usual at all, but not a bad idea in game creation, either). 

It might also be a good idea to expand the function to account for this possibility; I'll see if it's necessary after I get this PR up and running.

SolarLune:
Ah, it just dawned on me that because it's a group instance and doesn't "truly exist", it doesn't need to have any dimensions. The actual object will be swapped in - group instances aren't "anything" in BDX. I think I've got it.

A kind of weird side bug that I noticed is that it doesn't seem possible to instance a group that consists of another group (i.e. instance a level that consists of ground pieces). In engine, I think it should be fine to create groups of objects that are linked through parenting, but it doesn't seem to work actually in Blender. Oh well; that's a different bug.

PR submitted.

SolarLune:
OK, after testing this out with my in-progress game, it's not working; it's a problem with getting the dimensions for the __PDisplay object. I'll wait for @rafcolson to weigh in on this.

rafcolson:
I'm sorry SolarLune, made a stupid mistake. Probably because I'm not as used to Python as I was before. But I'm guessing this will fix both of the issues you mentioned.

SolarLune:
No problem, it does indeed fix this issue. Thanks!

PR #365, opened 2015-06-09
Fix to skip getting dimensions for groups. (Bug #364)

SolarLune:


rafcolson:
Hi, SolarLune, could you maybe check if #367 fixes both issues? It probably will. But I'm not sure if I tested every particular case with Groups.

Issue #265, opened 2015-04-24
Discussion: Object Panel


SolarLune:
Hey, there. 

I think an idea we had at some point was to create a new panel in the object properties panel to allow us a few conveniences? 

As I recall, it would be able to:

- Customize which class an object ran aside from the default "name-based" approach,
- Assign new game properties to the object (rather than having to use the Game Logic panel for it), and 
- Tweak execution priority to execute some objects before others.

Am I correct? Discuss!

rafcolson:
You're correct.

GoranM:
Yes, I think we agreed on something like that.

rafcolson:
I'd like to add something:

- a check box "Show Profiler" and another check box "Small" next to it (which will become active when the "Show Profiler" box is checked).

GoranM:
The "Object Panel" is for object specific data, not game specific options. The settings that already exist in blender are sufficient.

Also, I don't think there's a need to have size options for the profiler - It doesn't really matter that much, as long as the info is readable.

rafcolson:
Okay.But is it okay to make the profiler's display size around 81.25% of the current size? Or 87.5%?

GoranM:
I think that's fine.

rafcolson:
Someone working on this? Any progress?

SolarLune:
Ah, no, I started, but I'm really rusty with the add-on API and didn't get very far. If anyone else thinks they can take a crack at it, lemme know.

rafcolson:
Okay, thanks. I'll let you know if I will do that.

rafcolson:
So, I'm having a go at this. If you have any design preferences, please share. This is what I had in mind:
______________
\/ Java Script
[field for Script] [checkbox for Execution Priority]

\/ Properties [same as in Logic Editor]
+ Add Game Property
______________

SolarLune:
Very cool. I'd change the Java Script section. I think it should be called "Java Class", and there should be a checkbox to enable the custom class. If enabled, then the field for the class is enabled and you can type in the name of the class the object should correspond to. Execution Priority shouldn't just be a boolean, but rather an integer spinner. 

Ideally, we would sort the objects by priority and execute them in that order. I'd guess we'd just store a Hashmap of integer numbers (priority) to ArrayLists of objects. BDX would place the objects into the relevant lists for their priorities, and then loop through them from the "top down" to execute objects' logic.

[...]

PR #377, opened 2015-06-15
(#265) BDX Object Panel

rafcolson:
For "use_priority" I used the same icons as in the Logic editor.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -386,6 +386,12 @@ def projection_matrix(camd):
     return sum([c for c in mat], [])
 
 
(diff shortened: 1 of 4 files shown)


GoranM:
Unless there's a good reason for putting it in a new file, the code for the new panel should be placed in ui.py.

"BdxPanel" is not a very descriptive name. I think something like "BdxObject" would be much better (there's no need to specify "Panel", because inheritance makes that clear).

rafcolson:
PR updated. I replaced all single quotation marks by double quotation marks. Not that it really matters, but it's a bit more appealing.

Issue #380, opened 2015-06-17
String.format cannot be used in projects intended for web


billy1380:
The use of String.format - in InputMaps - I don't think there is a string.format emulation in gwt.

Extract from `gradlew html:dist`
```
Compiling module com.nilunder.testgame.GdxDefinition
   Validating units:
      [ERROR] Errors in 'jar:file:/xxx/testgame/core/libs/bdx.jar!/com/nilunder/bdx/inputs/Gamepad.java'
         [ERROR] Line 78: No source code is available for type com.badlogic.gdx.controllers.Controller; did you forget to inherit a required module?
         [ERROR] Line 86: No source code is available for type com.badlogic.gdx.controllers.Controllers; did you forget to inherit a required module?
         [ERROR] Line 136: No source code is available for type com.badlogic.gdx.controllers.PovDirection; did you forget to inherit a required module?
      [ERROR] Errors in 'jar:file:/xxx/testgame/core/libs/bdx.jar!/com/nilunder/bdx/inputs/GdxProcessor.java'
         [ERROR] Line 24: No source code is available for type com.badlogic.gdx.controllers.ControllerAdapter; did you forget to inherit a required module?
         [ERROR] Line 36: No source code is available for type com.badlogic.gdx.controllers.PovDirection; did you forget to inherit a required module?
         [ERROR] Line 58: No source code is available for type com.badlogic.gdx.controllers.Controller; did you forget to inherit a required module?
      [ERROR] Errors in 'jar:file:/xxx/testgame/core/libs/bdx.jar!/com/nilunder/bdx/inputs/InputMaps.java'
         [ERROR] Line 79: The method format(String, String) is undefined for the type String
   [ERROR] Aborting compile due to errors in some input files
:html:compileGwt FAILED
```

billy1380:
String.format is currently used in InputMaps.java and Profiler.java. The first use can be easily replaced with +. The Profiler howerver makes extensive use of length modifiers so not sure whether we can settle for unformatted floats. The only other alternative is to probably to have a gwt implementation of the profiler that uses gwt number formatters.

@GoranM any thoughts.

billy1380:
For reference, the GWT issue

https://github.com/gwtproject/gwt/issues/3946

rafcolson:
Maybe we can use [String.Format](http://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html)

billy1380:
[...]

Issue #379, opened 2015-06-16
Html build issue (Controller)


billy1380:
The build complaining about controllers (I think because they a web implementation is not available). 

Extract from `gradlew html:dist`
```
Compiling module com.nilunder.testgame.GdxDefinition
   Validating units:
      [ERROR] Errors in 'jar:file:/xxx/testgame/core/libs/bdx.jar!/com/nilunder/bdx/inputs/Gamepad.java'
         [ERROR] Line 78: No source code is available for type com.badlogic.gdx.controllers.Controller; did you forget to inherit a required module?
         [ERROR] Line 86: No source code is available for type com.badlogic.gdx.controllers.Controllers; did you forget to inherit a required module?
         [ERROR] Line 136: No source code is available for type com.badlogic.gdx.controllers.PovDirection; did you forget to inherit a required module?
      [ERROR] Errors in 'jar:file:/xxx/testgame/core/libs/bdx.jar!/com/nilunder/bdx/inputs/GdxProcessor.java'
         [ERROR] Line 24: No source code is available for type com.badlogic.gdx.controllers.ControllerAdapter; did you forget to inherit a required module?
         [ERROR] Line 36: No source code is available for type com.badlogic.gdx.controllers.PovDirection; did you forget to inherit a required module?
         [ERROR] Line 58: No source code is available for type com.badlogic.gdx.controllers.Controller; did you forget to inherit a required module?
      [ERROR] Errors in 'jar:file:/xxx/testgame/core/libs/bdx.jar!/com/nilunder/bdx/inputs/InputMaps.java'
         [ERROR] Line 79: The method format(String, String) is undefined for the type String
   [ERROR] Aborting compile due to errors in some input files
:html:compileGwt FAILED
```

billy1380:
This looks interesting http://www.badlogicgames.com/wordpress/?p=2743

billy1380:
Submitted a pull request with the changes. Project seems to build after including the controller module.

PR #381, opened 2015-06-17
Fix for issues #379 and #380

billy1380:


GoranM:
The first commit should be split into two: You should have one for adding eclipse files to gitignore, and another for removing the use of String.format.

If your formatForDisplay method works (@rafcolson can probably help you test that) , I'll accept your code, because it resolves a bug, but it should be considered a temporary fix, because, ideally, we would be able to make a proper alternative for String.format, which would just work when compiling with GWT.

It should be possible to do that via "super source", as documented [here](http://www.gwtproject.org/doc/latest/DevGuideOrganizingProjects.html#DevGuideModuleXml) (under the heading: "Overriding one package implementation with another").

billy1380:
@GoranM don't worry about accepting the code with the current commits... I can easily recreate the commits by creating another branch and applying the changes one at a time. I just wanted some feedback. Give me a nod if you prefer that.

I have built the code and run it since the last post but the project crashes after loading resources as can be seen [http://spacehopperstudios.com/testenv/testgame/](http://spacehopperstudios.com/testenv/testgame/ "BDX Test Game"). I know it loads some resources successfully because I can see them come down in the dev tools.

In superdev mode the message is "Cannot read property type_0 of undefined" but I would rather someone else had a go with before I raised a bug especially that I think that it might be a libgdx issue.

About String.format - I think String.format ultimately wraps java.util.Formatter and the implementation of the format method is a grizzly thing (because of the versatility mixed with locale awareness). Been looking for some code with a compatible license that I might be able to base it on... but found nothing so far (except some suggestions on the gwt issue I linked to). One thing that might be useful is if we can find code from the apache harmony project we might be able to use that: thoughts?

Anyway... let me know how the testing goes and I can sort the commits however you prefer.

billy1380:
I setup an html5 project from scratch with gdx-setup.jar and built it for testing using superdev and it runs fine, so it might be a bdx/my code changes after all :sob:

rafcolson:
So do I have to test your commit?

billy1380:
If you get a chance, please do.

rafcolson:
I've tested the commits. Everything seems to work fine.

billy1380:
Nice one... Thanks very much. I will close this and submit another pull request for the String.format changes only.

billy1380:
@GoranM new pull requests submitted.

PR #389, opened 2015-06-20
removed spaces from parameters and fixed float formatting

billy1380:


rafcolson:
It's still the same as before, which is not surprising because it's virtually the same commit as before.

The text is too long because of the redundant spaces. And that's why "fps" is not completely shown ("fp"). That's why the bars start inside the text.

billy1380:
Apparently in String.format the 4 in 4.1 is the width of the entire field. This means that the issues can either be fixed by calling the current method `addFloat` with 2, 1 or changing the meaning of the 3rd parameter to field padding instead of integer part padding. I have opted for the 2nd solution just in case these functions are later recycled into a String.format implementation.

GoranM:
I'm not sure what this PR is supposed to fix - Didn't @rafcolson already fix the formatting issues with your patch?

billy1380:
It removes the spaces from the parameters and puts them into the formatting method which now behaves like String.format did

rafcolson:
>It removes the spaces from the parameters and puts them into the formatting method which now behaves like String.format did

It doesn't, at least on my end. As I already explained:

>The text is too long because of the redundant spaces. And that's why "fps" is not completely shown ("fp"). That's why the bars start inside the text.

@billy1380 Could you please confirm if this is also true on your end, or not. If not, then maybe we should look first why we have different results. If so, then this commit should be adjusted so it does not change anything to the profiler visually.

billy1380:
I have adjusted the commit since and the fps now appears and the bars are no longer overlapped.

rafcolson:
Now it's good. You could always merge both commits to one by doing ```git reset HEAD^``` and then doing ```git commit --amend -am removed spaces from parameters and fixed float formatting```.

GoranM:
Please follow rafs advice, and make this into a single commit.

billy1380:
I am not sure what that has achieved; anyway... I ran the command and pushed the changes. It now looks like a mess to me.

If you are really hell-bent on getting a single commit for the changes I can re-do the changes on a clean branch.

Let me know.

billy1380:
Actually... I will just do that anyway... closing.

Issue #400, opened 2015-07-11
About GameObject Touching functions


SolarLune:
Hey, there. So there's two points I want to consider.

1) I feel like the current touching functions are OK, but also hold back a bit of information. For example, if you check if the player object is touching a game object named "Spikes", it won't return the game object you're touching if you are, but rather just a boolean. I know Java has something to do with that (it's more awkward to type `touching("blah") != null` vs. `touching("blah")`), but I think we should err on the side of providing too much information when we have these kinds of functions and properties. It makes things easier when it comes to doing more complex things, since you don't have to handle loops or checks yourself if you want to, say, damage the player for how much damage power a bullet has. Rather than checking to see if you've been hit, looping through objects to find the one you're looking for, and then checking its value, you can just check the function to see if it's true and get the GameObject itself to work with from there.

2) As a side-note, it'd be nice if there were a couple more functions for checking to see if a GameObject is touching a specific kind of object. Two identifying mechanisms in BDX at the moment are components and game properties. So, I propose two more functions, GameObject.touchingProperty() and GameObject.touchingComponent(). Both would take strings, and would be the name of the property to check for, or the name of the component to check for, respectively. They would loop through touchingObjects and check to see if the objects have a property or component by the name specified. If they find it, they would return a reference to the object (in line with the above point; otherwise, it'd just be a boolean).

These would both be nice because currently, there's no way to assign a "tag" to an object and check for that tag easily (i.e., how spikes, lava, or a bottomless pit can be hazardous to a player object).

hitComponent or hitProperty varients would also be appreciated.

GoranM:
1:

If you look at the implementation of the "touching" method, you'll notice that it's simply a convenience for `touchingObjects.get("name") != null`.

So, if you need a reference to a touching object, you can get it like that.

BTW: Wouldn't it make more sense to have the bullet deal the damage, instead of having the object damage itself? That could also simplify your code somewhat, because you don't have to check for everything that could possibly damage you on every object that can take damage.

2:

I think we should follow existing conventions (which seem sensible), and have `touchingObjects.getByProperty` and `touchingObjects.getByComponent`.

If you want to have convenience methods that return booleans, `touchingProperty` and `touchingComponent` would make sense, as would the corresponding `hit*` methods.

SolarLune:
1) Ah, I see.

Having the bullet deal damage would be better; it was just an example, though. I'm sure as gameplay becomes more complex, having direct access to the game object being touched might be a good idea, though.

[...]

PR #401, opened 2015-07-12
(#400) - Additional GameObject hit / touching functions.

SolarLune:
Also, a fix for hit(objectName) so that it works with multiples of the same named object.


GoranM:
It might be better to make ArrayListGameObject a public static class on the GameObject itself, since it's not really a general purpose thing.

SolarLune:
OK, PR updated.

Issue #403, opened 2015-07-14
Suggestion: Allow Custom Component Names


SolarLune:
Hey, there. So, I think it'd be helpful if we could customize the names of Components. This way, we could utilize multiple instances of the same class of components and differentiate between them. 

Here's why I like the idea: I have a basic driving game that'll have health and fuel for the car. They are custom Components that dictate how much health or fuel you have left, and different things will be able to influence them (i.e. taking damage reduces health, boosting reduces fuel, etc). They're components because they have convenience functions to make things simple (i.e. health.add(10) will add 10 points and not go over a maximum amount, health.sub(10) will never go below 0 and won't work if the component is invincible).

They're currently two different classes, but functionally, they're the exact same (the Fuel class is just an extended Health class). If I'll need to get them later without prepared references, I'll need them to have different names (so I can use components.get("Health") or components.get("Fuel")). However, since they're the same, what I'd like to do is generalize them into simple `Gauge()` component classes that I can also re-use in different projects for different situations. The idea would basically be:

```java
Gauge g = new Gauge(this);
g.name = "Health";
components.add(g);

g = new Gauge(this);
g.name = "Fuel";
components.add(g);
```

With this concept, I can have multiple instances of classes while still being able to access them by name later. An alternative would be something like:

```java
components.add(new Gauge(this), "Health");
components.add(new Gauge(this), "Fuel");
```

This one is simpler, but would require a new add() function for the components list that allows one to provide a string to name the component. Without a customized name, it would default to how it works now - just the simple name of the class.

Thoughts?

GoranM:
Yes. A name field would be fine.

SolarLune:
PR Submitted. I'd like to have the ability to do a one-liner like above (or by putting the name of the component in the constructor) rather than spreading it out across multiple lines, but I don't think there's a simple way to do it without breaking existing extended components...? I guess we could add a name() function that returns the instance...
```java
components.add(new Gauge(this).name("Health"));
```

Do you think it'd be worth it?

GoranM:
> I don't think there's a simple way to do it without breaking existing extended components...?

I take it you tried, and it didn't work for some reason?

> Do you think it'd be worth it?

Not really.

SolarLune:
Actually, I guess it worked okay; I think I was just looking at my own component and thinking it wouldn't work without realizing I could just update it to use the new constructor, haha. I've updated the fork with a new constructor that takes a string.

SolarLune:
[...]

PR #406, opened 2015-07-16
(#403) - Additional Component constructor.

SolarLune:
As a note, I don't just call `this(g, this.getClass().getSimpleName())` in the new constructor because you can't use instance methods in a constructor. apparently.


GoranM:
As I said in the issue, I don't think the benefits of this are worth the additional code.

If you have a widely applicable component like "Gauge", you can set the constructor for Gauge to take a name, or any number of arguments.

Is there a significant benefit to having this feature on the base Component class?

SolarLune:
OK, I see what you mean - we can implement it in our extended components themselves. That makes a lot more sense and is simpler as well. 

Thanks. PR closed.

Issue #413, opened 2015-08-10
Suggestion: Mouse wheel support


rafcolson:
As the title suggests.

SolarLune:
Agreed, mouse wheel support would be great to have as an option. I guess it might be implemented like the joystick axes, in that there's a "deadzone" that you would have to scroll past for it to be a valid input? It'd be nice to add it to InputMaps as well, eventually.

GoranM:
Yes, mouse wheel support would be a nice addition.

> I guess it might be implemented like the joystick axes, in that there's a "deadzone" that you would have to scroll past for it to be a valid input?

The wheel usually has detents to turn in discrete steps, rather than continuously as an analog axis, so I'm not sure if a "deadzone" would have any utility there.

rafcolson:
I tried with ```Gdx.InputProcessor``` but couldn't get it working.

Maybe better use ```java.awt.event.MouseWheelListener```?

GoranM:
What did you try?

rafcolson:
I'm not so familiar with the workings of it. I extended InputProcessor to a custom class IP and added this to gdx.input. Then I accessed IP.scrolled(1) and IP.scrolled(-1) to get the necessary info, but I'm always getting false.

GoranM:
You're not supposed to extend InputProcessor - We already have a [GdxProcessor](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/inputs/GdxProcessor.java) that does that.

You simply need to add the [scrolled](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/InputProcessor.html#scrolled-int-) method, which will set the necessary data on the mouse structure.

rafcolson:
Alright. Thanks a lot.

rafcolson:
It's difficult for me to see how it works. I would appreciate it if you could add the changes, or @SolarLune.

GoranM:
Can you be a bit more specific about what you're finding difficult to understand?

If you actually look through the GdxProcessor code, you can see how we update mouse button state via the implemented touchDown and touchUp methods.

rafcolson:
Okay, I got this far. I assume we don't have to "log" anything, or do we? And if so, what do we log?

GoranM:
> Okay, I got this far.

... ?

You just have a method that always returns true.

> I assume we don't have to "log" anything, or do we? And if so, what do we log?

It depends on how you want to track wheel state.

[...]

PR #422, opened 2015-08-14
(#413) Basic Mouse Wheel support

rafcolson:


GoranM:
I'm not sure if a "ScrollLog" is the best approach here.

You could just have an int on Mouse, which can be reset to 0 in Bdx.java (in the main method, where the input is updated).

Also, I think wheelMove would be a better name for the method - scrollAmount implies that the returned value could be an arbitrary number, which is not true.

As for indentation - that's not an error: I intentionally left-aligned internal classes, to conceptually reinforce their "private static" role - Even though they're contained within a class, they should be considered as external, self-contained units, which should have no direct access to the properties of their host.

rafcolson:
PR updated. As you can see ```mouse.wheelMove``` is reset after all scenes are updated.

Issue #421, opened 2015-08-14
Suggestion: Don't Export if Scene Doesn't Have Valid Camera


SolarLune:
Hey, there. 

So basically, if you try to use BDX without a camera set (like in a layer different from the one with the camera), BDX will crash out (logically, of course) on start. It'd be nice if this was replaced with a simple error message on export, like "Please set a valid starting camera object for this scene and layer."

GoranM:
The right way to address this would be to generate default camera data, when necessary, just like the BGE.

rafcolson:
I'm assuming this would mean that BDX should also be able to generate a default object. This could also be the foundation for possibly ```scene.addPrimitive()```, ```scene.addEmpty()```, ```scene.addLight()```, ```scene.addCamera()```. But that's another suggestion. Anyway, it would be great if the code for generating data for such default camera could also be used for these other kind of objects.

I've tried this a couple of months ago, but I was having some trouble with building the the JsonValue ```g.json```.

GoranM:
> I'm assuming this would mean that BDX should also be able to generate a default object.

No, because the exporter can simply generate the camera data, which BDX can load normally.

rafcolson:
How could BDX load a camera if there is no camera? Are you talking about the Custom Camera ?

GoranM:
I'm not sure what you mean by "User Camera".

If you have a scene without a camera, the exporter could be modified to create the data for a default camera, as if it actually existed in the scene. Since BDX only knows what the exporter data encodes, if a .bdx file contains camera data, BDX will simply use it.

rafcolson:
>I'm not sure what you mean by "User Camera".

I'm not sure how it works, but the BGE generates the User Camera, so it should be accessible in the bpy, I think. I think it would be nice if BDX would also import this camera if not in Camera View.

>If you have a scene without a camera, the exporter could be modified to create the data for a default camera, as if it actually existed in the scene. Since BDX only knows what the exporter data encodes, if a .bdx file contains camera data, BDX will simply use it.

I understand. But it would be great if the exporter would also create the data for a default object (which would be a large part of the default camera data anyway), and so it would be possible to use this default object, preferably an empty (which is actually a NO_COLLISION cube with no material), for features such as ```scene.addEmpty()```, ```scene.addPrimitive("CUBE")```, ```scene.addPrimitive("UV_SPHERE")```, ... In conjunction with a method ```GameObject.material()``` this would open up a lot of possibilities.

[...]

PR #427, opened 2015-08-22
(#421) Blender 3D View camera support

rafcolson:
This allows previewing with the 3D View.

A new camera will be replicated when:
- no active camera exists
- 3D View type is not "CAMERA" (mode = "User Persp" or "User Ortho")
A warning then is printed to the console.

Note: When the screen doesn't contain a 3D View region, an exception will be raised.

Extra note: [```name_object["__EMPTY__"] = generate_object()```](https://github.com/rafcolson/bdx/commit/edf6dd039fcd8c15a9b144bed5fed2fd8dcf09f9#diff-e3fabe6964c359c268b2a1dfc4ee1f00R525) only to illustrate how it is now possible to generate a default empty. But if this PR has a go, I'll leave that out for another commit.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -393,6 +393,35 @@ def get_cls_name(obj):
     return obj.name
 
 
+def relevant_region_3d_data():
+    
+    def get_areas_3d_data(areas):
+        return [[a, a.height * a.width] for a in areas if a.type == "VIEW_3D"]
+    
+    r3d = bpy.context.region_data
+    if r3d:
+        if r3d.view_perspective == "CAMERA":
+            return
+    elif bpy.context.scene.camera:
+        return
+    else:
+        a3d = get_areas_3d_data(bpy.context.screen.areas)
+        if not a3d:
+            screens = bpy.data.screens
+            if "BDX" in screens:
+                a3d = get_areas_3d_data(screens["BDX"].areas)
+                if not a3d:
+                    a3d = get_areas_3d_data(sum([list(scr.areas) for scr in screens], []))
+            else:
+                a3d = get_areas_3d_data(sum([list(scr.areas) for scr in screens], []))
+            if not a3d:
+                return
+        if len(a3d) != 1:
+            a3d.sort(key=lambda lst: lst[1])
+        r3d = a3d[-1][0].spaces[0].region_3d
+    return r3d
+
+
 def srl_objects(objects):
     name_object = {}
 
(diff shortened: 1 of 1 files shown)


GoranM:
> Extra note: `name_object["__EMPTY__"] = generate_object()` only to illustrate how it is now possible to generate a default empty. But if this PR has a go, I'll leave that out for another commit.

For this PR, we only need to generate camera data, so there's no need for a generate_object method.

Just set the required data: `new_cam = { ... }`.

-

Why are you printing a "WARNING"? Is there a situation in which this warning could be helpful?
 
-

If you loop through `bpy.data.screens['BDX'].areas`, instead of `bpy.context.screen.areas`, r3d should never be None.

-

In camera_names(), couldn't you just check `scene.camera` once?:

```python
if scene.camera:
    # make sure scene.camera.name is first
else:
    # insert "__CAMERA__"
```

rafcolson:
Thanks for your reply, @GoranM 
>For this PR, we only need to generate camera data, so there's no need for a generate_object method.

Okay.

>Why are you printing a "WARNING"? Is there a situation in which this warning could be helpful?

You're right. I was thinking the same thing.

>If you loop through ```bpy.data.screens['BDX'].areas```, instead of ```bpy.context.screen.areas```, r3d should never be None.

I don't think that would work as intended always. Users may not have a 'BDX' screen in there blend project. But this is my plan:

If your mouse is over a 3D View area and P is pressed, that particular 3D View is replicated. Otherwise if the Export and Run button is used, the 3D View of the largest area of the current screen is replicated. If none is found and there is an active camera in the scene, this camera is used. Otherwise the largest 3D View area of the 'BDX' screen is replicated, if present, else an exception is raised: "No active camera or 3D View data in 'BDX' screen was found."

>In camera_names(), couldn't you just check scene.camera once?:
```java
if scene.camera:
    # make sure scene.camera.name is first
else:
    # insert "__CAMERA__"
```

Not exactly, because of the OR. Otherwise you'd have a duplicate line:
```java
if scene.camera:
    # make sure scene.camera.name is first
else:
    # insert "__CAMERA__"
if r3d and r3d.view_perspective != "CAMERA":
    # insert "__CAMERA__"
```

GoranM:
> If your mouse is over a 3D View area and P is pressed, that particular 3D View is replicated. Otherwise if the Export and Run button is used, the 3D View of the largest area of the current screen is replicated. If none is found and there is an active camera in the scene, this camera is used. Otherwise the largest 3D View area of the 'BDX' screen is replicated, if present, else an exception is raised: "No active camera or 3D View data in 'BDX' screen was found."

Sounds good.

> Not exactly, because of the OR.

I thought we just wanted to use the default view camera when there is no active camera, but I guess you're trying to make it work like in the BGE, where it's always using the default unless you're in camera mode?

If so, then yea, that makes sense.

rafcolson:
>Sounds good.

I'm afraid it won't be possible though. I'm still searching for a way to get the "hovered on" region.

>I thought we just wanted to use the default view camera when there is no active camera, but I guess you're trying to make it work like in the BGE, where it's always using the default unless you're in camera mode?

Yes, that was my intention. I assume that would be okay?

rafcolson:
PR adjusted.

GoranM:
Instead of `get_3d_view_region_data()`, perhaps `region_3d()` would be a more fitting name?

--

If a region is found, the method should just return it:

```python
r3d = bpy.context.region_data
if r3d:
    return r3d
...
```

If you need to check on view_perspective, you can do it outside.

--

I don't think you should throw exceptions inside the method. The method should just return the relevant region_3d reference, or None.

You can then throw a single exception in `camera_names()`:

```python
if not scene.camera or r3d and r3d.view_perspective != "CAMERA":
    if not r3d:
        raise Exception("No active camera or 3D View data")
    cam_names.insert(0, "__CAMERA__")
```

--

There's no need to import the operator module - You can use a lambda to sort:

```python
key=lambda lst: lst[1]
```

rafcolson:
>If a region is found, the method should just return it:
```java
r3d = bpy.context.region_data
if r3d:
    return r3d
...
```
If you need to check on view_perspective, you can do it outside.

It should only return r3d if view_perspective is not "CAMERA" else None. As you said:

>The method should just return the relevant region_3d reference, or None.

GoranM:
I don't really understand what you mean: A region_3d is a region_3d, regardless of the view_perspective value, isn't it?

rafcolson:
Yes. I'll try to adjust the code in a few days.

rafcolson:
This what I could do. Since we want BDX to use the active camera whenever pushing "export and run" (whether context has region data or not) we should check for an active camera (with ```scene.camera``` or ```view_perspective```) inside the method. The method should only return 3D View Region data if relevant. So ```relevant_region_3d_data()``` would be a more descriptive name for it.

As suggested I took the Exceptions out of the method and put one in ```camera_names()```. I also used lambda instead of the operator module.

I'm sorry I didn't comply to all of your suggestions, but in my opinion (or to my knowledge) it would make things more complicated in order to preserve equal results.

rafcolson:
When running BDX in camera view (or when pushing 'export and run' button) while the camera is not in an active layer, BDX will fail. So I've added an Exception message for this case: ```"No active camera in active layer(s)"``` This should eventually solve the the second part of the problem mentioned by SolarLune [here](https://github.com/SolarLune). Notice that when not in camera view BDX will be able to run with the generated camera. This allows previewing in whatever layer.

Issue #424, opened 2015-08-20
Suggestion: Input Map Replays


SolarLune:
Hey, there. So, I was thinking that input map replays would be a cool feature to add. Some games have an attract mode / demo mode, where the game plays itself while you watch. Other games have special powers or abilities that similarly hinge on replays (think of the "Clone" weapons from the Megaman games, or the time traveling recording mechanics of games like Braid). Cinematics can also be the product of replays, theoretically (the developer creates the cinematic by moving the characters around one by one, for example).

Adding replaying capabilities to the input maps would allow this to be done, and it should keep the actual replay data relatively simple - rather than a video, for example, the data would just be the names of the inputs being pressed.

The idea is to keep it simple with just a few functions, at least initially. 

`startRecording()` - Begins recording the specified replay data. This doesn't advance the recording process - the developer does this using the `frame()` function below.

`stopRecording()` - Stops recording the specified replay data.

`frame(int frameNum)` - Sets the frame for the replay data. This function alone should allow for various effects should the developer so choose, such as playing, stopping, speeding up replay input sequences, or slowing them down. When recording, this function seeks the recording and allows the developer to record inputs. When not recording, this function plays back the inputs that are stored within the recording.

`frame()` - Returns the current frame number.

`load()` - Loads recording data to be played or saved. Either an array of arrays of Strings (inputs) can be provided, or a file name can be provided.

`save(String replayDataName, String outputFileName)` - Saves the specified recording data to a text file.

`currentInputs(int frameNumber)` - Returns an ArrayList of the currently pressed inputs for the specified frame. If no frame number is given, it defaults to the current frame.

`on(Boolean on)` - Whether the replays are actively pushing inputs (on), or if the replay's just going through the process without affecting controls (off).

`ignore - ArrayList<String>` - Allows you to set certain inputs to be ignored.

I don't think this would support inputs with more finesse, like analog sticks or pressure-sensitive triggers, but it'll get the job done for simple "on-off" controls. In the future, this might be able to be expanded to have a "deep copy" mode that will allow it to work with actual clones of inputs for maximum accuracy at the cost of increased memory usage.

To facilitate replays, some sort of `push()` function that would allow you to push a result to an input in the InputMap will most likely need to be added so that the replays can push inputs when they're played back.

I suppose a Replay would be a class instance, not a static class that works off of bdx directly, for example.

GoranM:
[...]

PR #429, opened 2015-08-30
(#424) - Input map replay support.

SolarLune:
Added force____() functions to force input states to InputMaps.


GoranM:
It might be worth the effort to optimize a bit, by making Logger extend HashMap instead of Array, and storing only frames that have some active inputs, so we don't have to store a bunch of `[]` entries.

The type annotations in the json are taking up too much space, relative to the actual data we need to store. It's true that we have an abundance of memory, but we don't have to waste all that space, so we shouldn't.

Make it store and load plain json. I think you can do that via `setElementType` as they demonstrate on their wiki.

Issue #434, opened 2015-09-08
Suggestion: modify game.blend and remove redundant line


rafcolson:
Just some cleanup.

The startup blend file can use some adjustments.
- remove Game Properties area, since those are accessible through the Object panel now (and we probably want to encourage that)
- move BDX to the top of the Object panel (for every screen)
- remove Video Editing screen and show grid in 3d View as was suggested by SolarLune

Removal of redundant line in ```createproj.py```.

Edit: added two other things.

GoranM:
Sounds fine.

You should also re-enable display for parent-child relationship lines, assuming there's no good reason to keep them hidden.

rafcolson:
Okay. Do you like those these relationship lines in all screens? Note that I added some other proposals too. (Sorry about that).

rafcolson:
Also changed Player Resolution to 1024 X 576. Is that okay?

rafcolson:
PR merged. Thanks again.

PR #435, opened 2015-09-08
(#434) game.blend changes and removal of redundant line

rafcolson:


rafcolson:
Changed resolution to 1024 X 576 too.

GoranM:
Why did you change the resolution?

rafcolson:
I don't know exactly which commit this was discussed before, but we (SolarLune and myself) preferred a more standard resolution.

Issue #444, opened 2015-10-01
Suggestion: 2D Screen Filter Downsampling


SolarLune:
Hey, there. 

So basically, I think being able to down-sample / "render smaller and then upscale" 2D filters would be a nice feature to add. For some context, I'm working with the Bloom filter in my game, and trying to get it to be smooth and run quickly is just a bit difficult, which reminds me of the same problem with Bloom filters in the BGE, haha. 

So, I'm thinking that it should be easier on the GPU to render the bloom shader small, and then upscale it later - I believe this is simple downsampling.

As a guess, this would require a couple of changes. One would be to make a `size` or `resolutionPercentage` property that would set the size of the FrameBuffer the ShaderProgram uses (i.e. resolutionPercentage = new Vector2f(0.5f, 0.5f) to set it to be half the size of the screen). I'd put it on the ShaderProgram because every ShaderProgram can theoretically have different resolution percentages (i.e. a Bloom filter might be downsampled while a Tint filter would probably be run at full resolution).

The other change would be a simple boolean variable called something like `overlay` on the ShaderProgram. This would indicate whether BDX should draw the outputted framebuffer straight:

`Scene render > Effect > Output`

... or overlay it on the previous, unaltered render output:

`Scene render > Effect > Scene Render + Effect > Output`

This way, the "low quality" bloom filter result would be overlaid on top of the normal, high-fidelity game render for the final output for that ShaderProgram.

This is to present the idea - if it's OK'd, then I'll look into it to see if it'll work / if it actually does improve things GPU workload-wise. I'm not sure if simply changing the size of the frame buffer would work, or if we would actually need to render it small for the downsampling to take effect. I guess I'll see when I test it.

SolarLune:
OK, I think I've made some headway into this.

I've added some changes to the API to facilitate downsampling. Here's an example of the API currently.

```java
ShaderProgram shader = ShaderProgram.load("bdx/shaders/2d/default.vert", "bdx/shaders/2d/bloom.frag");
    
        shader.renderSize.x = 0.25f;  // Downsample to 25% of the size of the screen
        shader.renderSize.y = 0.25f;
        shader.overlay = true;            // Overlay the shader over the original screen capture (so the shader should have an ending alpha color of less than 1 to see the effect over the original).
        
        ShaderProgram.nearestFiltering = false;  // Filter all RenderBuffers linearly rather than with nearest filtering.
            
        scene.filters.add(shader);
```

[...]

PR #445, opened 2015-10-03
(#444) - Down-sampling for 2D Filters.

SolarLune:


GoranM:
`renderSize` implies a set size, not a scaling factor. I think `renderScale` would be a more appropriate name.

--

Instead of having an `overlayBuffer`, couldn't you simply leave `frameBuffer` uncleared?:

```java
if (!filter.overlay)
    frameBuffer.clear()
```

That would just draw the `tempBuffer` (downsampled image) over the original rendering already in `frameBuffer`, wouldn't it?

--

Having setSize return a brand new buffer is really confusing. If you need to create a new buffer, do it with `new RenderBuffer(...`.

SolarLune:
Fixed all of those. 

Thanks for the tip on leaving frameBuffer uncleared - I didn't want to add a new RenderBuffer, either, haha.

GoranM:
We should probably shorten those long lines:

```java
int width = Gdx.graphics.getWidth() * filter.renderScale.x;
int height = Gdx.graphics.getHeight() * filter.renderScale.y;
if (tempBuffer.getWidth() != width || tempBuffer.getHeight() != height)
    tempBuffer = new RenderBuffer(spriteBatch, width, height);
```

SolarLune:
Updated!

Issue #446, opened 2015-10-04
Bug: Profiler display (and overlay scene camera) wrong with User camera


rafcolson:
You might have noticed this already. But you can test this by creating a new project and change the view to User before running the game for the first time. Enabling Show Framerate and Profile shows that the Profiler is not viewed through ```"__PCam"```, which is the Profiler orthographic camera, but through the User camera.

Solution is to not add ```"__CAMERA__"``` to the Profiler's scene cameras.

rafcolson:
PR submitted.

rafcolson:
PR updated.

rafcolson:
PR updated.

rafcolson:
Thanks for the merge.

PR #447, opened 2015-10-04
(#446) Bug: Add "__CAMERA__" only in context scene

rafcolson:


GoranM:
There's no need to pass the scene reference around - You can just use the [scene global](https://github.com/GoranM/bdx/blob/master/blender/bdx/exporter.py#L728).

rafcolson:
I missed that. Thanks.

Issue #438, opened 2015-09-28
Transparent Objects Show Lower Scene


SolarLune:
Hey, there. So this is a bug that I happened upon while solving another bug. 

Essentially, transparent objects (or objects / sprites with transparent pixels) show the scene below when a shader is used on the scene. 

As far as I can tell, it's not the shader, as I bypassed actually drawing the scene using the shader - it has something to do with drawing objects to the RenderBuffer and the blending mode on materials. I don't think it's how we draw the RenderBuffer to other RenderBuffers or the screen, as only transparent objects are the problem ones - opaque objects show up fine.

Perhaps I'll be able to get back to this to attach an example later.

GoranM:
> I don't think it's how we draw the RenderBuffer to other RenderBuffers or the screen, as only transparent objects are the problem ones - opaque objects show up fine.

Opaque objects would naturally everwrit

GoranM:
Oops, accidentally closed the issue.

Re-opened.

SolarLune:
OK, here we go.

https://www.dropbox.com/s/0cm2ye0bst3tn4r/ShaderTransparencyIssues.blend?dl=0

The example blend file should work with any project that has 2D filters in the bdx/shaders/2d folder. Comment out the `scene.filters.add` line and the object becomes

SolarLune:
Here's a screenshot showing the problem a bit

![screen shot 10-05-15 at 10 43 pm](https://cloud.githubusercontent.com/assets/4733521/10300981/96db8b2a-6bb2-11e5-85b4-5673913759ec.PNG)

I've sceenshotted a tweaked version of the example - As you can see, the background scene and the background cube is visible through Sacky, even though you should only be able to see the ground or Sacky.

GoranM:
I'll preface this by saying that I don't really understand the underlying details of how OpenGL actually hashes this out, so I could be wrong:

The blending functions, set via the BlendingAttribute, are `GL_SRC_ALPHA` for source, and `GL_ONE_MINUS_SRC_ALPHA` for destination. This means that that the following calculation is applied to produce the resulting pixels in frameBuffer (s = source, d = destination, r = result):

```
(sR*sA) + (dR*(1-sA)) = rR	
(sG*sA) + (dG*(1-sA)) = rG
(sB*sA) + (dB*(1-sA)) = rB
(sA*sA) + (dA*(1-sA)) = rA
```

[...]

PR #455, opened 2015-10-12
Fixing blending to be transparent across scenes. (Bug #438)

SolarLune:


GoranM:
There are merge conflicts. You'll need to rebase from current master, and resolve those before this can be merged.

--

I'm not really sure if this will work in all cases, because super.render will still call glBlendFunc (when BlenderAttribute changes), and that may override the glBlendFuncSeparate call you just made (at least for that frame).

I don't really care that much, but if you do, it's probably something you should test.

SolarLune:
I think that should do it.

Nah, I don't care too much about the singular frame switch - it's already not an ideal fix because other blending modes don't work correctly against other scenes, so it's just something to be revisited and corrected outright later. 

A theory that might be the solution could be that we would make our own blending attribute so that we can use glBlendFuncSeparate. I think it would involve doing our own sorting of blended / transparent objects ourselves, though. I didn't want to get that far into it, so I just went for a simpler fix.

Meanwhile, the core issue of transparency straight-up not working correctly when Filters are applied should be fixed, though, which is fine for now, I think.

Issue #437, opened 2015-09-12
Suggestion: Join objects tool to reduce GL draw calls


rafcolson:
We were discussing this before. I didn't seem like a good idea to everyone, but in this issue I'll try to show the benefits, more specifically on Android machines.

The initial idea is to join objects (that share the same model) into new objects. Those new objects would then replace the actual objects visually. The actual objects are now invisible. This for the sole purpose to reduce GL draw calls.

This would imply that the actual objects shouldn't be moved, rotated, scaled, ... as long as they are "joined".

I would like it to be part of ```Scene.java```, so the function could work as following:
```java
LinkedListNamed<GameObject> joinedObjects = g.scene.addJoined(LinkedListNamed<GameObject>, boolean hideOriginals)
```
By default ```hideOriginals``` will be ```true```.

That's it for now. When I'm finished, I'll let you know.

Any ideas, suggestions, remarks are welcome.

Edit: Eventually this could evolve into a greater solution, by updating ```joinedObjects``` every time one of the actual objects is transformed. However, I think that would be rather expensive, but I'll definitely check that out.

SolarLune:
This sounds great, though it seems like having a simple `GameObject.join()` function might be a bit more versatile, in particular if you want to update a joined object?

> Edit: Eventually this could evolve into a greater solution, by updating joinedObjects every time one of the actual objects is transformed. However, I think that would be rather expensive, but I'll definitely check that out.

Seems like the easiest way to confirm this, and to also add some utility, would be to have a `joinedMaster` GameObject reference that points to the joined "master" object. This way, you have some sort of relationship that you can trace back later. When one of the joined objects is transformed, they can make the master update their joined children.

The same could be done for the "master" object - it would have a `joinedSubs` list that would refer to the joined children.

An `unjoin()` function would also be nice to delete portions of the model later (the portions of the model added would be connected to the references of the GameObject used to join them; that way, you can join or unjoin a model to an object).

rafcolson:
>This sounds great, though it seems like having a simple ```GameObject.join()``` function might be a bit more versatile, in particular if you want to update a joined object?
...

[...]

PR #456, opened 2015-10-20
(#437) Joining functionality to reduce GL draw calls

rafcolson:


GoranM:
You're using `materialTransformed()` in only one place, so it seems like you could just inline it, and annotate the block with a comment, like you did for other steps in `updateJoinedMesh()`.

--

The naming seems a bit confusing: You have a lot of variables with "material" in their name (or some shorthand for material, like "mat", or "mt"), but you're not really doing anything with materials.

You're working with mesh vertices, so your variable names should reflect that.

--

It's probably not a good idea to get vertex data from scene json, because that doesn't actually represent active data on the Model.

I think it would probably be best if you get that data directly from the actual Model reference, [like we do in SpriteAnim](https://github.com/GoranM/bdx/blob/master/src/com/nilunder/bdx/components/SpriteAnim.java#L206-L209).

rafcolson:
PR updated. I've also added a ```RuntimeException``` as a guideline not to join more than 10922, or the maximum length of the short array would be exceeded.

Do notice that ```updateJoinedMesh()``` will also set all objects in ```joinedMeshObjects``` invisible, and that removing the joiner from the scene will set those to be visible. But removing objects from ```joinedMeshObjects``` wouldn't do that, so objects have to be set visible manually. I guess this is okay?

SolarLune:
I think it might be best not to set visibility at all, since that's something the developer can do with not too much additional effort, and it's not something that makes too much sense (it's only when the function is run that the objects are hidden, not when they're added). It's a minor issue, though, so I don't care much either way. 

Great work in seeing this through!

rafcolson:
I think I would agree that changing visibility shouldn't be done by the joiner itself, so I've changed that. I think @GoranM would think so too. PR updated.

Issue #468, opened 2015-11-10
Suggestion: Gamepad Index


SolarLune:
Hey, there. 

It'd be nice to have a gamepad index option to allow for specifying which particular gamepad the controls should map to. This would be useful for multiplayer, of course.

The only thing that makes this a bit more difficult is that input maps also need to be adjusted for multiple gamepads to work correctly. Maybe we could adopt an optional "g:CONTROL NAME:GAMEPAD INDEX" setup, where if there is a third argument, it would be the number of the gamepad index? So you'd have:

```java
Bdx.imaps.put("left", "k:left", "g:left"); // Left on the d-pad on the first gamepad, or perhaps on any gamepad?

Bdx.imaps.put("left", "k:left", "g:left:1");  // Left on the d-pad on the second gamepad
```

GoranM:
Looking at the changes that need to be made, in order to support multiple gamepads, adjusting input maps doesn't strike me as the difficult part ... Anyway:

The input device is designated by the first marker, so I think it would be best to incorporate the device index into that marker, to have something like "g1:left".

"g" should just be an alias for "g0".

SolarLune:
^ Huh, so you mean that "the hard part" is elsewhere? Perhaps in the Gamepad class itself?

The reason that I thought the gamepad index should come last is because the string is already delimited and split by colons, so it'd be relatively trivial to add the index at the end. But I guess it's easy to go with the gamepad index right after the first letter.

GoranM:
Well, there are several places where changes would need to be made, in order to have a `Bdx.gamepads` list (Bdx, Gamepad, GdxProcessor, and maybe a few others), so that seems like the bulk of the work to me - InputMaps is just mapping to that, so that shouldn't be too hard.

The ":" delimiter is used to separate the device from the input marker ("device:input"). If you have an additional gamepad, you have an additional device, so the device marker should be different.

That's why "g1:left" makes more sense to me than "g:left:1".

SolarLune:
So I'm taking a crack at this, but it's not really working - the gamepads aren't getting updated, though I can see they are being detected. Any ideas? I'm pretty sure I'm just forgetting to update all of the gamepads somewhere somehow, but input handling with LibGDX seems pretty complex (to me, anyway).

rafcolson:
Can't help. But as I recall I always had to attach the gamepad before starting the game, else it was not recognized. (same with bge)

GoranM:
[...]

PR #470, opened 2015-11-16
(#468) - Support for multiple gamepads.

SolarLune:
- New index field that indicates gamepads' "slot" number (if it's the first, second, etc).
- New numConnected() static function indicating how many Gamepads there are.
- Gamepads now implement Named and have a name() function to allow them to be differentiated by name rather than by hashcode.
- Bdx.gamepad is now an ArrayListNamed - Bdx.gamepads, to allow you to get gamepad by slot or by name.
- Input maps can now take gamepad index numbers in the descriptors for input actions (e.g. Bdx.imaps.put("jump", "k:left", "g1:left")).

### DIFF
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -3,6 +3,7 @@ package com.nilunder.bdx;
 import java.util.*;
 
 import com.badlogic.gdx.*;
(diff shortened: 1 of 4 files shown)


GoranM:
Is the numConnected method really necessary, given that Bdx.gamepads has a size method?

SolarLune:
Ah, whoops, that's right. Fixed!

PR #475, opened 2015-12-15
(#235) - Various Shader-related fixes and improvements.

SolarLune:
Fixed bug where lights weren't updating (like, set a light to Dynamic before this commit and watch it float mysteriously). To do this, we update the light data every frame, alongside the object logic.

Moved the ShaderProvider and BDX Shader into a new file (BDXShaderProvider.java).

Cleaned up ShaderProgram, removing references to Shaders, as the Program doesn't need to have it anymore.

Moved nearestFiltering setting out of ShaderProgram constructor to prevent it being re-set with each ShaderProgram instantiation.

Added canRender() function to BDXDefaultShader, fixing a bug where LibGDX would re-use custom shaders for default materials.

### DIFF
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -8,11 +8,9 @@ import com.badlogic.gdx.files.*;
 import com.badlogic.gdx.graphics.*;
 import com.badlogic.gdx.graphics.g2d.*;
 import com.badlogic.gdx.graphics.g3d.*;
-import com.badlogic.gdx.graphics.g3d.attributes.BlendingAttribute;
-import com.badlogic.gdx.graphics.g3d.attributes.ColorAttribute;
-import com.badlogic.gdx.graphics.g3d.attributes.IntAttribute;
-import com.badlogic.gdx.graphics.g3d.shaders.DefaultShader;
-import com.badlogic.gdx.graphics.g3d.utils.DefaultShaderProvider;
+import com.nilunder.bdx.gl.BDXShaderProvider;
+import com.nilunder.bdx.gl.RenderBuffer;
+import com.nilunder.bdx.gl.ShaderProgram;
 import com.nilunder.bdx.inputs.*;
 import com.nilunder.bdx.audio.*;
 import com.nilunder.bdx.utils.*;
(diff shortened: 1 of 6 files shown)


GoranM:
Your description implies (heavily) that this PR should contain 5 commits.

--

Is the new file really necessary? Assuming that it is, I'm not sure if it belongs in the utils package.

--

It seems that your IDE auto-inserts "Created by SolarLune on ..." comments. The commit log clearly shows who contributed what and when, so I think these comments are redundant, and should therefore be removed (there is one already in the code-base, in [ShaderProgram.java](https://github.com/GoranM/bdx/blob/dda7dc2de87bb3f571765ab2f8747512970d5900/src/com/nilunder/bdx/ShaderProgram.java#L12))

SolarLune:
> Your description implies (heavily) that this PR should contain 5 commits.

No, it's just one commit with everything in it. Do you want me to split this up into 5 separate commits? If so, how would I do that?

> Is the new file really necessary? Assuming that it is, I'm not sure if it belongs in the utils package.

For organization purposes and to maintain clarity of code-base, it seems logical to not just keep pushing things into Bdx.java, but it's just a place for code. It's up to you.

> It seems that your IDE auto-inserts "Created by SolarLune on ..." comments. The commit log clearly shows who contributed what and when, so I think these comments are redundant, and should therefore be removed (there is one already in the code-base, in ShaderProgram.java)

Ah, yeah, my mistake. It does that when I create new files. I'll take those out.

GoranM:
> No, it's just one commit with everything in it.

I didn't say that it was more than one commit; I said that your description implies that it should be.

> Do you want me to split this up into 5 separate commits?

You don't have to bother with it now, but in the future, it would be helpful if you could structure things in a way where logically separated changes are placed in their own commits.

> For organization purposes and to maintain clarity of code-base, it seems logical to not just keep pushing things into Bdx.java, but it's just a place for code. It's up to you.

Well, if it's just a few classes, I don't think it's that bad to leave it in, but a new file would also be fine.

I'm just not sure it belongs in the utils package. Perhaps it would be best to make a new package, containing all our gl-related classes (ShaderProgram, RenderBuffer, and BDXDefaultShader).

SolarLune:
> I didn't say that it was more than one commit; I said that your description implies that it should be.
> You don't have to bother with it now, but in the future, it would be helpful if you could structure things in a way where logically separated changes are placed in their own commits.

OK, I'll keep separate commits in mind for the next time.

> I'm just not sure it belongs in the utils package. Perhaps it would be best to make a new package, containing all our gl-related classes (ShaderProgram, RenderBuffer, and BDXDefaultShader).

OK.

PR updated!

PR #482, opened 2015-12-29
(#235) - Per pixel lighting commits.

SolarLune:
Scene.lights list added (not sure why I tagged it with the wrong issue).
Specularity supported and added in.
Lights weren't being properly removed before; they are now.
Rendering more lights by default is now supported.


GoranM:
The first commit should be changed to reference the proper issue.

--

The 3rd commit ("Fix to properly remove lights") should be included in the first commit.

--

We should probably just have one loop through the toBeAdded and toBeRemoved lists, so instead of this:

```java
		for (GameObject g : toBeAdded) {
			if (g instanceof Light)
				lights.add((Light) g);
		}
		for (GameObject g : toBeRemoved) {
			if (g instanceof Light)
				lights.remove(g);
		}

  		if (toBeAdded.size() > 0){
  			objects.addAll(toBeAdded);
  			toBeAdded.clear();
 		}
 		if (toBeRemoved.size() > 0){
 			objects.removeAll(toBeRemoved);
 			toBeRemoved.clear();
 		}
```

It would probably be better to have this:

```java
		for (GameObject g : toBeAdded) {
			objects.add(g);
			if (g instanceof Light)
				lights.add((Light) g);
		}
		toBeAdded.clear();

		for (GameObject g : toBeRemoved) {
			objects.remove(g);
			if (g instanceof Light)
				lights.remove(g);
		}
		toBeRemoved.clear();
```

Same for the loop in addInstances.

--

If it's easier for you to just make one commit, that would be fine in this case, because it's all basically tied to lighting.

SolarLune:
OK, PR updated, and it's all in one commit.

Issue #506, opened 2016-01-18
Suggestion: Profiler changes/enhancements


rafcolson:
There are a couple of things I'd like to change in order to expose variables ```Bdx.profiler.avgTickRate``` and ```Bdx.profiler.avgTickTime```, and to add an option to display variables of ```com.badlogic.gdx.graphics.profiling.GLProfiler```.

- splitting up the initialization process (by using class initialization ```{//initialization}```)
- updating ```avgTickRate``` and ```avgTickTime``` regardless if the profiler is visible or not
- additional methods to enable/disable display for "tick info", "processes" and "gl"
- additional method to scale the profiler (useful with small screens, eg. on mobile)

SolarLune:
This sounds OK to me, but I don't quite get what you mean by "splitting up the initialization process".

I'm going to probably cut the code that scales the profiler in accordance with the screen size (not your proposal to scale it using a method) when I get back to viewports, as I believe it was problematic. Just letting you know.

I like the idea of allowing the user to scale the profiler to size it up or down, though.

rafcolson:
> I don't quite get what you mean by "splitting up the initialization process".

It will be clear once I've finished the update.

GoranM:
Ok.

rafcolson:
PR submitted.

SolarLune:
Looks OK, I suppose, though if there's nothing in the new classes that are publicly accessible, we could just go with glVisible() and subsystemsVisible() functions, right?

rafcolson:
I'm not sure. We do want ```gl``` to be public to have access to gl specifics. Maybe we could put ```start()```, ```stop()``` and ```remove()``` in ```subsystems``` instead?

rafcolson:
PR updated.

rafcolson:
PR updated.

rafcolson:
PR merged. Thanks a lot guys, for your comments and suggestions. As always, this made the result much better.

SolarLune:
Great to hear, and great work on finishing this. Thanks for updating the API, as well; it'd be nice if you could explain what each function does in the GL module (i.e. what does "calls()" return).

rafcolson:
>it'd be nice if you could explain what each function does in the GL module (i.e. what does "calls()" return).

Updated.

SolarLune:
Cool. Thanks!

Issue #503, opened 2016-01-17
Bug: Successfully build *.apk crashes on Android with new advanced lighting.


rafcolson:
Just build an apk of the default blend file. This will give no problems. Also installing the apk will give no problems. But when starting the apk it will stop immediately. Here's a [test.apk](https://www.dropbox.com/s/gol10m3d3rhg4ag/test.apk?dl=0), which you can test for yourself.

GoranM:
What's the actual crash exception?

rafcolson:
I can't say. And I don't know if it's possible on Android to get such information. The app just stops with a message, like: "The app has stoppped".

GoranM:
Android should put everything into the system log, which can be accessed via [`adb logcat`](http://developer.android.com/tools/help/logcat.html).

rafcolson:
Can't reproduce any errors on ```Nexus 5 API23 x86 (Android 6.0)```. Everything seems to work fine.

GoranM:
You're not supposed to use CatLog, or any other "app" - As the article you linked to explains, apps are restricted in what system data they can actually access.

You should use the adb utility that comes with the android sdk (it's in platform-tools). You can use that to establish a connection between your PC and your android device, and then run the logcat command (the details should be on the page I linked in my previous post).

rafcolson:
Silly me. Didn't try to connect my phone to the PC yet, but I'll will do that first thing.

>One thing is sure. Initializing Sacky with ```Bdx.display.advancedLighting(false)``` will prevent this crash.

However, this is not true anymore, since Issue #504 was fixed. Probably it's because now both modes use customized shaders. So the bug's definitely in those.

rafcolson:
So here's what logcat logged:
```java
01-24 16:23:47.345 13424-13438/? E/AndroidRuntime: FATAL EXCEPTION: GLThread 934
 Process: com.raco.test.android, PID: 13424
 com.badlogic.gdx.utils.GdxRuntimeException: Fragment shader compilation failed.
 ERROR: 0:179: 'max' : no matching overloaded function found 
 ERROR: 0:180: '<=' :wrong operand typesno operation '<=' exists that takes a left-hand operand of type 'float' and a right operand of type 'const int' (or there is no acceptable conversion)
 ERROR: 0:181: 'assign' :cannot convert from 'const int' to 'float'
 ERROR: 0:195: 'max' : no matching overloaded function found 
[...]

PR #512, opened 2016-01-22
(#506) Profiler changes/enhancements

rafcolson:
Here's a ```Sacky.java``` to test with:
```java
import com.nilunder.bdx.*;

	private boolean test;
	private int caseIndex;
	
	public void init(){
		test = true;
		caseIndex = 0;
	}
	
	public void main(){
		if (test){
			Bdx.profiler.start("Test");
		}
		
		if (Bdx.mouse.btnHit("left")){
			if (caseIndex == 3){
				caseIndex = 0;
			}else{
				caseIndex += 1;
			}
		}
		if (caseIndex == 0){
			Bdx.profiler.subsystemsVisible(false);
		}else if (caseIndex == 1){
			Bdx.profiler.glVisible(true);
		}else if (caseIndex == 2){
			Bdx.profiler.subsystemsVisible(true);
		}else if (caseIndex == 3){
			Bdx.profiler.glVisible(false);
		}
		
        if (Bdx.keyboard.keyHit("t")){
			System.out.println(Bdx.profiler.avgTickRate);
        }else if (Bdx.keyboard.keyHit("e")){
			if (!Bdx.profiler.gl.isEnabled()){
				Bdx.profiler.gl.enable();
			}
        }else if (Bdx.keyboard.keyHit("d")){
			System.out.println(Bdx.profiler.gl.drawCalls());
        }else if (Bdx.keyboard.keyHit("p")){
			Bdx.profiler.gl.print();
        }else if (Bdx.keyboard.keyHit("space")){
			applyForce(0, 0, 300);
		}
		
        if (test){
			Bdx.profiler.stop("test");
			if (Bdx.keyboard.keyHit("enter")){
				Bdx.profiler.remove("test");
				test = false;
			}
		}else if (Bdx.keyboard.keyHit("enter")){
			test = true;
		}
	}
	
}
```
___

## API Bdx

<a name="profiler" href="#profiler">#</a> **profiler** - [Profiler](https://github.com/GoranM/bdx/wiki/API-Profiler)

BDX profiler. Show by enabling "Show Framerate and Profile" in Blender (in Game Info or Display panel of Render Properties).
___

## API Profiler

This class measures and displays info about the average tick rate and tick time, execution times for various subsystems and GL statistics.

User created subsystems can be timed and are displayed beneath internal subsystems. When subsystems and the GL profiler are set to invisible, only the tick info will remain visible.

<a name="gl" href="#gl">#</a> **gl** - Gl

Extends [GLProfiler](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/profiling/GLProfiler.html). Calculates and shows various GL statistics: the number of calls, draw calls, shader switches, texture bindings, vertex count and triangle count.

By default the GL profiler is not visible, and so disabled. Setting to visible will automatically enable the GL profiler, but setting to invisible will not disable it.

If the BDX profiler is not visible, statistics are only updated if the GL profiler is enabled.
___

<a name="scale" href="#scale">#</a> **scale**(float f)

Sets the scale.

<a name="subsystemsVisible" href="#subsystemsVisible">#</a> **subsystemsVisible**(boolean visible)

Sets visibility of subsystems.

<a name="glVisible" href="#glVisible">#</a> **glVisible**(boolean visible)

Sets visibility of the GL profiler.

<a name="start" href="#start">#</a> **start**(String name)

Stores the start time in nanoseconds of a subsystem.

<a name="stop" href="#stop">#</a> **stop**(String name) - Float

Returns the delta time of a subsystem between start and current time in milliseconds. Possible to time subsystems directly after each other without having to store the start times.

Example:
```java
Bdx.profiler.start("a");
float a = Bdx.profiler.stop("a");
float b = Bdx.profiler.stop("b");
```
<a name="remove" href="#remove">#</a> **remove**(String name)

Removes execution time. Use with adding and removing objects.

Example:
```java
public class BossA extends GameObject{

	public void main(){
		Bdx.profiler.start("BossA");
		// code
		Bdx.profiler.stop("BossA");
	}
	
	public void onEnd(){
		Bdx.profiler.remove("BossA");
	}
	
}
```
___

## API Gl

Extends [GLProfiler](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/profiling/GLProfiler.html). When enabled, collects statistics about GL calls and checks for GL errors.

<a name="listener" href="#listener">#</a> **listener** - [GLErrorListener](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/profiling/GLErrorListener.html)

By default the GL listener will throw a GdxRuntimeException with error name.

___

<a name="enable" href="#enable">#</a> **enable**()

Enables collecting and updating of GL statistics, and GL error listener.

<a name="disable" href="#disable">#</a> **disable**()

Disables collecting and updating of GL statistics, and GL error listener.

<a name="reset" href="#reset">#</a> **reset**()

Resets GL statistics.

<a name="print" href="#print">#</a> **print**()

Prints GL statistics.

<a name="calls" href="#calls">#</a> **calls**() - int

<a name="drawCalls" href="#drawCalls">#</a> **drawCalls**() - int

<a name="shaderSwitches" href="#shaderSwitches">#</a> **shaderSwitches**() - int

<a name="textureBindings" href="#textureBindings">#</a> **textureBindings**() - int

<a name="vertexCount" href="#vertexCount">#</a> **vertexCount**() - int

<a name="triangleCount" href="#triangleCount">#</a> **triangleCount**() - int

### DIFF
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -276,9 +276,10 @@ public class Bdx{
 		}
 		mouse.wheelMove = 0;
 		
-		if (profiler.visible){
-			profiler.update();
-
+		profiler.updateVariables();
+		if (profiler.visible()){
+			profiler.updateVisible();
+			
 			// ------- Render profiler scene --------
 			
 			Scene profilerScene = profiler.scene;
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -119,8 +119,9 @@ public class Bdx{
 		}
 
 	}
-
-	public static final float TICK_TIME = 1f/60f;
+	
+	public static final int TICK_RATE = 60;
+	public static final float TICK_TIME = 1f/TICK_RATE;
 	public static final int VERT_STRIDE = 8;
 	public static float time;
 	public static String firstScene;
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -293,6 +293,9 @@ public class Bdx{
 			}
 			modelBatch.end();
 		}
+		if (profiler.gl.isEnabled()){
+			profiler.gl.updateFields();
+		}
 	}
 	
 	public static void dispose(){
--- a/src/com/nilunder/bdx/utils/Profiler.java
+++ b/src/com/nilunder/bdx/utils/Profiler.java
@@ -9,68 +9,79 @@ import java.util.Map.Entry;
 import com.badlogic.gdx.utils.TimeUtils;
 import com.badlogic.gdx.Gdx;
 
+import com.nilunder.bdx.Bdx;
 import com.nilunder.bdx.Scene;
-import com.nilunder.bdx.Text;
 import com.nilunder.bdx.GameObject;
+import com.nilunder.bdx.Text;
 
 import javax.vecmath.Vector3f;
+import javax.vecmath.Vector4f;
 
 
 public class Profiler extends LinkedHashMap<String, Long>{
 	
-	private final int TIC_RATE = 60;
+	private final int TICK_RATE = Bdx.TICK_RATE;
+	private final Vector3f SCREEN_SIZE = new Vector3f(448, 448, 1);
+	private final Vector4f BG_COLOR = new Vector4f(0.125f, 0.125f, 0.125f, 0.5f);
+	private final float SPACING = 0.6f;
+	private final float BAR_HEIGHT = 0.4f;
+	private final float BAR_WIDTH = SPACING * 4;
+	private final float BAR_POSITION = SPACING * 18;
+	private final float BG_WIDTH = SPACING * 23;
+	private final String EXC_MSG = "User created subsystem names should not start with: \"__\"";
+	
 	private LinkedHashMap<String, Long> startTimes;
 	private long totalStartTime;
 	private long lastStopTime;
 	private LinkedHashMap<String, Long> nanos;
 	private LinkedHashMap<String, Float> percents;
 	private ArrayList<String> names;
-	private ArrayList<Long> ticTimes;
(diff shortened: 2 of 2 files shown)


rafcolson:
I'll have to make some adjustments in order for it to work properly. If I'm sure it's finished I'll give a sign. Sorry.

rafcolson:
Finished. There's an unforeseen problem with the GLProfiler though. When enabled I'm guessing the ```listener``` is causing performance issues on Android (with emulator, I couldn't test with actual phone due to bug #503). Maybe you guys still remember I did some testing with gl draw calls before BDX was updated to the new LibGDX. Then the GLProfiler hadn't a listener and wasn't causing any issues. Or maybe it's the emulator...

GoranM:
I don't really understand the rationale for adding "subsystems" - It just seems like an additional level of indirection, so I'm now sure how that would be better than simply having those methods directly on the profiler.

rafcolson:
I thought this way the Profiler class would be more readable, keeping things separated. But I agree that this is not necessary, so I'll change it back.

rafcolson:
PR updated. ```Bdx.profiler.gl``` now has methods to get the actual calls, draw calls, vertices, ... This will be useful for developing for Android, for which the developer may want to get these stats without showing the profiler, because showing would slow down performance due to the many objects in the profiler.

Another improvement is that the profiler will not count the gl stats of the profiler itself. So we get clean numbers.

I'll update the API in my first post soon.

SolarLune:
Ah, a function to print out the profiler information might be a nice convenience function to stand in for low-performance devices. It'd print out the GL information if you have it enabled, etc.

The API looks OK to me, anyway.

rafcolson:
Okay, coming up.

rafcolson:
PR updated.

Issue #524, opened 2016-02-02
Sugestion: Swap model textures


Fr0sk:
I think it would be nice to have a way to swap a model texture with another by code, so we can give some customization to our players. Right now the only way I can think of is to create a object for each texture I have and swap it with the current object being rendered. Also (don't know if should be here) is it impossible to have texture files in folders inside the textures folder? Because in mid/large projects it can get a mess preaty soon...

SolarLune:
> I think it would be nice to have a way to swap a model texture with another by code, so we can give some customization to our players. Right now the only way I can think of is to create a object for each texture I have and swap it with the current object being rendered.

Yeah, that would be nice to have. Currently, like you mentioned, textures are kinda locked to materials, and materials are locked to models under BDX. However, being that we're using LibGDX under the hood, it might be possible to do manually.

Anyway, it'd be nice to alter these two aspects in-game via code, at least for replacing the material on a model, initially.

> Also (don't know if should be here) is it impossible to have texture files in folders inside the textures folder? Because in mid/large projects it can get a mess preaty soon...

That should be possible, but you're right. It looks like textures are expected to reside directly in the textures directory at the moment. That should be pretty easy to add in, though.

Fr0sk:
> Yeah, that would be nice to have. Currently, like you mentioned, textures are kinda locked to materials, and materials are locked to models under BDX. However, being that we're using LibGDX under the hood, it might be possible to do manually.

> Anyway, it'd be nice to alter these two aspects in-game via code, at least for replacing the material on a model, initially.

Well, I'm  guessing that under the hood BGX is just creating libGDX modelInstances of the models, if so, at least provide a way to access the modelInstance and that would make possible to swap either materials or textures. Don't know if I'm right though... 

> That should be possible, but you're right. It looks like textures are expected to reside directly in the textures directory at the moment. That should be pretty easy to add in, though.

About this, I think that if the textures are easily swappable this might be easy to overcome, since code is not (or at least shouldn't be) locket to one dir and no subdirectories, but yeah, it would be a really nice addition.

SolarLune:
> Well, I'm guessing that under the hood BGX is just creating libGDX modelInstances of the models, if so, at least provide a way to access the modelInstance and that would make possible to swap either materials or textures. Don't know if I'm right though...

[...]

PR #525, opened 2016-02-02
(#524) Material change functions.

SolarLune:


GoranM:
I don't think you have to do anything with m_ps. With m_verts ordered, it should be possible to change:

```python
for m, tris in m_tris.items():
```

to

```python
for mat in mesh.materials:
    m, tris = mat.name, m_tris[mat.name];
```

And that should just work.

SolarLune:
I just tried this, but it doesn't work if a material is used that vertices aren't using.

GoranM:
You mean there is a mat.name which doesn't exist as a key in m_tris, and you get a key error?

You can just skip over those - If they don't map to any vertices, they are not actually used, so it shouldn't be a problem.

SolarLune:
> You mean there is a mat.name which doesn't exist as a key in m_tris, and you get a key error?

Yeah, thanks.

Your proposed solution didn't work for meshes that have no materials, so I added that into srl_models, as well.

PR updated.

GoranM:
Do you want me to make a new release now?

SolarLune:
Yeah, sure. Thanks for the merge.

Issue #474, opened 2015-12-08
Suggestion: Alpha Function


SolarLune:
Hey, there! I was wondering what you guys thought of convenience alpha() function. It would just set the alpha of the object by calling the color function, but with the object's color on all parameters except for the alpha. For example this:

```java
Vector4f c = color();
c.a = c.a + 0.1f;
color(c);
```

could become this:

```java
alpha(alpha() + 0.1f);
```

Thoughts?

rafcolson:
Close to redundant, but I would say yes.

SolarLune:
You know, I don't really like the redundancy either, but I do like the alpha function. What if the color function didn't allow you to set the alpha component, and it just altered the color of the object? They are two separate concepts (the color of an object, and the transparency of an object); they just happen to be stored together in a single value (the diffuse color attribute). Maybe the color() function could just take and return Vector3fs?

I mean, when you think about it, you can't, for example, fade an object's color to black by doing this:

```java
color(color().mul(0.1f));
```

... Because it will also alter the object's alpha component. Maybe we should also adjust the color functions to just deal with the diffuse color, and not the alpha value?

rafcolson:
That makes sense to me. I personally would prefer these functions:
```color()``` - Vector3f rgb
```color(Vector3f rgb)```
```color(float r, float g, float b)```
```color(Vector4f rgba)```
```color(float r, float g, float b, float a)```
```alfa() - float a```
```alfa(float a)```

SolarLune:
Eh, I dunno. I already removed a couple of the convenience color functions in favor of just having a few; I think I'd rather not have additional functions if possible.

GoranM:
Sounds fine.

SolarLune:
OK, so what do you think, @rafcolson? Do you think there can be too many convenience functions?

rafcolson:
Well, I think we could have as many, as long as there are use cases.

I would prefer to also have the current (Vector4f) Color method. I don't mind if we don't have those with the rgb(a) arguments.

SolarLune:
> I would prefer to also have the current (Vector4f) Color method. I don't mind if we don't have those with the rgb(a) arguments.

I'm not sure if I understand - can you clarify?

rafcolson:
Sorry.

I meant that this is okay for me:

color() - Vector3f rgb
color(Vector3f rgb)
color(Vector4f rgba)
alpha() - float a
alpha(float a)

SolarLune:
[...]

PR #534, opened 2016-02-16
(#474) Alpha function.

SolarLune:
Also, different convenience functions for color().
Also, now only overriding colorNoChildren for Lights, so they should work correctly with the GameObject's color() functions.


SolarLune:
Closing in case a better solution can be found (see #474).

Issue #533, opened 2016-02-15
Suggestion: Minor Text Improvements.


SolarLune:
Hey, there. I think we could make a few improvements to Text objects.

1) As is documented, you need to fix the scaling on text manually as the size of text objects in BDX is smaller than the size of text objects in Blender. Is this always a certain amount? If so, we could make BDX just automatically fix this, right?

2) We could stand to change the `text` attribute and `set()` function to just a couple of `text()` functions - one to set the text, and one to get the text.

3) We could export the string that text objects are displaying in Blender to display them in BDX. This is an issue because if you get the `text` property of Text objects before setting them in BDX, it'll crash BDX because `text` is null.

4) We could add a `capacity` integer value that indicates how many characters the text object can display.

SolarLune:
Updated the issue with more suggestions.

GoranM:
> As is documented, you need to fix the scaling on text manually as the size of text objects in BDX is smaller than the size of text objects in Blender. Is this always a certain amount?

As far as I remember, no - It seemed to vary based on some characteristic of the given font, and how it's interpreted by the whole Hiero system.

> 2) We could stand to change the text attribute and set() function to just a couple of text() functions - one to set the text, and one to get the text.

That would be fine.

> 3) We could export the string that text objects are displaying in Blender to display them in BDX.

We already export the relevant data, so you should be able to just use it.

> 4) We could add a capacity integer value that indicates how many characters the text object can display.

Ok.

SolarLune:
> As far as I remember, no - It seemed to vary based on some characteristic of the given font, and how it's interpreted by the whole Hiero system.

Oh, really. Dang, that's unfortunate. I thought it was always the same number (0.72).

> We already export the relevant data, so you should be able to just use it.

Not sure what you mean - I mean that the `text` field is blank on Text objects until you use set() to set the text manually. Right?

GoranM:
> Do you mean that the string that the Text object is displaying in Blender is exported?

Yes.

SolarLune:
If we export the string, and we can set the text object to the string from Blender, do we need to export the text mesh's vertices from Blender? Could we just cut out the entire `vertices_text()` function?

GoranM:
Setting the vertices to specific data is redundant (provided we set the text at initialization), but that doesn't remove the need for vertices themselves.

[...]

PR #543, opened 2016-02-18
(#533) Minor Text improvements.

SolarLune:
set(String) is now text(String).
Added text() function to read the text string.
Added capacity value to indicate how many characters the Text object can show.
Text object string now set to the string from Blender.


SolarLune:
Sorry about that - had to rebase to fix some conflicts. Should be good to go now.

Issue #537, opened 2016-02-17
OutOfMemoryError when number of mesh vertices > 32767


aliasaila:
libgdx doesn't support more than 32767 vertices within a single mesh, so there needs to be some additional handling for when this limit is exceeded.

[https://github.com/libgdx/libgdx/wiki/Importing-Blender-models-in-LibGDX#maximum-vertices](https://github.com/libgdx/libgdx/wiki/Importing-Blender-models-in-LibGDX#maximum-vertices)

Fr0sk:
Maybe we can throw a pop-up warning explaining the situation and asking the developer to either use a decimate modifier, use a [mesh poly reducer script](http://wiki.blender.org/index.php/Extensions:2.4/Py/Scripts/Mesh/Mesh_poly_reduce) (both of these automatic) or manually strip down the mesh in several vertex groups. I don't recall if there is a way of doing the last one automatically, but maybe that would result in unwanted behavior. But this limit is not libgdx or bdx specific. Most engines have one of those limits specified on the link you posted, so maybe a warning would be just enough. Just my 2 cents though.

SolarLune:
Hm... Since the `short`'s limit only influences the number of vertices in the mesh when we're actually building the mesh, it might be possible to auto-split the mesh according to the number of remaining vertices to build (i.e. it creates one MeshPart per 32767 vertices per material, rather than just creating one MeshPart per material). 

This would mess with the material switching, though, as each MeshPart is currently treated as a separate material slot. This could be solved by simply giving each MeshPart name its material slot number. Then, when we want to change the material in a slot, we can loop through the MeshParts and change the material of the parts that have that material ID.

So basically, can we just automatically split up meshes that exceed 32767 vertices?

Fr0sk:
That would not break moving meshes and possibly bone animated meshes too?

SolarLune:
We don't have bone animated meshes, so I dunno about that. So maybe we should wait until we have animated bone support, I guess? To make sure it doesn't break it?

Fr0sk:
Since it's on a issue, I though it was somewhat on the way. If we create a empty holder object and then subdivide the mesh and parent each part to the holder, and the armature too, that would work?

SolarLune:
> Since it's on a issue, I though it was somewhat on the way.

[...]

PR #545, opened 2016-02-19
(#537) Popup error when vertex number exceeds 32767

Fr0sk:


GoranM:
Shouldn't this code be in the engine, rather than the exporter? The limitation is on the libgdx side - it's not a blender problem.

@SolarLune - Comments on the commit don't seem to be counted as comments in the PR, even though they're shown here. I would suggests restricting your comments to the PR, to ensure that all relevant parties receive a notification.

SolarLune:
Oh, I didn't know. OK, I'll keep that in mind. Thanks!

Fr0sk:
Sorry, today I'm on a birthday and don't have the time to review the commit to change it. As @SolarLune, I also think this error should be thrown in Blender, as it happens when someone change a mesh in a way that exceeds the 32767 vertices. Since a mesh change to be applied has to be exported through blender, I think that preventing exporting to finish when that happens is better. 

> Basically, we're already looping through each object elsewhere - there's no need to do it again in execute().

I coded this check here because I think all the checks should be made before any of real work begin. That's just the way I code, but this project isn't mine, of course I will do it the way you guys decide :) 

But to be clear when I pick this up tomorrow, the check stays on the blender side? Should I put it on another object loop?

GoranM:
It seems more elegant to simply catch the OutOfMemoryError where it actually happens, and throw our own custom error instead, but if you want to do it in the exporter, that's fine too.

rafcolson:
>It seems more elegant to simply catch the OutOfMemoryError where it actually happens

I would agree. This is consistent with any error messages during runtime. @Fr0sk It's up to you, but here's the commit if you think it's fine either way.

SolarLune:
Yo @Fr0sk, you still there about this?

Fr0sk:
Hey @SolarLune. It seems that I got caught up with my university work and have little to no time to get into this side projects, so my apologies. I think @GoranM and @rafcolson are probably right, so it might be better to throw it where it really happens. If anyone wants to pick it up, feel free ;)

rafcolson:
Thanks @Fr0sk. PR submitted.

rafcolson:
@GoranM Could we close this? @Fr0sk probably isn't around to do so..

GoranM:
Ok.

Issue #469, opened 2015-11-15
Suggestion: Viewports


SolarLune:
Hey, there. 

So, it'd be nice if we had viewport support, since multiplayer games would benefit from this. The idea's pretty simple - we'll be able to set up a viewport and indicate where that viewport should be rendered on-screen (simply, its position, width, height, and maybe rotation, if the sprite batch supports it easily enough), and which camera is used to render to the viewport. Viewports also have clipping modes, but I'll save those for later.

The engine would probably have a default viewport that is rendered to if there *isn't* one already in the collection, so that we can save on code and also make it easy to specify split-screens (so that we're not wasting time rendering the game normally, and then also in two or more viewports).

GoranM:
Ok.

SolarLune:
OK, so I guess we could do this with either the built-in LibGDX viewports, or perhaps doing it ourselves by simply changing how and where we draw the finished texture after the shader process. I tried using the LibGDX Viewports, but it seems like it changes how the camera renders the scene? It seems like we're manually touching the camera in the scene's update loop, which might be what the problem was, as I read that the LibGDX Viewport updates the camera for us. I don't know, maybe I'll give it another go later to see if maybe it was something I did wrong. 

Anyway, in case it doesn't work out with LibGDX's Viewport, I'll also see about giving a custom Viewport method a try. I don't suppose it'd be TOO much more code to get it working. I'm thinking of a custom Viewport class (which is associated with a camera, to render out to that Viewport), and a Viewport collection per scene (so you can have multiple Viewports for split-screen, character renders, etc). You create a Viewport, provide it with a camera to render (null meaning the scene's camera), and then add it to the stack. Seems simple enough.

As for how it scales or stretches... Well:

- Letterbox is good because it maintains aspect ratio and shows the entire rendered camera output.
- Scaling is good because it doesn't have black bars (which can be annoying).
- Extending is good because it keeps the game viewed the same no matter where it's viewed or how big the scene is. However, I think we would have to change how the camera renders to account for this (as it can extend the area the camera's view itself captures), so this one's probably out, at least at first?

So maybe we should support Letterbox and Scaling at first?

The only thing that could be a bit weird would be the project() function on the camera; being that there could be multiple viewports, I'd guess we would alter it to take a viewport index. If one isn't provided, it'll assume you mean the default camera and the entire screen. Otherwise, it'll work with the Viewport's defined size and camera.

Thoughts?

GoranM:
[...]

PR #546, opened 2016-02-21
(#469) Further work to construct...

SolarLune:
... The camera in LibGDX, and use that, as well as expose relevant functions in our Camera class.


SolarLune:
The update() method works by calling the update() function of whatever object data is at the time. Since LibGDX's Camera class has an update() function, Java compiles that fine, but since it's in, actuality, either a PerspectiveCamera or OrthographicCamera, their update() function would override the Camera class's. The FOV appeared to work fine, when I tested it.

But anyway, go with getting the data from the projection matrix, like Goran mentioned, if it worked. That apparently is the only way to accurately get the data from Blender, so go with it. You could also still export whatever information that's not present in the Projection matrix, if that's an issue (i.e. the orthographic camera data).

Anyway, I'll close this now.

Issue #547, opened 2016-02-22
Suggestion: OnRender()


SolarLune:
Hey, there. So, basically, the issue that I'm seeing is that drawLine() functions don't work because they execute before the game world is drawn. So, to rectify this, I suggest an `onRender()` function that runs in GameObjects and Components after the scene is rendered. This way, any drawing logic can run afterwards.

An alternative solution would be to drop the `scene.update()` call to after the rendering process, but that would mean that changes take effect one frame after the game renders. I'm not a fan of this solution.

Anyway, thoughts?

GoranM:
You could have drawLine() buffer commands, and then execute them later.

SolarLune:
Hmm, that might be a better idea, yeah. I'll give it a shot.

SolarLune:
Seems to work fine. PR submitted!

SolarLune:
This was merged, so I'll close this now. Thanks!

rafcolson:
I don't think it's merged yet, is it?

SolarLune:
Ah, it's not - whoops! Oh well; guess I'll re-open this until it is?

SolarLune:
OK, now it was merged, so I'll close this, haha.

PR #549, opened 2016-02-23
(#547) Moved drawing functionality to a queue system.

SolarLune:
Now drawing happens after the scene is finished rendering (so you can see the draw functions' results).


SolarLune:
Eh, if there's no draw commands in the array list, then it won't run the for loop; all it'd do is just clear the drawCommands list, so I think it'd be fine.

PR #550, opened 2016-02-23
(#469) Camera.data -> CameraPerspective or CameraOrthographic

rafcolson:
Better put this on hold. The clipping values for both camera types seem not correct.

Edit: The clipping values seem fine now, but there's still a problem. If you would zero out the rotation of the camera in the default scene and move along the Y-axis so you're just above sacky and run BDX, then you won't see a thing. This only happens if the rotation on X and Y are zero. Any idea why this is happening? Must say these Blender to LibGdx camera is not my thing. Even when you think you almost got it, then...

Please, help.


GoranM:
It seems like you're still exporting and using data that doesn't correspond to values that are actually encoded in the blender projection matrix ...

I already explained why this causes problems in the issue thread, in response to SolarLune's latest attempt.

The data required by the libgdx camera needs to be extracted from the blender projection matrix, which we already export, not from additional export values like "angle_y".

There should be no need for the exporter to change.

rafcolson:
>It seems like you're still exporting and using data that doesn't correspond to values that are actually encoded in the blender projection matrix ...

I'm sorry this was somewhat of a disappointment to you. I'll do as you ask and get the values from the projection matrix (except for zoom).

rafcolson:
Unfortunately the issue with the camera's rotation remains. Feel free to help a hand.

GoranM:
I think you need to extract zoom as well; If there's a discrepancy between "angle_y" and matrix encoded fov, there's probably a discrepancy between "ortho_scale" and matrix encoded zoom.

Also, note that libgdx cameras have this silly process in which they use an "up" vector, along with a "direction" vector to calculate the view matrix, so you'll need to extract those from the inverted camera transform (that is essentially the view matrix, which libgdx cameras need to re-derive with their convoluted method).

rafcolson:
>I think you need to extract zoom as well; If there's a discrepancy between "angle_y" and matrix encoded fov, there's probably a discrepancy between "ortho_scale" and matrix encoded zoom.

Did you notice there was a discrepancy? I want to extract zoom, but wouldn't know how to do this. Is there some documentation I could look into?

> Also, note that libgdx cameras have this silly process in which they use an "up" vector, along with a "direction" vector to calculate the view matrix, so you'll need to extract those from the inverted camera transform (that is essentially the view matrix, which libgdx cameras need to re-derive with their convoluted method).

Thanks for the tip.

GoranM:
> Did you notice there was a discrepancy?

I didn't test it, but logically, since we're talking about values that affect the field of view, it should be visible.

You can test it with the same procedure I described in the issue.

> I want to extract zoom, but wouldn't know how to do this. Is there some documentation I could look into maybe?

I don't think blender's "ortho_scale" corresponds to the libgdx idea of "zoom" (which is 1 by default), so there's no direct documentation that I know of.

However, you can read the libgdx source code, to figure out how they calculate the projection matrix for an ortho camera (start in [OrthographicCamera.update](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/OrthographicCamera.java#L55)), so you can do some math to reverse the process, and extract the relevant values from our blender exported projection matrix.

SolarLune:
I _believe_ Blender's ortho_scale can be set up to work directly the same with the zoom field in OrthographicCameras if you set the orthographic camera's viewport to be a size of (1, height / width).

rafcolson:
>I didn't test it, but logically, since we're talking about values that affect the field of view, it should be visible.
>
>You can test it with the same procedure I described in the issue.

I had tested this and didn't see a discrepancy with zoom.

>I believe a value provided to Blender's ortho_scale will work the same with the zoom field in OrthographicCameras if you set the camera's viewport to be a size of (1, height / width).

That would make sense because currently ```OrthographicCamera``` zoom is calculated like so: ```zoom = ortho_scale / width```.

@GoranM: If this seems to be the case, is it okay to leave it like that?

rafcolson:
I also grouped the operations in ```Scene``` for object initialization, also removed the redundant if for fonts. I managed to fix the rotation issue by updating the up vector of the libgdx's camera with the realtime up axis. So everything works now. I checked if there's any discrepancy with ```ortho_scale / width``` again, but couldn't see anything wrong with it. It seems to be correct.

So I think this is up for a merger.

GoranM:
> If this seems to be the case, is it okay to leave it like that?

If ortho scale seems fine, I think it would be better to just extract it from the matrix:

```java
float ortho_scale = 2 / projection[0];
```

That way, we don't have to change the exporter.

rafcolson:
>That way, we don't have to change the exporter.

Thanks. All done now. The profiler works as it should. Note that with future viewport related commits the profiler should normally not need any adjustments.

Pfew. I'm glad that this is done.

GoranM:
Good work.

SolarLune:
Hey! Really nice work on this, yeah - a few nice new features, even forgetting about viewports. 

Just a reminder not to forget to document the new code that was added (like I've done before, haha). If you would rather, I can document it for you.

Thanks!

Issue #566, opened 2016-03-01
Suggestion: Index Arguments for Color(), Tint(), Shadeless(), blendMode()


SolarLune:
Hey, there. So currently, the color() setters work on all materials on the GameObject. This can be undesirable, so I propose an index argument that would specify which material to work on.

Here's the proposed API:

```java

- Getters -

color(int index) - Color
color() - Color - (shorthand for color(0))

- Setters -

color(Color color) - Sets the color for all materials on it and its children
colorNoChildren(Color color, int index) - Sets the color for just the object for the material in the specified slot
colorNoChildren(Color color) - Would color all materials for just the object

```

Note that color() can't be specified to use an index because it would do that for its children, too, which may not make sense.

EDIT: Oh, and pretty much the same thing for tint(), too.

EDIT 2: Might as well extend it to the shadeless functions, as well.

EDIT 3: BlendMode, too, I guess.

GoranM:
It sounds like a method that should be on the Material, not on a GameObject. Might be time to create our own subclass of libgdx material, with all the relevant methods.

SolarLune:
OK, I think I got this working, then. PR submitted. Here's a quick list of the changes.

- We have a Material class now. It extends the LibGDX Material.

- The color, tint, blendMode, and shadeless functions are now on the Material.

- I added two functions (`copyToMaterials` and `copyToGameObjects`) to allow the user to easily set the color / tint / alpha / whatever across multiple objects / materials.

- The GameObject material functions now take and return our Materials.

- The materials returned from the GameObject are stored on the GameObject itself (because each GameObject can have its own set of materials, of course).

- By default, the materials in use on the part and on the GameObject are references to the scene materials. This means that changing the values of one material changes them for all objects using them in the scene. This can be easily worked around by duplicating the Material and assigning it, using one of the GameObject's material functions.

rafcolson:
Nice. But why put ```Material``` in ```gl```? Couldn't it be in ```bdx```?

SolarLune:
> Also be sure to change line 306 in Profiler.java to background.materials.get(0).color(BG_COLOR);.

Thanks! Fixed. Now it should be more friendly to objects with "no material" (it assigns them the default material).

> Nice. But why put Material in gl? Couldn't it be in bdx?

Because it's a display related class. It's not on the same "level" as GameObjects, Lights, Cameras, etc. It's more akin to Textures, Sounds, Music, etc. An asset, and a supporting class. That's my thought, anyway.

[...]

PR #567, opened 2016-03-02
(#566) Material class.

SolarLune:


GoranM:
The copy method seems highly counter-intuitive. I think most people would expect to have a `material.copy()` that would return a new copy of the calling material (as is the case with the base class). Beyond being unconventional `material.copy(other)` is also very confusing, because it's not immediately clear if I'm copying from material to other, or from other to material. And then there's a question of what exactly is being copied; color, tint, blendMode and shadeless are not the only properties of a material.

Also, I'm not completely clear on why we need a "materials" list on the GameObject - couldn't we just use `modelInstance.materials`?

rafcolson:
>Also, I'm not completely clear on why we need a "materials" list on the GameObject - couldn't we just use ```modelInstance.materials```?

It should be noted that ```modelinstance.materials``` allows modification of materials without having to traverse the nodes, which is stated [here](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/g3d/ModelInstance.html#materials).

And so, if we want to be able to alter a material for all objects having that material, we should change the material on the model, through the model itself, or through ```GameObject.modelInstance.nodes```.

SolarLune:
> The copy method seems highly counter-intuitive. I think most people would expect to have a material.copy() that would return a new copy of the calling material (as is the case with the base class). Beyond being unconventional material.copy(other) is also very confusing, because it's not immediately clear if I'm copying from material to other, or from other to material. And then there's a question of what exactly is being copied; color, tint, blendMode and shadeless are not the only properties of a material.

Do you have a suggestion of another way, or another name for the function? I was trying to make an easy method to imitate the current ability to set the color on children of a GameObject, which is possible with the copy functions in the Material class.

> It should be noted that changing a material in modelinstance.materials doesn't affect other objects with the same materials.

> Edit: modelInstance.materials doesn't contain BDX Material's, so I think that's the reason why we would need such a field on GameObject.

Yep. Any materials read from the modelInstance materials list are LibGDX materials, and can't be cast to BDX Materials. The method I decided to go with was to store a list of BDX materials for reading, and give the ModelInstance our BDX materials when writing.

rafcolson:
>Do you have a suggestion of another way, or another name for the function?

```java
material = otherMaterial;
```
Would also make those not to be unique, right?

SolarLune:
Could you clarify / explain a bit what you mean, @rafcolson?

rafcolson:
>Could you clarify / explain a bit what you mean, @rafcolson?

It depends on what you mean by 'copying' the material. You could be replacing the attributes of the material (color, blending, ....) or you could be replacing the material itself. So if you'd want the first thing, maybe call it ```material.replaceAttributes(otherMaterial)``` or separate methods like ```material.replaceColor(otherMaterial)```. If you'd want the second thing, maybe call it ```material.replace(otherMaterial)```. It could only suggest replacement by something. But the word "copy" could imply two things: copy from or copy to. (I would also suggest the same thing on ```GameObject```, to have ```replaceMaterial(material, slot)``` instead of ```material(material, slot)```).

Since you were asking if there is another way of doing this I answered with that example:
```java
g.modelInstance.materials.first() = other.modelInstance.materials.first();
```

SolarLune:
Note that the copy functions do not replace the materials. It copies our relevant attributes (color, tint, shadelessness, and blend mode) to the target material, materials, or materials in GameObjects. 

Its purpose is to stand in for the "apply to children / apply over all materials" aspects of the current color() and tint() implementations. This way, you could easily color a material as you wish, and then apply that coloring over a GameObject's other materials or all of its children by just doing:

```java
Material mat = materials().get(0);
// mat.color() or mat.tint() or whatever
mat.copyToMaterials(materials());  // And then copy that coloring across all other materials
mat.copyToGameObjects(children); // And then copy it for its children, too
```

The purpose isn't to replace the material on the target object or anything like that.

With that in mind, perhaps ```copyColoring`` or something. The point, though, is to not have a function for each of our custom features, because that seems like a lot of unnecessary code (i.e. `copyColor`, `copyShadeless`, `copyTint`, and so on).

rafcolson:
Again, ```copyColoring()``` or just ```copy()``` is confusing in my opinion. It should at least be ```copyTo()```. But then also, why have this exception of applying changes to other materials/objects from a method on the material? We also don't have methods like ```GameObject.copyModelTo(objects)``` just for those rare occasions where we would want a bunch of objects have the same model. And also, replacing all attributes just for the sake of having the same material sounds a bit silly, why not replace the material then?

SolarLune:
I'm not bent on the name, it was just a suggestion. `copyTo` is fine, but I already tried having two functions with the same name and different arguments (of both type ArrayList, but different element types); Java wasn't having it. So, we need to differentiate the functions by name.

> But then also, why have this exception of applying changes to other materials/objects from a method on the material? We also don't have methods like GameObject.copyModelTo(objects) just for those rare occasions where we would want a bunch of objects have the same model. And also, replacing all attributes just for the sake of having the same material sounds a bit silly, why not replace the material then?

I feel like you're missing my point, but I'm not sure. The goal of the copy functions in their current form is to copy relevant aspects of one material to another material, not to replace or duplicate a material. Your example seems like it's saying that we're trying to copy materials to other GameObjects or something; we're not. We don't want materials to always be the same with the copy functions. I think we basically just want to be able to have what we already have - the ability to set the color of several objects / materials at once, while also having the ability to specify individual materials to alter.

As for the use case for this class, it's fairly simple; you have an object with multiple children. You want to color one object, say, the "core" object, and have the other objects (the children) follow suit.

rafcolson:
I'm sorry about the trouble I'm giving you with this. It's only because I'm not comfortable with the naming part and also the couple of functions I find to be very uncommon.

My first point is that functions ```copyToMaterials()``` and ```copyToGameObjects()``` are too uncommon for those to be in the ```Material``` class, in my opinion. Do you agree that these functions would only be used in some rare occasions? I'm sure there are work cases, otherwise you wouldn't be bringing it up, but I honestly couldn't imagine one right now.

My second point is that such functions are in a way an exception to the rule. If granted, these would be the first that apply changes to their arguments, instead of their own class instances. They would coexist next to methods doing the very same thing, but to the class instance itself, for example ```copyAttributesTo()``` next to ```replaceAttributes()```. If there's a function which applies changes to more than one instance, the same function surely should also exist for making changes to just one. Consequently we'd have all these methods: ```copyColorTo()``` next to ```replaceColor()```, ```copyTintTo()``` next to ```tint(color)```, ```copyBlendModeTo()``` next to ```blendMode(array)```, ```copyShadelessTo()``` next to ```shadeless(shadelessness)```.

Then, why not have such kind of functions in other classes then? I already gave an example, but there are plenty to give. For example: ```g.copyVelocityTo(objects)``` next to ```g.velocity(vel)```, ```g.copyRotationTo(objects)``` next to ```g.rotation(rot)```, ```g.copyMaterialsTo(objects)``` next to ```g.material(mat)```, ```l.copyEnergyTo(lights)``` next to ```l.engergy(e)``` and so on... 

I would only be in favor of this if there would be many use cases, but there aren't. So I hope you understand why I'm trying to convince you to keep our classes clean.

My last point is that "replace" would be the right description for whenever you're replacing things. If you're not replacing the whole material, but only its attributes, "replaceAttributes" would be the most obvious description. Using just "copyFrom" (or "copyTo" for that matter) would raise the question 'what am I copying?'. And assuming we won't have functions which copy 'to', for the aforementioned reasons, I also wouldn't use descriptions like "copyAttributesFrom". It may suggest there's also a method "copyAttributesTo", while "replaceAttributes" is as clear as can be.

I do think naming is important in the long run. If we think about it now, we won't be sorry afterwards.

SolarLune:
Nah, I agree with your points overall.

1) As I said, I don't much care about my proposed name for the functions; anything is fine.

2) It'd be nice to reverse the way the `copy()` function works, but that's not really possible, because the general idea is to copy the colors from one material to all materials in a provided collection. We could, I suppose, make an ArrayListMaterial, which has a copyFrom() function on it (or something like that).

> I'm not opposed to the idea of copying everything, so two materials are virtually the same. But I'm questioning if replacing the material altogether wouldn't fit the same need? If you want to restore the material, in either case you would have to store the old material. It doesn't matter if you're replacing all attributes or if you're replacing the material itself then.

3)  No, because we still want the materials to be separate (i.e. because they have different textures). We just want to copy relevant attributes, while still leaving them separate and unique. The use case that I provided earlier would be for this - separate objects with individual materials (and probably individual textures) that you want to apply a color to.

> Edit: In retrospect I really want to emphasize that we should at least keep the GameObject.color() methods and even expand to GameObject.color(color, index)

4) I think I agree, but I think it's also true that we can't really keep this up indefinitely; at some point, we'll need to split Material-related functions off into a Material class of its own, right?

SolarLune:
I guess a mixture would be best? Leave the functions the way they are, but allow people to get individual materials as they wish? That way, if they wanted to color a singular material, they can, but the color functions can still functionally work as they have so far.

rafcolson:
>2) It'd be nice to reverse the way the ```copy()``` function works, but that's not really possible, because the general idea is to copy the colors from one material to all materials in a provided collection. We could, I suppose, make an ArrayListMaterial, which has a copyFrom() function on it (or something like that).

I'm not sure if this wouldn't make things unnecessarily complicated. If the general idea is to copy colors from one material to all materials on a GameObject, I guess ```g.color(mat.color())``` would be doing just the same, wouldn't it?

>3) No, because we still want the materials to be separate (i.e. because they have different textures). We just want to copy relevant attributes, while still leaving them separate and unique. The use case that I provided earlier would be for this - separate objects with individual materials (and probably individual textures) that you want to apply a color to.

You're right. I didn't think of that. Then "copyAttributesTo" wouldn't give a correct description, since it also implies that textures are copied. It sure is a hassle. Would it be common to copy color, tint, blend mode and shadeless all at once? If not, I would go for separate functions, such as ```copyColorToMaterials```. On the other hand, if we are keeping the existing ```GameObject.color()``` functions, I believe such functions would be unnecessary, as they're doing the same thing.

>4) I think I agree, but I think it's also true that we can't really keep this up indefinitely; at some point, we'll need to split Material-related functions off into a Material class of its own, right?

I don't think we should ever lose the ```GameObject.color()``` and ```GameObject.tint()``` functions. ```GameObject.blendMode()``` and ```GameObject.shadeless()``` are more specific and certainly not as common. Although they could be useful I wouldn't mind losing those. I suggest to keep all of those instead of having the ```copy...ToMaterials()``` and ```copy...ToGameObjects()``` functions on ```Material```. Then we have best of both ways. Do we want to alter an attribute on all materials of the object, we use the ```GameObject``` methods, for example: ```g.color(color)```. Do we want to alter one specific material, we use the ```Material``` methods, for example: ```g.materials.get("someMaterial").color(color)```.

@GoranM What's your opinion on all of this?

GoranM:
> No, because we still want the materials to be separate (i.e. because they have different textures). We just want to copy relevant attributes, while still leaving them separate and unique. The use case that I provided earlier would be for this - separate objects with individual materials (and probably individual textures) that you want to apply a color to.

If you create a brand new material instance, isn't that a separate material?

Fundamentally, copying a material implies the creation of a new material instance that carries the same properties (again, this is what the libgdx base class does). If you just want to copy specific properties from one material to the other, then you should just use the relevant getters/setters (ie: `a.color(b.color())`).

----

Let's step back and imagine the kind of API that makes sense, given what we want to do:

If I want to set the color for all materials on gobj:

```
gobj.materials.color(c);
```

If I want to change the color of a specific material:

```
gobj.materials.get(idx).color(c);
```

That seems fairly simple.

rafcolson:
>```gobj.materials.color(c);```

I think that's it. So an ```ArrayListMaterial``` after all. I'm fine with that instead of keeping the methods on ```GameObject```. Then also add functions ```Material.unique()```, ```ArrayListMaterial.unique()``` would be in order, so you could do ```g.materials.unique()``` or ```g.materials.first().unique()``` in case you don't want to affect objects with the same material(s).

SolarLune:
OK, I'm almost done implementing you guys' suggestions. I started from scratch, and it all has turned out pretty well, I think. I just got rid of the GameObject.material functions altogether in favor of just having the set() function on the ArrayListMaterial serve the purpose of replacing materials. It all works rather well, and seems pretty slick. A couple of questions:

1) Should each modelInstance be given a copy of the materials, or a shared one? 

2) What do you guys think of an ability on Colors to automatically apply themselves on Materials, to allow for one-liners? Currently it's awkward:

```java
anotherObject.color(anotherObject.color().addAlpha(0.1f));
```
With the proposed PR, it'll be worse:
```java
anotherObject.materials.get(0).color(anotherObject.materials.get(0).color().addAlpha(0.1f));
```

But if we had the Material's color() and tint() color returning functions pass the reference to the Material on to the Color, then we could add an `apply()` function that would write it to the Material for us. This function could be called from our color setting functions automatically:

```java
anotherObject.materials.get(0).color().addAlpha(0.1f);
```

As a side-note, there wasn't really too much of a need for Material.unique(); it's just creating a new Material with the "to be unique'd" material provided as an argument. Then you set it. 

There is an ArrayListMaterial.unique(), though.

rafcolson:
>1) Should each modelInstance be given a copy of the materials, or a shared one?

Better a shared one, since it's possible. It makes sense to make those only unique when necessary. I don't know about memory. But a scene with a hundred objects having the same material would require a hundred materials then... Altering the material through the ModelInstance sure was handy...

>With the proposed PR, it'll be worse:
```java
anotherObject.materials.get(0).color(anotherObject.materials.get(0).color().addAlpha(0.1f));
```

```java
Material mat = g.materials.get(0);
mat.color(mat.color().addAlpha(0.1f));
```
Would this also work? If so, then I'm fine with that. But I agree it's awkward.

SolarLune:
> Altering the material through the ModelInstance sure was handy...

The way it used to be required copies of materials, so making duplicates for each material wouldn't really be any different from how it is now, I think.

> Would this also work? If so, then I'm fine with that. But I agree it's awkward.

Yeah, that would work.

GoranM:
> 1) Should each modelInstance be given a copy of the materials, or a shared one?

As raf said, shared by default would make sense.

> Material.unique()

Unlike `ArrayListMaterial.unique()`, that doesn't really make a lot of sense. I mean, I don't know how you plan to implement that, but in order to do what you intend, you have to make a copy of the material in question, and replace the existing instance in the list with a newly created copy instance.

It's best not to hide that behind some convoluted system. So, if someone want's to make a specific material "unique", they should just do:

```java
g.materials.set(idx, g.materials.get(idx).copy());
```

SolarLune:
@GoranM - I agree about the first point. What about the second point I brought up in my second-to-last post, about adding functionality to allow for automatic Color setting?

SolarLune:
OK, the PR's been updated. Lemme know what you guys think.

GoranM:
> What about the second point I brought up in my second-to-last post, about adding functionality to allow for automatic Color setting?

Having Color hold a reference to a Material is really confusing, because, conceptually, they're not related. That's exactly the kind of incidental complexity that we should try to avoid.

Raf's code snippet shows that's it's no big deal to get the material first, so I think that's fine.

SolarLune:
Fair enough. Still would like one-liners somehow, haha.

OK, anyway, the PR should be good to go.

Issue #559, opened 2016-02-29
Profiler Clean-up


SolarLune:
Hey, there. I think we could do a bit to kinda clean up the profiler's headings - there's a few things that we don't really need to profile, like updating the camera or input logic. There's just not enough time spent on it to profile it, I think. "Visuals" and "Render" or "Worldstep" and "Collisions" could be compressed, as well, into a single profiler listing - simple stuff like that. "Children", for example, could be compressed with "Render" or "Logic".

We also need might need to move some code around to be properly profiled, like moving the BDX components list after the call to stop the profiler for input checking (which isn't really necessary, as it doesn't profile all of it, just for presses).

Finally, we need to document what each category in the profiler does.

rafcolson:
Sounds good. I would personally be interested to see these categories:

- **logic** (no change)
- **scene** (input, visuals, camera, children)
- **physics** (worldstep, collisions)
- **graphics** (graphics, render)
- **outside** (no change)

SolarLune:
PR Submitted.

rafcolson:
This should fix for #576.

SolarLune:
So I should rebase my changes over that commit and remove the extra start() calls?

rafcolson:
If you want to you can just copy paste, that would be easiest. But you don't have to. I can submit a PR after this PR (like it is now) is submitted.

SolarLune:
OK, you can submit the PR, then; I'll submit the PR was the clean-up as it is.

SolarLune:
This was merged, so I'll close this now.

PR #576, opened 2016-03-07
(#559) Profiler clean-up.

SolarLune:


SolarLune:
Oops, missed the # to tie the issue and PR together.

(#559) Profiler clean-up.

rafcolson:
Thanks. But I think you could leave out a couple of "start" 's. It should be enough to add ```Bdx.profiler.stop(subSystem);```. The "start" will be the last stop time as being documented in the Profiler's API.

SolarLune:
Y'know, I tried that, but I was getting some weird results when I didn't; for example, Outside was getting negative values, and the percentages weren't adding up to 100%.

rafcolson:
You can do this only with subsystems that aren't added yet. I suppose this should work in any occasion so I'll try to fix that later.

SolarLune:
Oh, that might've explained it. Thanks for the info!

GoranM:
Branch has conflicts. Please rebase from master.

SolarLune:
OK, that should do it.

PR #578, opened 2016-03-09
Fixes two issues with the profiler.

SolarLune:
1) Using the materials ArrayList now, rather than the old color() function that was removed.
2) Adds a material for the __PBackground object in the profiler blend file (with the same material name). The material has transparency checked, so the background now shows up correctly behind the bar graphics and text.


rafcolson:
>2) Adds a material for the __PBackground object in the profiler blend file

Hmm, this is not desirable. Isn't there another way to fix this issue? Because as far as I can remember. Any material added to the Profiler.blend will also show up in the list of materials in the user's blend file. For this reason I used the default material. I suppose everything should show up correctly also when no material is added. Maybe you can could treat objects without a material as objects that have material transparency checked?

SolarLune:
> Because as far as I can remember, any material added to the Profiler.blend will also show up in the list of materials in the user's blend file. For this reason I used the default material.

I checked, and the material isn't visible in the generated game.blend file. Maybe it was something else? Maybe from when the profiler was in the generated game.blend file (I think it was at some point) and not just a BDX file in the scenes folder?

> I suppose everything should show up correctly also when no material is added. Maybe you can could treat objects without a material as objects that have material transparency checked?

We could do that as well, I suppose. It's up to you guys.

rafcolson:
>I checked, and the material isn't visible in the generated game.blend file.

That's fortunate. Nothing to worry about then.

rafcolson:
@SolarLune I just checked if the material is visible in the generated game.blend and it is. Are you sure it wasn't on your end?

Anyway I'm gonna try to fix it.

SolarLune:
Haha, wow. I just fell for something that we all know about and that I personally documented well.

The profiler background's material's the default material, so changing its color changed the bar's color, too. Calling `background.materials.unique()` actually fixes the issue.

This PR can be undone.

rafcolson:
No big deal. And thanks for that info.

Issue #581, opened 2016-03-09
Issue: Can't Extend Custom Material Shaders From Lighting Shaders


SolarLune:
Hey, there. 

So basically, the issue is that we can't use custom shaders that extend from the lighting shaders. The problem comes in because when creating the default lighting shaders, we allow LibGDX to create its own ShaderProgram, which adds on a prefix to the beginning depending on the attributes of the renderable it's created for. However, when we pass in our custom ShaderProgram, this prefix isn't added on. Since we can't edit the shaderprogram's fragment and vertex source later, we can't add in the prefix later, and since the lighting shaders reference things in the prefix, it crashes. So, we can't base custom shaders off of the existing lighting shaders. 

Basically, BDX will crash if you try to use a custom material shader with the built in default.frag and default.vert shaders.

A solution could be to create our prefix for ShaderPrograms using a fixed prefix no matter what (since we know what our renderables require).

Another solution could be for our ShaderProgram to contain the actual LibGDX shader program data, which would be recreated as necessary when LibGDX feels like it. Our ShaderProgram will then overload the setUniform functions, so the users would generally create a Program and then call setUniform on it. The function calls will only go through if the program data has been initialized.

I'll try the first solution, well, first.

SolarLune:
Tried the first solution; I thought it best that I didn't do that (since it'd be fixed across all renderables).

I ended up going with an alternative solution.

Basically, I'm adding a new kind of ShaderProgram, the MaterialShader, and renaming the old ShaderProgram to ScreenShader. Why?

- Because 3D material shader programs don't need any of the fields on the ScreenShader, like renderScale or overlay.
- Because 3D material shader programs should NOT be compiled immediately like the ScreenShader can be, but rather should only be compiled when the renderer goes to append the prefix to the beginning of the shader source and actually create the shader.
- Since the 3D material shader program can be recompiled through the above point, the user shouldn't be able to store a reference to it (since it would be broken when the shader program is recompiled). To facilitate this, the MaterialShader is a shell around which the program data exists (in the `programData` field). For the difference, see:

```java

screenShader = ScreenShader.load("bdx/shaders/2d/default.vert", "bdx/shaders/2d/bloom.frag");
screenShader.begin();
screenShader.setUniformf("strength", 1.0f);
screenShader.end();

materialShader = MaterialShader.load("bdx/shaders/3d/default.vert", "bdx/shaders/3d/static.frag");

// And then later

if (materialShader.compiled()) {
[...]

PR #582, opened 2016-03-10
(#581) Custom Material Shaders fix.

SolarLune:
Adding MaterialShader and renaming ShaderProgram to ScreenShader.
Calling ArrayListMaterials.unique() will rename cloned materials to try to keep them unique.

### DIFF
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -9,11 +9,7 @@ import com.badlogic.gdx.graphics.*;
 import com.badlogic.gdx.graphics.g2d.*;
 import com.badlogic.gdx.graphics.g3d.*;
 import com.badlogic.gdx.graphics.g3d.shaders.DefaultShader;
-import com.badlogic.gdx.graphics.g3d.utils.DepthShaderProvider;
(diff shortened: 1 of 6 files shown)


GoranM:
> Calling ArrayListMaterials.unique() will rename cloned materials to try to keep them unique.

Aren't they already unique, even if they have the same name? I'm not really sure if this is necessary.

SolarLune:
They are unique, yeah, it's just that the material shader system is dependent on the material's name, so if two materials share a name that's in the matShaders HashMap, they'll both get the shader applied to them. By automatically incrementing a value in the name, this is avoided to some extent. 

Do you think it would be better not to have it?

GoranM:
Shouldn't the clones of a given material use the same shader? If that's the case, then they should have the same name.

If there's a circumstance in which a cloned material needs to use a different shader, I suppose that we could have a shader reference on the material itself - That seems like it would make more sense, given that we have our own material class.

SolarLune:
Yeah, I guess that does make sense. I'll leave the unique() function alone and approach the shader to material setup another time, then.

PR updated.

SolarLune:
PR fixed, as it was touching ArrayListMaterials.unique(), which is supposed to be taken out, and so would've caused a conflict. I think it should be good to go at this point.

PR #586, opened 2016-03-13
Update README.md

SolarLune:
Adding a picture to the readme.


GoranM:
Considering all the media we already have, I'm not really sure we need a screenshot.

Even if we did, I think the wiki would probably be a better place for it.

SolarLune:
I'm inclined to agree, but I'm not sure where else we actually have any screenshots that show what the engine looks like in Blender. 

Even though it's said that BDX is "Blender-integrated", I think people would rather just take a look at a screenshot and say, "Oh, the entire thing's in Blender."

rafcolson:
>Even though it's said that BDX is "Blender-integrated", I think people would rather just take a look at a screenshot and say, "Oh, the entire thing's in Blender."

Maybe there could be a section "Screenshots" in the Wiki. People could have a look there to get a general impression.

Issue #606, opened 2016-03-25
Suggestion: hitObjects


SolarLune:
Hey, there. A `hitObjects` list of objects that the object just came into contact with would be nice to have. The alternative is deriving it yourself using the touchingObjects and touchingObjectsLast lists.

GoranM:
Ok.

SolarLune:
Hey, so I made this a function, as maintaining another list is unnecessary. Cool?

SolarLune:
This was merged, so I'll close it now. Thanks!

PR #611, opened 2016-03-26
(#606) hitObjects() function to return the objects hit.

SolarLune:


GoranM:
Using the method internally will probably be slower, because it's an additional allocation, but since game logic is usually not the bottleneck, I guess it's ok.

If you're going to use `hitObject()` internally, you should leverage the fact that it returns an `ArrayListGameObject`, which has a `get(String name)` method. So, in GameObject.hit(String name) you can just do `hitObjects().get(name)` instead of looping.

SolarLune:
OK, PR updated. Now the hit functions all use the ArrayListGameobject get and getBy functions.

Issue #603, opened 2016-03-24
Suggestion: Camera.zoom() Relative to View Size


SolarLune:
Hey, there. 

I think rather than returning a somewhat obtuse number, the Camera.zoom() functions could set and return the value relative to the width of the camera. Basically, rather than zoom() returning, say, 0.0374999998, it could return 30, for 30 Blender Units across (like how it works in Blender).

GoranM:
Sounds good.

SolarLune:
OK. PR submitted.

SolarLune:
PR merged, so I'll close this now. Thanks!

PR #612, opened 2016-03-26
(#603) Camera.zoom() now is in Blender Units.

SolarLune:


src/com/nilunder/bdx/Scene.java
```diff
@@ -333,11 +333,15 @@ else if (settings.getString("type").equals("SPOT"))
...
+					c.width(ds.x);
+					c.height(ds.y);
 					c.projection(new Matrix4f(projection));
 					c.fov(c.fov());
 				}else{
 					c.initData(Camera.Type.ORTHOGRAPHIC);
-					c.zoom(2 / (projection[0] * ds.x));
+					c.width(ds.x);
+					c.height(ds.y);
+					c.zoom((2.0f / (projection[0] * ds.x)) * ds.x);  // Extra * ds.x because zoom() now multiplies / divides by camera width
```
rafcolson: `c.zoom(2 / projection[0])` should do it.

SolarLune: You mean "2" instead of "2.0f"?

SolarLune: Ah, that does work. PR updated!

Issue #622, opened 2016-04-02
GameObject.props shared between instances


SolarLune:
Hey. 

So it seems like changes to a GameObject's props in-game are shared between instances of GameObjects. Each GameObject should have its own HashMap, of course.

SolarLune:
I believe that should do it. PR submitted.

rafcolson:
I'm all for it of course but as I recall @GoranM didn't think it would be
beneficial. Maybe it's good to give an example of the benefits.

SolarLune:
That's weird; if it's left the way it is, then you can't use props to identify instances of a common class. Like, I have the player object and an enemy both shooting bullets, but I use props to distinguish which they are (because they're fundamentally the same; it's just what they can damage that changes).

rafcolson:
Exactly.

rafcolson:
Maybe lines 534-535 could also be:
```java
g.props = new HashMap<String, JsonValue>(gobj.props);
```

SolarLune:
I think the goal with that section is to write the instance's props over the GameObject's established ones. Is there anything wrong with the proposed PR, or what behavior would your addition change?

rafcolson:
Oh right. Forgot about that. Sorry. Of course we don't want to lose the group instance props!

GoranM:
The api docs state:

> Unless otherwise labeled, all fields should be considered read-only.

So, you're not supposed to change props at runtime; The system assumes that you're just going to use them to set your own local data, which you can then modify however you wish.

SolarLune:
Is there a reason why the system can't be adjusted to have props be unique to each instance?

rafcolson:
>The system assumes that you're just going to use them to set your own local data, which you can then modify however you wish.

I see you point but it's not really a matter of modifying data. I think customizable props can be useful in two cases:
- With adding instances of a GameObject which has no use for a class except for one or more fields to differentiate between those instances. Example: Coins could look the same but have different values. No class needed for every coin as a component could do the animation. No class needed for any other reason too, as all the coin does is do the animation and return its value (for the bank).

- With adding instances of a GameObject which uses a class, but can't be identified through a field so easily. For example: Hero wanders and encounters one type of enemy after the other. In this case it could be argued that a 'common' field as an identifier would do. But this would also require every type of enemy to have a class, just for the sake of providing a field. But not every enemy would require a class. A component for the animation, and when Hero hits this type of enemy, he gets hurt, but can't destroy it. For example a cactus.

[...]

PR #623, opened 2016-04-02
(#622) Each object instance should have its own HashMap.

SolarLune:


GoranM:
This would make line 534 redundant, so I think you can safely remove it.

SolarLune:
PR updated.

Issue #572, opened 2016-03-06
Set Scaling to -1


SolarLune:
Hey, there. I'm trying to flip a mesh to face the opposite direction. As far as I know, apart from making duplicate meshes that are mirrored, the only possible way to do this is to reverse the scaling on the X axis. However, BDX doesn't support negative values for scaling, right? Why / is it possible to fix / is there another, alternative solution for this?

GoranM:
If I remember correctly, the getLocalScaling method on the bullet collision shape will always provide a vector with positive elements, regardless of what is provided via setLocalScaling.

If you store the scale on the game object itself, that should allow for negative scale (for the visual part - the physics should just ignore it, as it does now).

SolarLune:
Hmm, I tried replacing that line in scale() with a stored scale vector, but it doesn't work for collision shapes that aren't Triangle Mesh or Convex Hull. It looks like the other shapes might call ConvexInternalShape.setLocalScaling, which sets the local scaling absolutely...

Would it be possible to, say, alter the modelInstance's transform for these cases where the object has an internal collision shape (though I don't know how to do it)?

GoranM:
It should be more than just a one line change; Search for all uses of getLocalScaling, and make sure those vectors are replaced with the new scale stored on GameObject.

SolarLune:
OK, I think I've got it now. Testing seems fine on my end; lemme know if there's anything wrong.

GoranM:
Once the PR is up, the discussion should continue there.

SolarLune:
This was merged, so I'll close this now. Thanks!

PR #626, opened 2016-04-04
(#572) Scaling can be negative now.

SolarLune:


GoranM:
> Scaling is absolute (so if you have an object that's 0.5 scale and check the child's scale, it'll be 0.5 as well (unless you alter it)). Not sure if this is good or not (maybe it should be locally influenced?).

I don't know. I mean, absolute seems sensible to me. It also makes implementation a bit simpler, since object scale can be set directly as CollisionShape scale.

Do you have an argument for one over the other?

SolarLune:
Not really, but we have a localScale field, so I thought that maybe it would be better to have them be separate and both available or something. So the localScale field would hold the scale for just the GameObject, and the scale() function would return the scale of the object after "counting" the parents' scaling. Or something like that.

But I don't really mind either way; just ensuring whichever approach we go with is documented would be fine.

Issue #610, opened 2016-03-26
Suggestion: Model Class


SolarLune:
Hey, there. 

So Goran suggested a Model class, which would be used to modify a GameObject's modelInstance easily and quickly. I'd like it to handle vertex manipulation and UV sliding easily and quickly. Not sure if this would require a Vertex class as well...

Any ideas for what it should do / how it should work?

rafcolson:
Naturally `GameObject.model` would be an instance of a LibGdx Model extension. The `Model` class would contain methods for altering meshes, mesh parts, vertices, ... But what kind of alterations would we talk about then? Apart from uv scrolling, I can only think of armature deformation and texture deformation. Methods on GameObject for replacing the model could become ``Model.replace(model)`. Also the method to make the model unique becomes ```Model.unique()```.

It would make more sense to me if materials are accessed through the Model instead of the GameObject. Maybe we could extend `Array<...g3d.Material>` to `ArrayMaterials`? I'm not sure.

GoranM:
> So Goran suggested a Model class, which would be used to modify a GameObject's modelInstance

I think the vertex buffer is managed by Model, not ModelInstance, so we would need a class that extends Model, not ModelInstance.

@rafcolson 

Your suggestions seem reasonable, but I don't think they should be a part of the initial patch. For now, it would be best to just have a basic extension of Model, with simple methods to manipulate vertex data.

SolarLune:
If we just extend Model, then the modelInstance field on GameObject won't work, right? Don't we need to extend at least ModelInstance for it to be created correctly / for the added mesh altering functions to be accessible?

GoranM:
I don't really understand what you mean.

The data is on the Model. You would get a reference to a model, and the methods would be accessible from there.

SolarLune:
So we wouldn't have a reference to a ModelInstance anymore, it'd just be a reference to our Model. OK, I think I see. 

So what's the list of things that our Model class would need?

Like, would we have `moveVertex()` and `uvVertex()` functions, or something like that? Or just a list of Vertex instances representing vertices on the model that the user can move or adjust...?

To sum up, I made this initially for UV sliding across the entire mesh, so something that does that, in addition to UV adjustment for each individual vertex, would probably be cool.

GoranM:
> So we wouldn't have a reference to a ModelInstance anymore, it'd just be a reference to our Model. OK, I think I see.

[...]

PR #630, opened 2016-04-14
(#610) Mesh Class

SolarLune:
A Mesh class for in-game mesh alteration.

For #610.


GoranM:
I don't understand what you're "not so sure about" when it comes to using the transform functions. You're just passing in matrix data provided by the user; The correctness of the matrix is their responsibility.

SolarLune:
OK, I wasn't sure if I was creating the new Matrix3 and Matrix4 correctly, but it looks like it's right.

Issue #636, opened 2016-04-22
Bug: Execution Priority Can Only Apply to One Object


SolarLune:
Hey, there. 

The way it's coded now, execution priority can only apply to one GameObject that it's checked on because of an oversight in code (breaking out of the first loop in addInstances() and swapping objects around rather than placing them at the front of the list). PR submitted to fix this.

SolarLune:
PR submitted.

GoranM:
It wasn't an oversight: The logic clearly assumes that there will be one object that needs to execute before all others (the original intention was to provide an entry point for some basic scene initialization, nothing more).

Why do you need more than one object to be in the front?

SolarLune:
In my use case, I have a "System Manager" object run first to set up input bindings. I also have the Player object run first because it creates components that are checked for in the GUI scene (health and energy). I could check to ensure the player's components exist first, but I don't think I should have to, as the player obviously exists from the start. Just making it run first fixes the issue.

In my opinion, it doesn't make much sense to limit the execution priority to one single object, and it isn't something that's obvious in the Blender GUI. It's not like it's a field in the Render panel where you can set the name of the single object to run first or something - it's something that's available on any and all objects.

Regardless of the original intent, it's currently just a flag that sets execution priority for a GameObject. There's not really any sense in locking it down to a specific purpose and use-case.

GoranM:
Twiddling with execution order seems no less awkward than simply waiting until the object is initialized, but I guess it's subjective.

SolarLune:
Well, objectively, the difference is adding if statements to code for as many objects as the developer would need to wait for the first object, vs. clicking a button.

It's up to you, but it seems unnecessary to limit the system based on what it was originally built for.

SolarLune:
Ah, well, apparently this wouldn't have solved this use case anyway, and I'm just working around it at this point, so never mind. Maybe I'll look into a satisfactory solution to this when I happen upon it in the future. Thanks.

PR #637, opened 2016-04-22
(#636) Execution Priority Fix.

SolarLune:
Add GameObjects to front of list if priority bit is set.


GoranM:
From what you said in the issue, I assume that you need one specific object to initialize before another specific object (ie: for Player to initialize before the object that adds the gui scene in its init method).

If both Player and the object that adds the gui are marked for initial execution, I don't think this is guaranteed to work, because the order depends on that of the blender list, and that can change.

SolarLune:
Oh, wait, so adding a scene initializes all of the objects in that scene immediately? If so, then yeah, this wouldn't have fixed the problem for the use case I mentioned. It could still be useful, though. 

Well, I guess I'll close this in favor of looking to come up with a better solution for execution priority as I happen upon use cases in the future. Thanks.

Issue #643, opened 2016-04-25
Suggestion: Update from LibGDX 1.7.2 to latest stable or 1.7.3-SNAPSHOT


Lucrecious:
I noticed that controller connection/disconnection(on ControllerListener) was not implemented properly on the version of LibGDX that Bdx is using (1.7.2), so hotplugging isn't possible (even with polling from Controllers.getControllers()).

If there is an alternate way to achieve hotplugging, please show me. Otherwise, here's a thread with this information I mentioned:
https://www.bountysource.com/issues/3405158-detect-controller-hotplugging

Here's the Controller wiki:
https://github.com/libgdx/libgdx/wiki/Controllers

Thanks.

SolarLune:
Ah, I remember reading about this. I don't think there is a better way to achieve hotplugging (at the moment, at least).

Updating the version of LibGDX might be a good idea to support this.

Lucrecious:
Do you know how we can update it? I can't figure out a way to do it without getting a bunch of errors.

SolarLune:
I meant updating the version in master, like moving BDX to the latest version of LibGDX.

But if you want to update just your own version, it should be possible. Basically, you need to do [what Goran did when we updated BDX to 1.7.2](https://github.com/GoranM/bdx/commit/17953c52cffeb53881018d99218b9ee97a56b246). Looks like he mainly adjusted the build.gradle and build.xml files, and maybe replaced gdx-setup with the latest version from LibGDX. [After that, I believe you might have to create a brand new project in Blender and run that to make LibGDX download the latest dependencies. From there, you might be able to proceed as usual?](https://github.com/GoranM/bdx/issues/235#issuecomment-167465877)

GoranM:
> Looks like he mainly adjusted the build.gradle and build.xml files, and maybe replaced gdx-setup with the latest version from LibGDX.

Basically, but you can't just change the version variables in the existing build.gradle file, because that might miss whatever changes LibGDX people made in the new build.gradle file. So, in reality, you have to replace the old build.gradle file with a new build.gradle file that's generated by the new gdx-setup program. Then you add in the things that BDX needs on top of that (fileTree directives to include the relevant bdx jars, and the gdx-controllers plugin), and that gives you the proper file that can be committed to master.

Do you guys want me to do this, or does someone else want to give it a try?

SolarLune:
I'm good for now, haha; feel free to do it.

Lucrecious:
[...]

PR #644, opened 2016-04-27
(#643) - Fix the Display window configuration functions to use the...

SolarLune:
... new API.

### DIFF
--- a/src/com/nilunder/bdx/Bdx.java
+++ b/src/com/nilunder/bdx/Bdx.java
@@ -23,7 +23,7 @@ public class Bdx{
 	public static class Display{
 
 		public void size(int width, int height){
-			Gdx.graphics.setDisplayMode(width, height, fullscreen());
+			Gdx.graphics.setWindowedMode(width, height);
 			refreshGamepadsTimer.restart();
 			refreshGamepadsTimer.resume();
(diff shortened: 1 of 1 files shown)


src/com/nilunder/bdx/Bdx.java
```diff
@@ -37,8 +37,11 @@ public Vector2f center(){
 			return new Vector2f(Gdx.graphics.getWidth() / 2, Gdx.graphics.getHeight() / 2);
 		}
 		public void fullscreen(boolean full){
-			Graphics.DisplayMode dm = Gdx.graphics.getDesktopDisplayMode();
-			Gdx.graphics.setDisplayMode(dm.width, dm.height, full);
+			Graphics.DisplayMode dm = Gdx.graphics.getDisplayMode();
+			if (full)
+				Gdx.graphics.setFullscreenMode(dm);
+			else // Slight shrinking so that we don't try to set the size of the window to be the entire desktop
```
rafcolson: I think this we shouldn't temper with this. Isn't up to the developer to decide what to do in case the window is larger than the screen?

SolarLune: OK.

Issue #654, opened 2016-05-12
Memory Leak: Framebuffers Being Created Too Often


SolarLune:
Hey, there. 

So basically, there's a massive memory leak if you use two (or more) ScreenShaders with differing downsample sizes. When rendering the game, BDX will create a new Framebuffer every time the current ScreenShader's downsample size is different from the temporary render buffer's render percentage. This causes astronomical bloat and hogging of 100% CPU / RAM over a few seconds, resulting in a system freeze. The fix is simply to create RenderBuffers only when necessary.

PR submitted.

GoranM:
If you find a bug, and manage to resolve it yourself, you can just make the PR - there's no need for an issue in that case.

SolarLune:
Oh, OK. Thanks for the info! I thought it would look kinda weird to just propose somewhat major code changes out of the blue.

SolarLune:
The PRs were merged, so I'll close this now. Thanks!

PR #655, opened 2016-05-12
(#654) - Fixing memory leak caused by filter downsampling.

SolarLune:
Problem was caused by creation of new RenderBuffer for the tempBuffer whenever multiple ScreenShaders had different downsample values.

The solution is to remove the tempBuffer and create a map of screen sizes to render buffers. If a buffer for the screen size doesn't exist in the map, one is created and added to it. The tempBuffer RenderBuffer is replaced by getting the appropriate buffer from this map. When the screen size changes, the map is cleared (as the buffers may no longer have the same aspect ratio).


GoranM:
Branch has conflicts. Please rebase.

SolarLune:
PR updated. Thanks!

Issue #704, opened 2016-07-04
Bug: Material alpha not set at initialization


rafcolson:
Replace `Sacky.blend` with:
```java
import com.nilunder.bdx.*;

public class Sacky extends GameObject{
	
	private float alpha(){
		return children.get(0).materials.get(0).color().a;
	}

	public void init(){
		System.out.println(alpha());
	}
	
	public void main(){
		if (Bdx.keyboard.keyHit("space")){
			System.out.println(alpha());
		}
	}
	
}
```
You'll see that at any point the color's alpha value is not correct.

rafcolson:
PR submitted.

rafcolson:
PR was merged. Thanks!

PR #705, opened 2016-07-04
(#704) Fix: Material alpha

rafcolson:


GoranM:
> opacity = hasAlpha ? mat.get("opacity").asFloat() : 1;

Is there some unexpected/undesired behavior if you set an opacity value even if hasAlpha is false?

-

Why did you split:
 
```java
String texName = mat.get("texture").asString();` into:
```
into:
```java
String texName;
// ...
    texName = mat.get("texture").asString();
```
?

I'm pretty sure that the Java compiler is smart enough to avoid reallocating variable space on every iteration.

rafcolson:
>Is there some unexpected/undesired behavior if you set an opacity value even if hasAlpha is false?

Yes. If Transparency Alpha is smaller than 1 this will change the opacity (darken) the color with objects that doesn't use Transparency. Sometimes you might want to inspect an objects material (color and/or texture). In such cases it would be inconvenient to set Transparency to 1 each time.

>I'm pretty sure that the Java compiler is smart enough to avoid reallocating variable space on every iteration.

I'm sorry. That was a mistake. PR updated.

Issue #701, opened 2016-06-27
Discussion: Viewport -> utilize LibGDX viewports or not?


rafcolson:
[LibGDX Issue 4175: Suggestion: don't ignore screenX and screenY while updating Viewports](https://github.com/libgdx/libgdx/issues/4175)

[badlogicgames forum topic: Viewport.update() ignores screenX and screenY](http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=22932)

Continued from #469.

There are issues that come with utilizing LibGDX viewports for implementing a single viewport (per scene) in BDX. This would also apply for implementing multiple viewports.

With [ExtendViewport](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/viewport/ExtendViewport.java#L87), [FitViewport](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/viewport/ScalingViewport.java#L59) (Letterboxed) and [ScreenViewport](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/viewport/ScreenViewport.java#L40) the `update()` function sets the position of the viewport at the center or at the left bottom corner of the display screen. This happens while calling [`setScreenBounds()`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/viewport/Viewport.java#L206), which sets `screenX`, `screenY`, `screenWidth` and `screenHeight`.

But, this will not take into account the actual position of the viewport relative to the display size. This can be solved by setting the position again after calling `update()`, but `update()` is also calling `apply()` (which sets the gl viewport). So we also have to call `apply()` again, after setting the position with `screenX()` and `screenY()`. This was to be expected as it is mentioned that `screenX()` and `screenY()` are typically set by the `update()` function, which you can read [here](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/viewport/Viewport.java#L162) and [here](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/utils/viewport/Viewport.java#L171). This also forces us to store the `screenX` and `screenY` positions before calling `update()` in order to return to these values again.

With LETTERBOX the viewport won't update correctly, so we have to update `screenX`, `screenY`, `screenWidth` and `screenHeight`. With SCREEN we have to tweak with `setUnitsPerPixel()` so it updates correctly.

[...]

PR #713, opened 2016-07-11
(#701) Viewports LETTERBOX, EXTEND, SCALE and SCREEN

rafcolson:
API changes:

# API Viewport

A viewport represents a window for the scene's camera. Its size and position are relative to the display screen.
___

<a name="type-enum" href="#type-enum">#</a> **Type : enum**

The Type enum establishes the kind of viewport the Viewport object is, either `LETTERBOX`, `EXTEND`, `SCALE` or `SCREEN`. You can access it publicly anywhere you import the Viewport class.

`SCALE` will stretch the view to fill the display screen without preserving the aspect ratio. `LETTERBOX` will stretch the view while preserving the aspect ratio by adding bars to fill the display screen. And `EXTEND` will stretch the view while preserving the aspect ratio by extending the width or height to fill the display screen. `SCREEN` will only stretch when manually changing the resolution or size. 
___

<a name="scene" href="#scene">#</a> **scene** - Scene

Reference to the scene.
___

<a name="type-getter" href="#type-getter">#</a> **type**() - Type

Returns the type of this viewport.

<a name="type-setter" href="#type-setter">#</a> **type**(Type type)

Sets the type of this viewport.

<a name="orientation-getter" href="#orientation-getter">#</a> **orientation**() - Matrix3f

Returns the orientation of this viewport directly. Calls `scene.camera.orientation()`.

<a name="resolution-getter" href="#resolution-getter">#</a> **resolution**() - Vector2f

Returns the resolution for this viewport. Calls `scene.camera.resolution()`.

<a name="resolution-setter-Vector2f" href="#resolution-setter-Vector2f">#</a> **resolution**(Vector2f resolution)

<a name="resolution-setter-ints" href="#resolution-setter-ints">#</a> **resolution**(int width, int height)

Sets the resolution for this viewport. Calls `scene.camera.resolution(int width, int height)`. Changing this vector may also change the aspect ratio. With viewports of type `SCREEN`, this value can be altered to get 'pixel-perfect' results. [Link to Wiki here].

<a name="size-getter" href="#size-getter">#</a> **size**() - Vector2f

Returns the size of this viewport.

<a name="size-setter-Vector2f" href="#size-setter-Vector2f">#</a> **size**(Vector2f size)

<a name="size-setter-ints" href="#size-setter-ints">#</a> **size**(int width, int height)

Sets the size of this viewport.

<a name="sizeNormalized-getter" href="#sizeNormalized-getter">#</a> **sizeNormalized**() - Vector2f

Returns the normalized size of this viewport.

<a name="sizeNormalized-setter-Vector2f" href="#sizeNormalized-setter-Vector2f">#</a> **sizeNormalized**(Vector2f sizeNormalized)

<a name="sizeNormalized-setter-floats" href="#sizeNormalized-setter-floats">#</a> **sizeNormalized**(float width, float height)

Sets the normalized size of this viewport.

<a name="position-getter" href="#position-getter">#</a> **position**() - Vector2f

Returns the position of this viewport.

<a name="position-setter-Vector2f" href="#position-setter-Vector2f">#</a> **position**(Vector2f position)

<a name="position-setter-floats" href="#position-setter-floats">#</a> **position**(float x, float y)

Sets the position of this viewport.

<a name="positionNormalized-getter" href="#positionNormalized-getter">#</a> **positionNormalized**() - Vector2f

Returns the normalized position of this viewport.

<a name="positionNormalized-setter-Vector2f" href="#positionNormalized-setter-Vector2f">#</a> **positionNormalized**(Vector2f position)

<a name="positionNormalized-setter-floats" href="#positionNormalized-setter-floats">#</a> **positionNormalized**(float x, float y)

Sets the normalized position of this viewport.
___

# API Display

<a name="changed" href="#changed">#</a> **changed** - boolean

Returns whether the display changed since the last logic tick.
___

<a name="width-getter" href="#width-getter">#</a> **width**() - int

Returns the width of the game window.

<a name="height-getter" href="#height-getter">#</a> **height**() - int

Returns the height of the game window.

<a name="width-setter" href="#width-setter">#</a> **width**(int width)

Sets the width of the game window.

<a name="height-setter" href="#height-setter">#</a> **height**(int height)

Sets the height of the game window.
...
___

# API Scene
...
<a name="cameras" href="#cameras">#</a> **cameras** - ArrayListNamed\<Camera\>

An ArrayListNamed of all of the cameras currently in the scene.

<a name="viewport" href="#viewport">#</a> **viewport** - Viewport

Returns the scene's viewport.
...
___

# API Camera
...
<a name="resolution-getter" href="#resolution-getter">#</a> **resolution**() - Vector2f

Returns the resolution of this camera. Equivalent to the Blender resolution (in the Render tab).

<a name="resolution-setter-Vector2f" href="#resolution-setter-Vector2f">#</a> **resolution**(Vector2f resolution)

<a name="resolution-setter-ints" href="#resolution-setter-ints">#</a> **resolution**(int width, int height)

Sets the resolution of this camera. This is also the actual size of the texture rendered when rendering the camera's output to a texture. Changing this value may also change the aspect ratio.

<a name="width" href="#width">#</a> **width**() - int

Returns the width. Typically set by `Viewport.update()` when this camera is also the scene's camera.

<a name="height" href="#height">#</a> **height**() - int

Returns the height. Typically set by `Viewport.update()` when this camera is also the scene's camera.
...

<a name="screenPosition" href="#screenPosition">#</a> **screenPosition**(Vector3f worldPosition) - Vector2f

Returns the screen position of the specified world position.

<a name="screenPositionNormalized" href="#screenPositionNormalized">#</a> **screenPositionNormalized**(Vector3f worldPosition) - Vector2f

Returns the normalized screen position of the specified world position.

<a name="rayData" href="#rayData">#</a> **rayData**(Vector2f screenPosition) - Vector3f[]

Returns the source and direction of a ray at the specified screen position.

<a name="rayDataNormalized" href="#rayDataNormalized">#</a> **rayDataNormalized**(Vector2f screenPosition) - Vector3f[]

Returns the world coordinates of the specified normalized screen position.
___


GoranM:
@SolarLune 

Does this work fine for you?

src/com/nilunder/bdx/Camera.java
```diff
@@ -21,20 +24,26 @@
...
 	
+	private Vector2f resolution;
+	
 	public Type type;
 	public com.badlogic.gdx.graphics.Camera data;
 	public boolean renderingToTexture;
 	public RenderBuffer renderBuffer;
 	public ArrayListNamed<GameObject> ignoreObjects;
+	
+	{
```
SolarLune: Why'd you put these initialized variables in this block instead of directly setting them above (or in a function)?

src/com/nilunder/bdx/Camera.java
```diff
@@ -67,20 +76,47 @@ public float far(){
...
+	
+	public Vector2f size(){
+		return new Vector2f(data.viewportWidth, data.viewportHeight);
+	}
+	
+	public void size(Vector2f size){
+		size(Math.round(size.x), Math.round(size.y));
+	}
+	
+	public void size(int width, int height){
```
SolarLune: We should max() the width and height setters here like above to avoid the 0-size crash, or perhaps better yet, just call the width() and height() setters in the size() setter.

src/com/nilunder/bdx/Bdx.java
```diff
@@ -305,23 +329,23 @@ public static void main(){
 					filter.setUniformf("time", Bdx.time);
 					filter.setUniformi("lastFrame", 1);
 					filter.setUniformi("depthTexture", 2);
-					filter.setUniformf("screenWidth", Bdx.display.size().x);
```
SolarLune: These need to be `vp.size().x` and `vp.size().y`.

src/com/nilunder/bdx/Bdx.java
```diff
@@ -296,6 +316,10 @@ public static void main(){
 
 				Gdx.gl.glClearColor(0, 0, 0, 0);
 
+				Vector2f size = vp.size();
+				float near = scene.camera.near();
```
SolarLune: These can just be plugged in directly below, right? `near` and `far` are only used once.

src/com/nilunder/bdx/Bdx.java
```diff
@@ -305,23 +329,23 @@ public static void main(){
 					filter.setUniformf("time", Bdx.time);
 					filter.setUniformi("lastFrame", 1);
 					filter.setUniformi("depthTexture", 2);
-					filter.setUniformf("screenWidth", Bdx.display.size().x);
-					filter.setUniformf("screenHeight", Bdx.display.size().y);
-					filter.setUniformf("near", scene.camera.near());
-					filter.setUniformf("far", scene.camera.far());
+					filter.setUniformf("screenWidth", vp.x);
```
SolarLune: These need to be `vp.size().x` and `vp.size().y` to point to the size of the viewport / screen, not the position.

SolarLune:
Hey, I checked it and suggested some stuff. Otherwise, it seems OK. I'd have to do some stuff to try Gearend with it, but I don't think it's necessary to wait for that. After the issues I suggested are resolved, I think it'd be good to go.

Also, could we set `Scene.cameras` to be of type ArrayListNamed<Camera>?

rafcolson:
Thanks for those corrections @SolarLune. You can inspect the changes in the "corrections" commit.

>Why'd you put these initialized variables in this block instead of directly setting them above ...

Because there is no need for that. I wonder, because I never seen anyone of you use this initialization method, there's nothing wrong with it, is it?

I generally prefer to initialize everything that can be initialized at creation time in a separate block, for better readability. But I guess I could use the function instead.

>These can just be plugged in directly below, right? near and far are only used once.

Not if there's more than one active filter, right? I think it's a good practice to store those variables. But I guess it's negligible for performance.

>These need to be `vp.size().x` and `vp.size().y` to point to the size of the viewport / screen, not the position.

Dumb mistake, but the actual size of the viewport is `vp.w` and `vp.h`, right?

I should mention that I can't manage to avoid blurring of the filters when sizing viewports. Regardless of using `vp.size()` instead of `vp.w` and `vp.h`, which I've been trying too, it seems that the rendered image of the filter doesn't have the same resolution as the rendered image of the viewport. I told you before that I can't really understand the stuff you did with filters, so I wanted to ask you if you could maybe fix this? Here's a [test blend](https://www.dropbox.com/s/x4g6gy2lvvm9zxl/viewport_test.blend?dl=0) I've been using to compare results.

So maybe it would be better for @GoranM to wait until you have checked this. Although it's a minor thing, I wouldn't want you to get unexpected surprises during your project. Just give a sign if it's not a big deal, or if you're able to fix the issue with sized viewports and filters.

>Also, could we set `Scene.cameras` to be of type `ArrayListNamed`?

I had liked that too, but I intentionally kept this change out too, to respect @GoranM 's request to not break the API for this commit. I'll change it for now, because I think Goran would agree with this, but if not, I guess I'll be hearing it.

PR updated.

GoranM:
> I'd have to do some stuff to try Gearend with it, but I don't think it's necessary to wait for that.

It may not be necessary, but it would be prudent.

> I had liked that too, but I intentionally kept this change out too, to respect @GoranM 's request to not break the API for this commit. I'll change it for now, because I think Goran would agree with this

I don't think that would break anything; ArrayListNamed inherits from ArrayList, so it should have the same methods.

rafcolson:
>Although it might be best to get rid of x, y, w, and h, as they don't apply immediately when you set those fields. So someone could set the x and y of a viewport

Someone could do that, but shouldn't if those fields aren't exposed in the API. This is no exception, because we have more unexposed public fields in BDX. But if it's really a problem then I could add two unexposed methods `Viewport.xy()` and `Viewport.wh()`, which then are used in `Bdx.java` and `Pointer.java`.

>I checked the blend file, and the filtered result looks like it should, I believe. Does it seem overly blurry to you?

Compare results when the filter is active and when not. If not, than nothing is blurred when the viewport is sized. If it is, than there is little blurring. It may not be a big problem, but it shouldn't be like that, should it. The filter should have the exact same pixel dimensions as the viewport. With screen the issue is problematic when sizing the screen. So I think there's also a problem in `Bdx.resize()` regarding filters.

>It may not be necessary, but it would be prudent.

I agree. We don't want things to go wrong for you. There's no haste for this PR to be merged (right @GoranM ?). So take your time.

SolarLune:
> Someone could do that, but shouldn't if those fields aren't exposed in the API. This is no exception, because we have more unexposed public fields in BDX that shouldn't be tempered with. But if it's really a problem then I could add two unexposed getters Viewport.xy() and Viewport.wh(), which then are used in Bdx.java and Pointer.java. Let me know if that works for you.

That's unnecessary; couldn't you just use the position and size functions? But yeah, I guess if it's unexposed, it'd be OK.

> Compare results when the filter is active and when not. If not, than nothing is blurred when the viewport is sized. If it is, than there is little blurring. It may not be a big problem, but it shouldn't be like that, should it. The filter should have the exact same pixel dimensions as the viewport.

Yeah, I looked, and there wasn't any blur - it looked the same? Could you show a screenshot showing the blurring?

> I agree. We don't want things to go wrong for you. There's no haste for this PR to be merged (right @GoranM ?). So take your time.

OK, I'll test it when I can.

rafcolson:
>couldn't you just use the position and size functions?

No. As I explained before, the xywh values are not always the same as position() and size(), and they shouldn't be, as those getters don't get the actual position and size of the viewport with EXTEND, LETTERBOX and SCREEN. You can see this clearly in Viewport.update(), where the xywh values are updated.

>Yeah, I looked, and there wasn't any blur - it looked the same. Could you show a screenshot showing the blurring?

That's strange. You are using the test blend, right?

viewport_blurring_with_filter

SolarLune:
OK, so I tested it, and Gearend still works correctly.

> That's strange. You are using the test blend, right?

OK, I just saw it; it's because ScreenShader.nearestFiltering determines whether nearest or linear filtering is applied when drawing the filtered result on-screen, as you may or may not want that filtering. For example, when downsampling a bloom filter, you probably want linear filtering, but you also might want nearest for certain situations.

rafcolson:
>OK, I just saw it; it's because ScreenShader.nearestFiltering determines whether nearest or linear filtering is applied when drawing the filtered result on-screen, as you may or may not want that filtering. For example, when downsampling a bloom filter, you probably want linear filtering, but you also might want nearest for certain situations.

So I guess this is just something that comes with using filters. Thanks for that info. @GoranM I think we can proceed with a merge now, since everything seems to be okay for @SolarLune.

Issue #694, opened 2016-06-20
Bug: SpriteAnim.frame() returns < 0 for negative speed


SolarLune:
Hey, there. 

So if SpriteAnim.speed is less than 0, then `frame()` returns negative numbers (i.e. 0, -1, -2, and then it wraps).

I can fix this, but something else that came up is that SpriteAnim ignores the first frame of the animation when you first play it (so when playing an animation that goes {0, 1, 2, 1}, it would display {1, 2, 1, 0, 1, 2, 1, 0...}). Any ideas on this?

rafcolson:
>SpriteAnim ignores the first frame of the animation when you first play it (so when playing an animation that goes {0, 1, 2, 1}, it would display {1, 2, 1, 0, 1, 2, 1, 0...}).

It doesn't appear that way on my end. I replaced `Sacky.java` with:
```java
import com.nilunder.bdx.*;
import com.nilunder.bdx.components.SpriteAnim;

public class Sacky extends GameObject{
	
	public GameObject visual;
	public SpriteAnim sa;
	
	public void init(){
		visual = children.get("G_Sacky");
		sa = new SpriteAnim(visual, 32, 32);
		sa.add("startle", 0, new int[]{0, 1, 2, 3}, 1, true);
		sa.play("startle");
		visual.components.add(sa);
	}
	
	public void main(){
		if (Bdx.keyboard.keyHit("space")){
			Bdx.restart();
			System.out.println("RESTARTS");
		}else{
			System.out.println(sa.frame());
		}
	}
	
}
```
This returns: `-1, 0, 0, 0, ... , 1, 1, 1, ... , 2, 2, 2, ....`.

However, the first frame is always `-1`. I'm not sure why..

GoranM:
A packed .blend that demonstrates the bug/s clearly would be helpful.

SolarLune:
OK, I guess we should start with the first problem, then. Here's the problem blend file.

https://www.dropbox.com/s/dybuesdby5taof3/NegativeSpeedProblemPacked.blend?dl=0

GoranM:
The important thing to note here is that the actual in-game frames are displayed in correct sequence, both when speed is positive, and when speed is negative.

So, if frame() is returning incorrect values, it's highly likely that the bug is contained within that method.

Hint: Remember why you changed the code from simply returning the `active.playHead`, to returning `active.playHead - 1`? What does that assume about playback direction?

SolarLune:
[...]

PR #716, opened 2016-07-11
(#694) - Fix to SpriteAnim.frame().

SolarLune:


GoranM:
Unless there's a good reason not to, I think it would be best to just use `active.playDir`.

SolarLune:
That should do it. PR updated.

Issue #718, opened 2016-07-11
Framerate and Profiler is Per-Scene


SolarLune:
Hey, there. So basically, if you enable Framerate and Profiler and then disable it in another scene that gets added later, the profiler isn't visible.

Basically, the framerate and profiler should be set by default only once from Blender; apart from that, it should only be altered via code by the user.

GoranM:
It's probably a good idea to summon @rafcolson for anything profiler related.

rafcolson:
>It's probably a good idea to summon @rafcolson for anything profiler related.

I'll take a look at it today. There's also another related issue that should be addressed. When the profiler is enabled in multiple scenes Bdx generates duplicates of the font.

rafcolson:
>if you enable Framerate and Profiler and then disable it in another scene that gets added later, the profiler isn't visible.

Can't reproduce. What I did is enable Framerate and Profile, create an overlay scene (in which the Framerate and Profile is disabled by default) and added the code which adds the overlay scene in `Sacky.init()`. The profiler is visible when running the game.

SolarLune:
> However, when it's the other way around, then the profiler isn't visible. So, when the overlay has Framerate and Profile enabled, but the main scene hasn't.

Hmm, that's not what I'm getting in Gearend - the profiler is visible even though it's disabled in the starting scene. It looks like `put()` (and other functions) call reinitialize() on the Profiler, which means the Profiler can be disabled or enabled by another scene that is created. That might be the problem. 

I guess a way to test it would be to `put()` a property in the props HashMap before adding the overlay scene in `Sacky.init()` in your test file.

rafcolson:
Thanks for that info. This should fix it. However, I also noticed that with current master the profiler's background isn't visible anymore. In any case, this should be fixed with the Viewport commit too.

PR submitted. @GoranM It may be easier for me to adjust this commit after the Viewport commit is already merged. Thanks.

rafcolson:
PR merged. Could this be closed?

SolarLune:
Yeah, okay, sure. Thanks!

PR #721, opened 2016-07-16
(#718) framerateProfile only in main scene & don't end profiler scene & remove warning

rafcolson:


GoranM:
@SolarLune

Does this resolve the issue?

rafcolson:
I'm currently resolving conflicts with master, so...

rafcolson:
PR updated and tested.

SolarLune:
OK, I just tested this. 

It doesn't work the way that I personally think it should; I feel like it should check just the current, "main" scene for the profiler being active, and go with that, rather than loop through all scenes to see if any of them have the profiler enabled.

SolarLune:
I guess an alternative would be to alter Blender's actual behavior by having the Game menu settings (i.e. physics visualization and framerate and profiler) apply across all scenes whenever they're ticked or unticked. Personally, I think I'd prefer that, since it'd solve the problem at its base with turning on a debug setting, going to another scene, and finding it off.

rafcolson:
>It doesn't work the way that I personally think it should; I feel like it should check just the current, "main" scene for the profiler being active, and go with that, rather than loop through all scenes to see if any of them have the profiler enabled.

Okay, I'll change this.

>I guess an alternative would be to alter Blender's actual behavior by having the Game menu settings (i.e. physics visualization and framerate and profiler) apply across all scenes whenever they're ticked or unticked. Personally, I think I'd prefer that, since it'd solve the problem at its base with turning on a debug setting, going to another scene, and finding it off.

That would be a suggestion for a new issue. Feel free to make it at any time. For now, I'll be concentrating on fixing the issue at hand.

rafcolson:
PR updated. @SolarLune Like before, the profiler will only be enabled when it's enabled in the main scene. Give a sign to @GoranM when you think this is good to go.

rafcolson:
Problem fixed.

SolarLune:
OK, I think this is correct, but if you use `put()` and the Profiler's not visible, it'll print "Warning: "Show Framerate and Profile" is not enabled in the main scene" to the console.

Could I ask, why have the Profiler print that?

rafcolson:
It doesn't make sense to add properties if the profiler is not visible. The warning should make the user aware of this.

SolarLune:
I don't really see the point; if the profiler's invisible, then the user won't see the properties they put to the profiler. I don't think the user needs a message to be reminded of this - it's a basic concept of profiling their game. Also, this means that the console's cluttered with the message any time the developer's debugging by using Profiler.props.put. They basically have to remove the `put()` calls if they want to print out any other information.

Perhaps it would be best to remove it.

rafcolson:
>They basically have to remove the put() calls if they want to print out any other information.

Not quite. There is no sense in adding properties if the profiler isn't visible. I'm sure you agree with that. And so, logically, you would first check if the profiler is visible before doing anything else with it.
```java
if (Bdx.profiler.visible()){
	Bdx.profiler.props.put("myProp", "myValue");
}
```
The warning is just a reminder not to alter the scale or the visibility of the tick rate, subsystems, gl, props or add props when the profiler is visible.

That said, I don't mind removing it. Instead I'll put a warning in the API for the Profiler.

SolarLune:
> There is no sense in adding properties if the profiler isn't visible. I'm sure you agree with that. 

I mean yeah, I agree, but it's not an intentional thing. While you're debugging, you might put a value to the props HashMap, enable the Profiler to check the value, run the game, then later disable the Profiler. It's not like you intentionally are trying to put stuff to the Profiler while it's invisible.

> The warning is just a reminder not to alter the scale or the visibility of the tick rate, subsystems, gl, props or add props when the profiler is visible.

What happens if the profiler's invisible when making these changes? Is it that the Profiler doesn't apply the changes?

rafcolson:
>What happens if the profiler's invisible when making these changes? Is it that the Profiler doesn't apply the changes?

I only made sure there wouldn't be exceptions, so some things may be applied, but other things may. Like props can be added and removed, but changing the visibility of the props will not be possible.

SolarLune:
Hmm, I see. That makes sense.

rafcolson:
>Hmm, I see. That makes sense.

So, if you think this is ready, just give a hit, okay? And thanks for the review.

SolarLune:
OK, I tested it, and it seems like it works fine. Thanks!

PR #724, opened 2016-07-18
Bug-fix: Clear color should have no alpha component.

SolarLune:
This way it's possible to tell the difference in a screen shader between the cleared screen (no alpha) and a pixel from a rendered scene (which would have alpha).


GoranM:
Given access to the depth buffer, it should be possible to determine that without relying on alpha.

I don't think it makes sense to restrict clear color to a zero alpha; It can have non-zero values, so a user should be able to set that.

If you want to rely on alpha to isolate specific fragments, you should set the clear color to whatever suits you in that context.

Also, if you want the default clear color to have a zero alpha, I think that would be fine.

SolarLune:
> Given access to the depth buffer, it should be possible to determine that without relying on alpha.

We can access the depth buffer? You mean the depth texture? If so, yeah, that's a possible solution, but that means rendering the scene an entire additional time. I think it's best that that not be used unless absolutely necessary (i.e. for shaders that require use of the depth).

> I don't think it makes sense to restrict clear color to a zero alpha; It can have non-zero values, so a user should be able to set that.

Eh, visually it looks the same as not having alpha, so I don't really know of a reason why you'd want to have an alpha value of 1 for the clear color.

SolarLune:
> Also, if you want the default clear color to have a zero alpha, I think that would be fine

The default clear color doesn't have any alpha, so that's OK there.

GoranM:
> I don't really know of a reason why you'd want to have an alpha value of 1 for the clear color.

Assuming a render to texture scenario, where you don't want the background to be fully transparent, it would be nice if you could actually set the desired alpha value.

I don't see why it's necessary to restrict clear color to zero alpha; If you need it to be zero, you can just set it to zero.

rafcolson:
>I don't really know of a reason why you'd want to have an alpha value of 1 for the clear color.

I wonder if this capability would allow us to specify the color of the area outside a viewport of type LETTERBOX?

SolarLune:
> I wonder if this capability would allow us to specify the color of the area outside a viewport of type LETTERBOX?

I think we'd need to add in the ability to specify the color of the area outside the viewport. I don't think the alpha value would influence that.

> Assuming a render to texture scenario, where you don't want the background to be fully transparent, it would be nice if you could actually set the desired alpha value.

I don't really see the situation where you'd want the background to not be fully transparent, since you can just draw the color you want before drawing the texture (or applying the color beforehand in a shader, or whatever the case), but OK. We can close this, then.

SolarLune:
Oh, I guess I'll just add a note to the API that the alpha value will influence things. That's the best of both worlds.

Issue #723, opened 2016-07-17
Blender Memory Slowdown


SolarLune:
Hey, there. 

So I'm experiencing a slow-down in Blender as I run my BDX project; the time it takes to do memory-related operations like closing the blend file or undoing rises with time, until it starts taking quite a few seconds. 

I'm guessing it's due to the rise in memory allocation. Notice that when you run a BDX project, the memory readout in the header in Blender rises a little bit. For a small project, it's basically nothing, but for a moderately sized one, the usage rises by several megabytes at a time and can fairly quickly become a problem. 

I believe we might be copying some resources in the blend file when we export a run a project, which might be part of (or all of?) the issue.

Anybody got any ideas of where to look regarding this?

GoranM:
We're not creating any additional blender data, other than the profiler scene, which is removed soon after it's created.

The export process does create a lot of temporary python objects (like any other python function that does significant data transformation/manipulation work), but those objects should be garbage collected soon after the export function returns.

I would recommend using the built-in [tracemalloc](https://docs.python.org/3/library/tracemalloc.html) module to narrow it down.

SolarLune:
> I would recommend using the built-in tracemalloc module to narrow it down.

I'll keep that in mind, but first, I might have stumbled upon the core issue. It looks like whenever BDX exports and runs, for whatever reason, a copy of each mesh in the scene is created. You can see this odd behavior here, in this basic example blend file, or in your own project.

https://www.dropbox.com/s/518k1kcqfmk03bi/basicTest.blend?dl=0

Look to the left, at the Outliner. There's 4 meshes, one for each GameObject in the scene. After pressing P to run the game and exiting, note that there's now 8 meshes; 1 "original" version of each mesh, and 1 generically named copy for each GameObject in the scene. In my Gearend project, what was 258 meshes becomes almost 1000 after running the game once (probably because it's running across each game scene in my project file).

It might be due to creating a new mesh [here](https://github.com/GoranM/bdx/blob/master/blender/bdx/exporter.py#L106), in line 106 of exporter.py, and never cleaning that mesh up. [Maybe we need to manually remove it somehow afterwards?](https://www.blender.org/api/blender_python_api_2_76_2/bpy.types.Object.html#bpy.types.Object.to_mesh) (Note "remove if only used for export" at the bottom of that function description.)

GoranM:
Ah, now I remember.

[...]

PR #725, opened 2016-07-18
(#723) - Fixing mesh duplication on export bug.

SolarLune:
Only create a new mesh if there's modifiers applied on it.  Otherwise, use the object's mesh.
Remove the new mesh when done with it.


rafcolson:
That sure was a nasty bug. I was noticing the same problem the last couple of days. I'm glad that it's fixed, so thanks.

SolarLune:
No problem!

Issue #729, opened 2016-07-19
Suggestion: scene.xray() String... props


SolarLune:
Hey, there. 

I think it'd be useful to have a means of filtering raycasts to only hit certain objects, by checking for properties present in hit obejcts. This should be able to be done without breaking the API by tacking on a variable amount of String args at the end of `xray()` functions. That way you can specify which props to filter (rather than just one).

rafcolson:
Sounds good to me.

GoranM:
Ok.

SolarLune:
OK, PR submitted. Note that the properties are all AND, not OR (so objects have to have *all* of the properties you pass to the functions to be recognized).

SolarLune:
The PR was merged, so I'll close this now. Thanks!

PR #731, opened 2016-07-20
#729 - Additional props argument to Scene.rays().

SolarLune:
Allows for filtering collisions based on properties the object holds.

### DIFF
--- a/src/com/nilunder/bdx/Scene.java
+++ b/src/com/nilunder/bdx/Scene.java
@@ -627,11 +627,11 @@ public class Scene implements Named{
 		toBeRemoved.add(g);
 	}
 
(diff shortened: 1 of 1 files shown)


src/com/nilunder/bdx/Scene.java
```diff
@@ -672,7 +677,15 @@ public RayHit ray(Vector3f src, Vector3f vec, short group, short mask){
 
 			if (ray != null){
 
-				if (!hitObjects.contains(ray.object) || includeAll) {
+				boolean skip = false;
+				for (String prop : props) {
```
SolarLune: Note that this is necessary because passing the props to ray() will just return null if the object hit doesn't have the properties specified, which would finish the xray() function prematurely.

Issue #737, opened 2016-07-30
P_mapto_bdxexprun error with no 3D View area


rafcolson:
Nothing major. It's only noticeable when viewing the System Console, and only in a special case.
```python
Traceback (most recent call last):
  File "C:\Users\Raf\AppData\Roaming\Blender Foundation\Blender\2.77\scripts\addons\bdx\ui.py", line 142, in P_mapto_bdxexprun
    kmi = bpy.data.window_managers["WinMan"].keyconfigs["Blender"].keymaps["Object Mode"].keymap_items
KeyError: 'bpy_prop_collection[key]: key "Object Mode" not found'
```
I didn't notice this before, so to be sure, maybe you guys could check if you are getting this error too, after opening Blender with the Startup file, when having a [symlink](https://github.com/GoranM/bdx/wiki/Building-BDX#streamlining) to the blender/bdx/libs folder in your github repo.

Edit: I turned out to be an issue with the 3D View. It has nothing to do with symlinks. Sorry.

SolarLune:
OK, what's the process for testing this? Because I don't think I've ever gotten this error before, and I have a symlink from my project's `core` folder to my GitHub repo's `libs` folder.

rafcolson:
>OK, what's the process for testing this?

I'm getting the error when opening Blender with the Startup file. But I found the actual problem though. The screen of my Startup file doesn't contain a 3D View area. That's why `"Object Mode"` couldn't be found. I should change the try/except to an if clause. I think that would be better.

GoranM:
I'm surprised that entries in the keymap dictionary are apparently not added until the relevant area first becomes visible.

I wonder if this is intended behavior (some sort of "lazy loading"), or a bug in blender.

It would be nice if we could confirm that.

rafcolson:
>I wonder if this is intended behavior (some sort of "lazy loading"), or a bug in blender.

>It would be nice if we could confirm that.

Did you already ask for confirmation, by any chance? Else, I could do that in the [BlenderArtists Python support forum](https://blenderartists.org/forum/forumdisplay.php?11-Python-Support).

GoranM:
No, I didn't ask.

Thinking about it further, it doesn't really matter; Your patch is small enough, and it resolves the problem at hand, so there's no pressing need to dig deeper for a "better" solution (assuming one even exists).

rafcolson:
Okay. Thanks! PR is merged, so closing.

PR #738, opened 2016-07-31
(#737) P_mapto_bdxexprun only if 3d view on screen

rafcolson:


GoranM:
The existence of "Object Mode" in keymaps is the primary dependency, so I think it would make more sense to simply check for that:

```python
kmi = keymaps["Object Mode"].keymap_items if "Object Mode" in keymaps else None

if kmi:
    ...
```

rafcolson:
Yes, that's right. PR updated.

PR #761, opened 2016-08-20
Fix normalized method

agoose77:


SolarLune:
Hey, thanks for the fix!

Issue #763, opened 2016-08-27
Suggestion: Store Descriptors on Inputs


SolarLune:
Hey, there. 

I'm trying to do an input remapping system, and to do this, it'd be useful to if I could get the descriptors (e.g. `k:left` or `g:a`) from the Inputs. However, they don't appear to be available on the Input? It'd be nice to expose those.

GoranM:
Ok.

SolarLune:
OK, PR submitted.

SolarLune:
Hey, this got merged, so I'll close this now. Thanks!

PR #764, opened 2016-08-28
(#763) - Add Descriptor to Input.

SolarLune:


GoranM:
Conceptually, the descriptor is a string that describes the device, and the corresponding input on that device, so I think it would be a bit confusing to expose "descriptor" as a field that references a string array.

A string array would represent a descriptor in its parsed form, so if you want to expose that, I think something like "parsed_descriptor" would be a more suitable name.

Although, I'm not sure if parsed_descriptor should be a variable element array: It doesn't fit with the aforementioned "device:input" concept (which is fundamentally a 2 element thing), and it doesn't really offer a greater convenience (checking for the length of the array is no less difficult than checking the first element). So, with that in mind, I think it would make sense to just do:

```java
this.parsed_descriptor = d;
```

SolarLune:
> Conceptually, the descriptor is a string that describes the device, and the corresponding input on that device, so I think it would be a bit confusing to expose "descriptor" as a field that references a string array.

Oh, so basically if anything, it should be the same descriptor that was passed to create the Input in the first place. I suppose so; maybe `parsedDescriptor` would be better? 

> It doesn't fit with the aforementioned "device:input" concept (which is fundamentally a 2 element thing)

Well, gamepads have IDs, so "g0:a" isn't the same as "g1:a"; do you mean that the descriptor should be, for example, `"g0", "a"` for a gamepad descriptor of "g:a" or "g0:a"?

> So, with that in mind, I think it would make sense to just do:

So it shouldn't add the ID of the gamepad as the second element of the descriptor, but just have it as part of the first element of the String array?

P.S. I think camelCase is the more Java-oriented naming convention?

GoranM:
Yes.

SolarLune:
OK, PR updated.

SolarLune:
Wait, hold up, something's off with that `gpIndex` final at the end...

SolarLune:
OK, that should do it. PR's good to go.

SolarLune:
PR was merged, so I'll go ahead and close this. Thanks!

Issue #693, opened 2016-06-19
Suggestion: Option for higher res text


Lucrecious:
Hey!

I know that there has been a thread for this before, but this is going at the problem with a different approach.

Right now we're using Hiero to generate BitMapType fonts. This becomes a little bit of a hassle when we are not using blocky fonts or want big text. Either we make the text small enough so the pixelation is hidden or we play around with Hiero to get a higher resolution text. I don't think any of these solutions are very good.

My suggestion is my not have the option to use to use the FreeTypeFontGenerator?
https://github.com/libgdx/libgdx/wiki/Gdx-freetype

I understand that we might have to give up a little bit of speed, but if text is used sparingly, this shouldn't be much of a problem. I'm not suggesting to remove the already implemented text handling, just have another for anyone who wants to give up a little bit of frames for nicer text.

Anyways, here is a rough result with FreeType I did with BDX (very hacky solution right now):
![image](https://cloud.githubusercontent.com/assets/4584511/16179119/bc11df28-362a-11e6-8d57-146c30dd45c0.png)
This is 72 font size on a 1080x720 display. Looks pretty nice if you ask me.

Sadly, it seems that my hacky solution killed the profiler, so I wasn't able to check the ongoing frame rate.

I'd be happy to implement this feature, just wondering if anyone else is interested?

rafcolson:
This would be very cool, but why not replace Hiero altogether? It would be great if Text objects in BDX wouldn't have such a low resolution and wouldn't have incorrect sizes and spacings. If this can be fixed with FreeTypeFontGenerator along with getting these extra features for text, then wouldn't that be worth the change?

You can get the tickRate like this:
```java
float f = Bdx.profiler.avgTickRate;
```

Lucrecious:
Like I said on the post, I think the way Goran has implemented the Text in BDX allows for text in 3D space. This would be an overlay over everything (I'm not quite sure how to do this in 3D space, I think it'd be hacky at best).

rafcolson:
Ah, okay. Thanks for reminding me.

SolarLune:
Hmm, well, we could add something that would render objects on top of the scene, like the BGE's `postDraw` callback.

[...]

PR #781, opened 2016-09-16
(#693) - FontWriter for improved text in BDX.

SolarLune:


GoranM:
Why are you keeping gdx-tools.jar? I don't think we need it anymore, since we don't intend to continue using Hiero from this point forward.

Also, why is there more than one copy of fontwriter.jar? I don't think it should be a part of libs, as it's not a required runtime library.

Is it still generating more than one texture in certain cases? As far as I remember, Hiero didn't do that, so this is something that needs to be cleared up before we can merge.

SolarLune:
Yo, sorry about that. PR updated to remove those extra JARs.

Yeah, it generates another texture if the font size makes the font take up too much space for the texture space. I don't know if Hiero did that or not - I believe our font size was pre-set at such a small size to the texture space that it never came up. The default I have set is 32px for the font size, and 512x512 for the texture resolution size.

GoranM:
I don't think Hiero had an option to set the texture resolution. You could set the font point size, and then I'm assuming that Hiero would create a closest power of 2 texture that would actually fit all the characters at their selected point size.

Is there a way to configure fontwriter to do it that way, instead of cutting things up into multiple textures of some preset size?

SolarLune:
> Is there a way to configure fontwriter to do it that way, instead of cutting things up into multiple textures of some preset size?

Not to my knowledge, no. I don't see anything in the options, either.

> I'm assuming that Hiero would create a closest power of 2 texture that would actually fit all the characters at their selected point size.

Did you ever test it to confirm this is the way it worked? Because the only reason why I know this one'll split the fonts up into multiple images is because I tried larger font sizes - as I recall, images generated with Hiero has quite a lot of dead space in them due to the small font size?

GoranM:
> Did you ever test it to confirm this is the way it worked?

I just tested it, and Hiero actually does the same split, given a high enough point size (something higher than 72).

In hiero, you can only specify the point size, and the texture resolution is locked to 512x512. 

I suppose the issue can be solved with fontwriter by just retrying the operation with a higher texture resolution, if it previously generated more than one set of files. To do that, I think fontwriter would have to manage texture resolution internally, rather than letting the user define it.

The user would just define the point size, and other properties, and then font writer would set the texture resolution into which that point size can fit.

SolarLune:
> I suppose the issue can be solved with fontwriter by just retrying the operation with a higher texture resolution, if it previously generated more than one set of files. To do that, I think fontwriter would have to manage texture resolution internally, rather than letting the user define it.

I looked into a solution for this, but I haven't found anything particularly nice. It seems that LibGDX doesn't write the outputted textures to disk until the application actually closes (or something like that), because I can't actually "see" the extra created files in the OS or in FontWriter until the app closes. 

I also tried moving the generation code into the render() function, to ensure the app runs over multiple frames, and that didn't do anything, either.

So I guess we'd have to run it multiple times from Python?

GoranM:
> It seems that LibGDX doesn't write the outputted textures to disk until the application actually closes (or something like that)

That sounds really unusual.

Unless the work is being offloaded to a different thread, the data should be on the filesystem immediately after the write methods are called.

We need some concrete information on what's actually happening there, and why - Someone in the libgdx community should be able to explain it.

SolarLune:
OK, I've posted a thread up over [there](http://www.badlogicgames.com/forum/viewtopic.php?f=11&t=23581&p=96380#p96380). We'll see what happens from there.

SolarLune:
Still no responses on the thread on the forums. I did a quick Google, thinking that maybe there was some function to "flush" the FontWriter after writing the font image, but I didn't see anything.

Alternatives to proceed seem to be:

1) Leave it as it is. It works the same as Hiero did, so it's not like the current (understandably less than ideal) behavior is anything "new", it's just new to us now because we can change the font size, so we see it easily.

2) Spitball a function to auto-calculate texture size (so a font size of 16, say, generates a 256x256 texture, a font size of 32 generates a 512x512 texture, and so on). From [this chart](http://reeddesign.co.uk/test/points-pixels.html), it looks like it might be that the pixel size is approximately 1.33 * the font point size (from the chart, at least), so it might be possible to calculate the texture size pretty reliably for most fonts.

3) Run the command multiple times through Python. I haven't tested it, but it seems like it would work (though it is annoying to have to work in multiple languages for a single feature).

GoranM:
> Still no responses on the thread on the forums.

Once the thread drops off the front page, it's the probability of response drops significantly, so maybe you just got unlucky with your timing - Try giving it a bump.

They also have an irc channel, and I've seen many of them offer support on stackoverflow, so it might help to ask there as well.

You could also try to find out who is largely responsible for writing/maintaining BitmapFontWriter, and ask them for help directly.

I think you should also include the code that you used to test if the files are on the filesystem, as that's a pretty important, and it's not visible in the version of DesktopLauncher.java that you linked.

> It works the same as Hiero did

We didn't have to worry about multiple files with the Hiero system, because the point size was locked to a number low enough to fit comfortably into a 512x512 texture.

I really don't want to take the "spitball" or "repeat run" approach to this, unless there's absolutely no other way, and I don't think we have enough evidence to conclude that at this point.

SolarLune:
Bumped the thread a few days ago and updated with a basic explanation of how I checked to see when the files were written to disk (using LibGDX's FileHandle retrieval functions and looking at the file system itself) - no responses. Also tried IRC, but nobody really had any ideas about it.

At this stage, I believe there's not really any alternatives available to pressing forward with this apart from spitballing a guess as to the texture size, or running the .jar multiple times with values for the texture size.

GoranM:
> a basic explanation of how I checked to see when the files were written to disk

I don't think that's sufficient, because there could still be some subtle issue with the actual code you ultimately used.

As I said before, I think you should include the actual code.

> Also tried IRC, but nobody really had any ideas about it.

Did you try stack overflow, or contacting whoever is actually responsible for BitmapFontWriter directly, as I suggested earlier?

> At this stage, I believe there's not really any alternatives

At this stage, we don't really have enough information to reach a reasoned conclusion.

SolarLune:
> As I said before, I think you should include the actual code.

I don't think the code would do much, as the file isn't showing up in the file system - I'm pretty sure it's not being outputted until the application closes. The method I'm using also returns false when the file doesn't appear in the file system (on first run), and true otherwise (on subsequent runs, after the file is shown in my file manager), with no tweaking on my end.

But in either case, I've updated the forum posting with the code that I'm using to test whether the file is there or not. Here's the method in brevity:

```java
int t = 0;
			
@Override
public void render() {
				
	FileHandle path = Gdx.files.absolute("/home/solarlune/Documents/Projects/BDX/NewTest/android/assets/bdx/textures/__FNT_Bfont.png");
	System.out.println(path.exists());
				
	t++;
	if (t > 100)
		System.exit(0);
				
}
```

I added a simple timer to make sure it's not a matter of it just taking a second or something - this waits for roughly 1.5 seconds, and still, the same result. It's also the same result if (t > 1000). The font texture only appears in the file system after the application closes, and not before.

> Did you try stack overflow, or contacting whoever is actually responsible for BitmapFontWriter directly, as I suggested earlier?

OK, I'll try pinging who I suspect might be responsible for BitmapFontWriter based on the LibGDX [history](https://github.com/libgdx/libgdx/commits/master/extensions/gdx-tools/src/com/badlogic/gdx/tools).

_______

@NathanSweet - Hey, there. Sorry for the ping out of left-field, but I was wondering if you knew anything about our issue here? The problem we're having is with when the BitmapFontWriter class outputs its files. It seems like it only actually writes the texture _after_ the application exits. Do you think you could confirm that this is the case? Or maybe is there a way to flush the writer to make it output immediately?

As a side-note, the reason why we're doing this is because we're trying to get the BitmapFontWriter to output a font texture that's large enough to contain all of the outputted glyphs at whatever point size is specified. If the point size is too large, the BitmapFontWriter'll automatically spread the glyphs over multiple font textures. Do you know if it's possible at all to specify somewhere to the FontWriter to keep all of the glyphs on a single texture, just sizing it up as necessary?

GoranM:
> Here's the method in brevity

I think it would be best if you were to post the whole program, because you could be leaving out something that seems irrelevant to you, but that is actually very important for the issue at hand.

SolarLune:
OK, here it is on [PasteAll](http://pasteall.org/83310/java), and here it is here:

```java
package com.solarlune.fontwriter;

import com.badlogic.gdx.ApplicationAdapter;
import com.badlogic.gdx.Gdx;
import com.badlogic.gdx.backends.lwjgl.LwjglApplication;
import com.badlogic.gdx.backends.lwjgl.LwjglApplicationConfiguration;
import com.badlogic.gdx.files.FileHandle;
import com.badlogic.gdx.graphics.Color;
import com.badlogic.gdx.graphics.Pixmap;
import com.badlogic.gdx.graphics.g2d.PixmapPacker;
import com.badlogic.gdx.graphics.g2d.freetype.FreeTypeFontGenerator;
import com.badlogic.gdx.tools.bmfont.BitmapFontWriter;

import java.util.ArrayList;

public class DesktopLauncher {
	public static void main (String[] arg) {
		LwjglApplicationConfiguration config = new LwjglApplicationConfiguration();
		config.title = "FontWriter";
		config.width = 666;
		config.height = 444;

		ArrayList<String[]> files = new ArrayList<>();

		for (String a : arg)
			files.add(a.split("---"));

		new LwjglApplication(new ApplicationAdapter() {
			
			public void create() {

				for (String[] commands : files) {

					String inputFontPath = commands[0];        // Supply the full path
					FileHandle outputFolder = Gdx.files.absolute(commands[1]);
					String fileName = outputFolder.nameWithoutExtension();
					outputFolder = outputFolder.parent();

					int resX = Integer.valueOf(commands[2]);
					int resY = Integer.valueOf(commands[3]);
					int fontSize = Integer.valueOf(commands[4]);
					int shadowOffsetX = Integer.valueOf(commands[5]);
					int shadowOffsetY = Integer.valueOf(commands[6]);
					Color shadowColor = new Color(Float.valueOf(commands[7]), Float.valueOf(commands[8]), Float.valueOf(commands[9]), Float.valueOf(commands[10]));
					int outlineThickness = Integer.valueOf(commands[11]);
					Color outlineColor = new Color(Float.valueOf(commands[12]), Float.valueOf(commands[13]), Float.valueOf(commands[14]), Float.valueOf(commands[15]));
					boolean outlineRounded = Boolean.valueOf(commands[16]);

					BitmapFontWriter.FontInfo fontInfo = new BitmapFontWriter.FontInfo();

					fontInfo.padding = new BitmapFontWriter.Padding(1, 1, 1, 1);

					FreeTypeFontGenerator generator = new FreeTypeFontGenerator(Gdx.files.absolute(inputFontPath));
					FreeTypeFontGenerator.FreeTypeFontParameter parameter = new FreeTypeFontGenerator.FreeTypeFontParameter();

					parameter.color = new Color(Float.valueOf(commands[17]), Float.valueOf(commands[18]), Float.valueOf(commands[19]), Float.valueOf(commands[20]));
					parameter.borderColor = outlineColor;
					parameter.borderWidth = outlineThickness;
					parameter.borderStraight = !outlineRounded;
					parameter.shadowOffsetX = shadowOffsetX;
					parameter.shadowOffsetY = shadowOffsetY;
					parameter.shadowColor = shadowColor;
					parameter.size = fontSize;

					parameter.packer = new PixmapPacker(resX, resY, Pixmap.Format.RGBA8888, 2, false, new PixmapPacker.SkylineStrategy());
					FreeTypeFontGenerator.FreeTypeBitmapFontData data = generator.generateData(parameter);

					BitmapFontWriter.writeFont(data, new String[]{fileName + ".png"}, outputFolder.child(fileName + ".fnt"), fontInfo, resX, resY);    // Writes the .fnt file, I guess
					BitmapFontWriter.writePixmaps(parameter.packer.getPages(), outputFolder, fileName);

					generator.dispose();
					
					FileHandle path = outputFolder.parent().child("textures").child("__FNT_" + fileName + ".png");
					
					System.out.println(path);
					System.out.println(path.exists());

				}

			}
			
			int t = 0;
			
			@Override
			public void render() {
				
				FileHandle path = Gdx.files.absolute("/home/solarlune/Documents/Projects/BDX/NewTest/android/assets/bdx/textures/__FNT_Bfont.png");
				System.out.println(path.exists());
				
				t++;
				if (t > 1000)
					System.exit(0);
				
			}
			
		}, config);

	}

}

```

GoranM:
I don't think there's a need to isolate the check code in the render method. You should be able to check for files right after the code that writes the files in the create method.

-

The texture file is specified as `fileName + ".png"`, but you're checking for `"__FNT_" + fileName + ".png"`, which is not the same. Although, even if it was the same, it still wouldn't work, because you're looking in the wrong directory.

The java program writes the textures to the same directory as the .fnt files. The blender exporter then moves them to the textures directory, and renames them to have the "__FNT_" prefix, as can be seen [here](https://github.com/GoranM/bdx/blob/master/blender/bdx/exporter.py#L745-L747).

SolarLune:
Sweet, you figured it out! Thanks a ton! 

It was indeed that the files were being moved after renaming - the textures were showing up there in the fonts folder itself. I guess that teaches me to jump to conclusions, haha. I'll have to be more careful next time.

PR updated!

@rafcolson - After this is merged, feel free to look into tweaking things to make the profiler font clearer to read using shadows or outlines or a higher resolution texture (if you want to).

rafcolson:
@SolarLune - I will. This might be a important change to the engine, so thanks very much for all the time and effort putting into this FontWriter.

SolarLune:
No problem - glad it's done (hopefully, haha)!

Issue #780, opened 2016-09-14
Suggestion: GameObject.recalculatePhysicsMesh()


SolarLune:
Hey, there. So basically, there's no way to update the physics mesh of the game object should you adjust it? It'd be nice to have a `recalculatePhysicsMesh()` function to do this for us. 

We might later move the function (along with the other model-related classes like `useUniqueModel()` and `replaceModel()`) into the Mesh class, as well?

GoranM:
Ok.

SolarLune:
This was handled by the PR for #680, so I'll close this now. Thanks!

Issue #680, opened 2016-06-10
Suggestion: move "Model" methods to gl.Mesh


rafcolson:
I think it's a bit confusing to be having `Scene.models`, `GameObject.modelName()`, `GameObject.useUniqueModel()`, `GameObject.replaceModel()` along with `GameObject.mesh`, `GameObject.joinedMeshObjects` and `GameObject.updateJoinedMesh`. Especially when a Mesh is supposed to wrap the Model.

So I would like to suggest `Scene.meshes`, `Mesh.name()`, `Mesh.unique()` and `Mesh.replace()`.

SolarLune:
That all sounds fine to me.

I think currently, BDX creates a new Mesh to wrap the Model and provide editing functions whenever necessary, but it doesn't use that Mesh when drawing objects and stuff. Shouldn't the Mesh have a ModelInstance, or be the means through which BDX renders GameObjects? It seems kinda weird that users use GameObject.mesh to edit the model, and BDX uses GameObject.modelInstance to render it. 

So basically, maybe we should move GameObject.modelInstance to the Mesh class for BDX to use?

rafcolson:
I'm not sure. We couldn't have `Scene.meshes` contain references to the modelInstance, then could we? Because that wouldn't make much sense. For which case I still think it had been better if the mesh would be part of the modelInstance. And not having Mesh wrap the model, but the be an extension of LibGdx's Mesh. I pictured this kind of hierarchy:

GameObject -> Meshes (ext. ModelInstance / replace mesh method here - set(index, Mesh)) -> Mesh (ext. LibGdx's Mesh)

SolarLune:
I believe our Mesh class contains a reference to the Model, so I don't quite get how GameObject.meshes should extend LibGDX's ModelInstance.

> We couldn't have Scene.meshes contain references to the modelInstance, then could we? 

Well, scene.meshes could contain references to BDX Meshes, each of which would have the stuff to render with (including a reference to the correct Model).

_____

But never mind; I guess it doesn't really matter as long as it's clear and consistent. For now, just moving the functions into the Mesh class would be fine. We can refactor later if it doesn't work or make sense.

GoranM:
Sounds fine.

> Shouldn't the Mesh have a ModelInstance, or be the means through which BDX renders GameObjects? It seems kinda weird that users use GameObject.mesh to edit the model, and BDX uses GameObject.modelInstance to render it.

I think we discussed that in #610.

[...]

PR #785, opened 2016-10-02
(#680, #780) - Mesh refactoring.

SolarLune:
- Materials are now part of the Mesh, and so are linked to the Mesh (so changing the color of a material on a mesh will change the color of the material on all instances of the same Mesh). To make a material copy, either assign a new Material(), or assign a new Mesh using Mesh.copy(). The default material is the exception to this rule - if an object has no material assigned to it in Blender, it will get its own copy of the defaultMaterial.

- Model disposal is now handled by the Mesh on finalize().

+ updateBody() function to update the physics mesh of a GameObject, either using the existing display mesh, or another mesh if one is provided.

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -1,6 +1,7 @@
 package com.nilunder.bdx;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -8,7 +9,6 @@ import javax.vecmath.Matrix3f;
 import javax.vecmath.Matrix4f;
 import javax.vecmath.Vector3f;
 
-import com.badlogic.gdx.graphics.Mesh;
 import com.badlogic.gdx.graphics.GL20;
 import com.badlogic.gdx.graphics.VertexAttributes.Usage;
 import com.badlogic.gdx.graphics.g3d.Model;
@@ -32,6 +32,7 @@ import com.bulletphysics.linearmath.MatrixUtil;
 import com.bulletphysics.linearmath.Transform;
 
 import com.nilunder.bdx.gl.Material;
+import com.nilunder.bdx.gl.Mesh;
 import com.nilunder.bdx.utils.*;
 
 public class GameObject implements Named{
@@ -42,7 +43,6 @@ public class GameObject implements Named{
 	public ArrayListGameObject touchingObjectsLast;
 	public ArrayList<PersistentManifold> contactManifolds;
 	public ModelInstance modelInstance;
-	public com.nilunder.bdx.gl.Mesh mesh;
 	public RigidBody body;
 	public String currBodyType;
 	public String currBoundsType;
@@ -64,12 +64,11 @@ public class GameObject implements Named{
 	private Vector3f localScale;
 	private boolean visible;
 	private boolean valid;
-	private Model uniqueModel;
-	public ArrayListMaterials materials;
 	public boolean initialized;
 	public int logicFrequency;
 	public float logicCounter;
 	private Vector3f scale;
+	private Mesh mesh;
 
 	public class ArrayListGameObject extends ArrayListNamed<GameObject> {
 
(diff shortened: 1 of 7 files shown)


rafcolson:
Nice job @SolarLune. Thanks very much! But I noticed that that there are some indentation inconsistencies.

SolarLune:
Thanks for the heads-up, I fixed the indentation. I also prioritized those two functions to look for in the calling GameObject's current scene first.

GoranM:
The copy method doesn't actually copy the existing mesh. It just creates a new mesh based on the original json data, which doesn't include changes that were made since initialization.

-

The models of mesh objects in scene.meshes seem to be disposed twice; When the scene is disposed, and when the mesh finalizer is invoked.

I'm not sure if that can cause problems, but it may be a good idea for finalizers to only dispose models of copied meshes.

-

There seem to be a lot of changes where you're just removing a line with some tabs (which follow the natural indent of the block) with an empty line.

Unless there's a good reason for that, it shouldn't be included in the PR, because it just needlessly clutters the diff.

-

It seems like we're referencing our Mesh type more often than the libgdx Mesh, so I think it would be cleaner to refer to our Mesh by just "Mesh", and use fully qualified names for the libgdx Mesh.

SolarLune:
OK, I think I fixed everything. Might've missed a line or two somewhere with the formatting, though.

src/com/nilunder/bdx/GameObject.java
```diff
@@ -735,107 +689,149 @@ public void alignAxisToVec(int axis, Vector3f vec){
...
+	public void mesh(String meshName){
+
+		Mesh m = null;
+
+		ArrayList<Scene> sceneList = new ArrayList<Scene>(Bdx.scenes);
+		if (scene != null) {
+			if (sceneList.indexOf(scene) >= 0)
+				Collections.swap(sceneList, sceneList.indexOf(scene), 0);
+			else
+				sceneList.add(0, scene);
```
rafcolson: I don't understand this clause. Isn't the index of the current scene always larger or equal than zero? So couldn't you just do this instead?
```java
Collections.swap(sceneList, sceneList.indexOf(scene), 0);
```
Or maybe in some cases the current scene isn't part of `Bdx.scenes` yet, at initialization?

src/com/nilunder/bdx/GameObject.java
```diff
@@ -735,107 +689,149 @@ public void alignAxisToVec(int axis, Vector3f vec){
...
-			if (sce.models.containsKey(modelName)){
-				model = sce.models.get(modelName);
-				mOrigin = sce.json.get("origins").get(modelName);
-				mDimNoScale = sce.json.get("dimensions").get(modelName);
+
+		ArrayList<Scene> sceneList = new ArrayList<Scene>(Bdx.scenes);
+
+		if (scene != null) {
+
+			if (sceneList.indexOf(scene) >= 0)
```
rafcolson: The same here.

src/com/nilunder/bdx/GameObject.java
SolarLune: Yeah, apparently, GameObjects can not have their scene be a part of the scene list at initialization (I think it's the templates that are populated before the actual GameObjects are spawned in their positions). 

It's also possible to try to swap a mesh or update the physics body on a GameObject that isn't in the scenes list. For example, you go from scene "A" to scene "B", but hold scene "A" in memory. You could then swap out a mesh on a GameObject without it actually being in the scenes list.

rafcolson:
Ah okay. That explains it.

SolarLune:
OK, I just moved a line around in Scene.java to be able to remove the `if (scene != null)` checks.

SolarLune:
What do you guys think about adding an option to have each GameObject automatically use a copy of a mesh at all times? The central idea behind this is that it fundamentally de-links the GameObject from sharing meshes and materials with other GameObjects. This way, you don't have to assign a copy of a mesh to a GameObject whenever you want to alter its mesh or materials individually - it'll automatically set the copy for each and every GameObject.

rafcolson:
>What do you guys think about adding an option to have each GameObject automatically use a copy of a mesh at all times by default?

If added, how would it be implemented?

I guess it would make things slightly more convenient, doing:
```java
GameObject particle = scene.add("Particle", true);
```
Or maybe:
```java
GameObject particle = scene.addUnique("Particle");
```
Instead of:
```java
GameObject particle = scene.add("Particle");
particle.mesh(particle.mesh().copy());
```

SolarLune:
I was thinking that it would just be an option you'd enable in Blender - a switch that you could trigger, which would set a global boolean value somewhere, like in `Bdx`'s namespace. That way, you could disable it if desired mid-game as necessary.

When it's on, mesh replacement and material replacement functions always use a copy of the material or mesh provided.

SolarLune:
Ah, I noticed that I forgot to fix it to use BDX's Mesh class in a couple of lines. I'll fix those next after we get a decision on this.

GoranM:
> What do you guys think about adding an option to have each GameObject automatically use a copy of a mesh at all times by default?

It doesn't really seem necessary.

If you think it's essential, and you a have a good argument for it, I think that's something we should discuss in a different issue.

SolarLune:
OK, I think it should be good now, then.

SolarLune:
Whoops, sorry about that - the commit got mixed up in some old ones. It should be fine to merge now.

Issue #790, opened 2016-10-05
Bugs: Blender 2.78 Python requires extra arg with removing scenes and texts


rafcolson:
Just for our information. The newest version of Blender causes at least one compatibility issue:

Compare [BlendDataScenes.remove(scene)](https://www.blender.org/api/blender_python_api_2_77_0/bpy.types.BlendDataScenes.html?highlight=blenddatascenes) with [BlendDataScenes.remove(scene, do_unlink=False)](https://www.blender.org/api/blender_python_api_2_78_release/bpy.types.BlendDataScenes.html#bpy.types.BlendDataScenes).

This method is used for the BDX profiler and so it will cause an error in 2.78 when the profiler is running for the first time. I'm searching for a good solution, that will work with both versions.

Perhaps it'll be worth to report this to the Blender tracker because it's probably a mistake to set `unlink=false` by default. It seems that previous versions had `unlink=true` by default.

SolarLune:
Hmm, this is interesting and strange. What would removing a scene without unlinking it do, then?

GoranM:
> Perhaps it'll be worth to report this to the Blender tracker because it's probably a mistake to set unlink=false by default. It seems that previous versions had unlink=true by default.

I doubt you'll be able to find anyone who cares enough to restore the old default (even though that would be the right thing to do).

> Hmm, would it be possible to add a line to check the Blender version and call remove(scene, true) if the version's larger than 2.78?

Yes: `bpy.data.version`

rafcolson:
>I doubt you'll be able to find anyone who cares enough to restore the old default (even though that would be the right thing to do).

It appears that a sacrifice had to be made. If it wasn't for removing scenes, the default would be changed for removing something else, I assume. Anyway, you can read about it in [this Task](https://developer.blender.org/T49579).

I'm trying @SolarLune 's suggestion as soon as I can build current master.

GoranM:
Ah, it was a wider API consistency thing.

That actually makes sense.

rafcolson:
PR submitted.

rafcolson:
Thanks for merging.

I've submitted another PR for fixing the same issue with *Making internal java files external*.

We also have to investigate if this API change will also influence removing the extra created meshes when applying modifiers in [exporter.py](https://github.com/GoranM/bdx/blob/ad685883ff675776fa35567b61b104e04eba63a4/blender/bdx/exporter.py#L134).

rafcolson:
[...]

PR #794, opened 2016-10-07
(#790) Since 2.78 removing scenes requires extra arg do_unlink=True

rafcolson:
Tested with 2.77 and 2.78.

### DIFF
--- a/blender/bdx/ops/exprun.py
+++ b/blender/bdx/ops/exprun.py
@@ -59,7 +59,11 @@ class BdxExpRun(bpy.types.Operator):
         if prof_scene_export:
         
             # Remove temporal profiler scene:
-            bpy.data.scenes.remove(bpy.data.scenes[prof_scene_name])
+            version = float("{}.{}".format(*bpy.app.version))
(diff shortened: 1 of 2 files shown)


GoranM:
You could just make a float:

```python
version = float("{}.{}".format(*bpy.data.version))
if version > 2.77:
...
```

rafcolson:
Thanks.

I think it would be wise to check `if version >= 2.78`, because there might be builds between `2.77` and `2.78`. I'm not sure about it, but we should consider it because the Blender API is talking about [a tuple of 3 numbers. eg. (2, 50, 11)](https://www.blender.org/api/blender_python_api_2_78_release/bpy.app.html?highlight=bpy.app.version#bpy.app.version)

So you know, `bpy.data.version` will return the version of the blend file, not of Blender itself, like `bpy.app.version` does.

PR updated.

GoranM:
> So you know, `bpy.data.version` will return the version of the blend file, not of Blender itself, like `bpy.app.version` does.

Oh, that's right. My mistake.

Thanks for the info.

Issue #792, opened 2016-10-05
SpriteAnim.Animation.reset() Not Working Correctly


SolarLune:
Similarly to #791, the `reset()` function doesn't seem to be working correctly for Animations. This seems to be because the `PlayDir` property only updates when the component updates, meaning that if you call `play()` or `reset()` and alter the playback speed, the SpriteAnim will suddenly end the animation, heading in the opposite direction.

Here's an example blend file. Notice that pressing A to play a oneshot animation doesn't work - the SpriteAnim "slams" into the other end of the animation, rather than simply playing backwards. 

https://www.dropbox.com/s/kcxxo12ytf4e7l5/SpriteAnimOneshotProblem.blend?dl=0

It might also be a good idea to make it so that the `play()` function has no effect on non-looping Animations unless you manually call `Animation.reset()` or switch to another Animation first.

SolarLune:
As another example, here's something showing what I have to do in Gearend to correctly animate a block that can be activated and de-activated using a single animation played bidirectionally:

```java

public void on(boolean setOn) {
        
    if (sa != null) {
            
        sa.play("open");
    
        if (setOn) {
            sa.speed = -1;
            sa.frame(sa.active.size() - 1);
        }
        else {
            sa.speed = 1;
            sa.frame(0);
        }
            
    }
        
    super.on(setOn);
}
```

I shouldn't have to manually set the frames here to the beginning or end - the component should correctly reset in the `reset()` function, called in the `play()` function (I believe).

Note that what I propose would probably only work if the speed were set before `play()` were called, but that's OK. Perhaps a direction attribute could be added to the `play()` function, to make it clearer that the order is important?

GoranM:
> This seems to be because the PlayDir property only updates when the component updates

Well, it's supposed to update when the component updates, because the component is (and should be) responsible for managing the animation process. 

The core of the problem is in how the animation is actually updated, which depends on how the playHead is updated, which involves the playDir.

So, start by inspecting that code.

SolarLune:
OK, I think this should fix it. 

PR submitted.

rafcolson:
What do you guys think of revisiting the code of `SpriteAnim.java` in general. It seems there are also other issues related to the direction of the play head. For instance, getting the current frame will give incorrect results if I just changed direction (theoretically).
```java
sa.speed = 1;
int currentFrame = frame(); // if this is 3
sa.speed = -1;
int currentFrame = frame(); // then this will be 5
```
Because:
```java
[...]

PR #796, opened 2016-10-07
(#792) - Fix playDir not being set on play() call.

SolarLune:
\+ `play(animName, playSpeed)` function to assist in this.

Without setting speed before calling `play()`, this won't fix the issue (which is why the new `play(animName, playSpeed)` function is suggested, so people won't have to worry about the ordering).

### DIFF
--- a/src/com/nilunder/bdx/components/SpriteAnim.java
+++ b/src/com/nilunder/bdx/components/SpriteAnim.java
@@ -151,11 +151,13 @@ public class SpriteAnim extends Component<GameObject> {
 
 		if (active != next){
 			active = next;
+			active.playDir = speed * active.fps < 0 ? -1 : 1;
 			active.reset();
 			ticker.done(true); // immediate play
 		}
 
 		if (!active.looping && active.onLastFrame()){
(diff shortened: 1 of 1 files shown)


GoranM:
I don't really understand how using this new play method would assist the user.

Setting speed first seems no more difficult than adding speed as a second argument.

SolarLune:
Because setting the speed afterwards doesn't have an effect, I figure the function would be a better solution as it handles the ordering, and so has a consistent result.

GoranM:
> Because setting the speed afterwards doesn't have an effect

Well, it has an effect, just not the one you want for your particular case.

> I figure the function would be a better solution as it handles the ordering

The user calls the function, knowing that it needs to be called in specific circumstances, because in those circumstances, the order matters, and this new play method will set speed first.

So, the user is still handling the ordering, just with an additional level of indirection. The indirection would be useful if the operation was more involved, and calling a function was far easier, but that's not the case here.

SolarLune:
The idea was that because the function exists, the user would just use that rather than setting the speed themselves. It's a bit of a "hint", of sorts.

SolarLune:
Well, never mind. PR updated to remove the extra play function.

PR #811, opened 2016-10-17
Bug-fix: Fallback to Point lamp if Hemi or Area are used.

SolarLune:
Better than crashing out of nowhere.


GoranM:
I think it would be best if the exporter provides data as specified in Blender, and we write code in the BDX importer that will interpret it properly.

SolarLune:
OK, PR updated to simply set the type to point in BDX if it's not sun or spot.

Issue #809, opened 2016-10-14
Suggestion: GameObject.join(list) & GameObject.join(map)


rafcolson:
Continuing from our discussion in #786:

To be removed:
- `GameObject.joinedMeshObjects`
- `GameObject.updateJoinedMesh()`

To be added:
- `GameObject.join(HashMap<Mesh, ArrayList<Matrix4f>>> map, boolean include=false)`
- `GameObject.join(ArrayList<GameObject>> list, boolean include=false)`

The method on `GameObject` then would also preserve the transform of that object.

Why "join" and not "add"?

Because "adding" implies that the initial mesh is still a part of the whole. We can add an optional argument to include the initial mesh though.

GoranM:
I'm not really sure what `Mesh.join` actually does, given the arguments.

Also, if there's no `joinedMeshObjects` list, and no method that updates the mesh based on the current state of those objects, how are you going to update the mesh?

SolarLune:
> Because "adding" implies that the initial mesh is still a part of the whole. We can add an optional argument to include the initial mesh though.

"Join" also, to me, implies that the initial mesh is part of the whole - you're joining the specified object to the calling one. 

I guess you're thinking like, "Join these meshes together in me, the calling GameObject"? I think of it like, "Add (join) these meshes _to_ me, the calling GameObject".
__________

 Similar to Goran, could you explain how meshes are joined under this suggested format?

rafcolson:
Similar to Goran, could you explain how meshes are joined under this suggested format?

rafcolson:
>I guess you're thinking like, "Join these meshes together in me, the calling GameObject"? I think of it like, "Add (join) these meshes to me, the calling GameObject".

A more descriptive name would probably be `Mesh.replaceJoined()`. But I'm open to `Mesh.add()` but then the extra argument would have to be `exclude=false` by default.

SolarLune:
> Mesh.join() takes a map of lists of transforms per mesh. (I replaced String with Mesh to be able to join unique meshes too). It does the same thing what GameObject.updateJoinedMesh() does, except for updating things related to the GameObject, such as the transform and physics.

Cool, OK, I think I get it. 

I think it would be easier to have the `join()` function that takes GameObjects to be on the Mesh class as well. It does the same thing as the Mesh's `join()` function, it just takes into account the meshes' GameObject transforms. Putting it on GameObject would also mean that people have to look to a different place in the API to see what it does.

[...]

PR #816, opened 2016-10-18
(#809) GameObject.join(list) & GameObject.join(map)

rafcolson:
Deprecates:
- `GameObject.joinedMeshObjects`
- `GameObject.updateJoinedMesh()`

Notes:

- "map" can contain scaled transforms as well as unscaled transforms (with zero m30, m31, m32)
- unique meshes/materials are now supported
- only another instance of the object on which `join(list)` is called, can be added to the list


rafcolson:
PR updated.

Issue #812, opened 2016-10-17
Question: about GameObject.transform()


rafcolson:
I'm always forgetting that the transform doesn't include the scaling of the object. But I wonder, why is it like so? I noticed there are no methods to set or get a Vector3f scale from Matrix4f, only a float. I that the reason?

Edit: I also noticed that jbullet's Transform doesn't support scaling.

GoranM:
Bullet requires the transform to be of unit scale, and it will actually enforce that; if you pass in a scaled transform, the physics simulation will make it unit scale, which wipes out whatever scaling was there previously.

So that's why scale is tracked separately.

rafcolson:
That's unfortunate. But couldn't we return a scaled transform anyway? And when applying a scaled transform do what is done now, except also apply the scale by extracting it from the transform? Something like this:

```java
	public Matrix4f transform(){
		Transform t = new Transform();
		body.getWorldTransform(t);

		Vector3f v = new Vector3f();
		for (int i = 0; i < 3; ++i){
		    t.basis.getColumn(i, v);
		    v.normalize();
		    t.basis.setColumn(i, v);
		}
		
		Matrix4f m = new Matrix4f();
		t.getMatrix(m);
		
		// add scale
		m.setRow(3, scale.x, scale.y, scale.z, 0);
		
		return m;
	}
	
	public void transform(Matrix4f mat, boolean updateLocal){
		activate();
		
		// apply scale
		scale(mat.m30, mat.m31, mat.m33);
		
		Transform t = new Transform();
		t.set(mat);
		
		Vector3f v = new Vector3f();
		for (int i = 0; i < 3; ++i){
		    t.basis.getColumn(i, v);
		    v.normalize();
		    t.basis.setColumn(i, v);
		}

		body.setWorldTransform(t);

		// required for static objects:
		body.getMotionState().setWorldTransform(t);
		if (body.isInWorld() && body.isStaticOrKinematicObject()){
			scene.world.updateSingleAabb(body);
			for (GameObject g : touchingObjects)
				g.activate();
		}
		//
		
		updateChildTransforms();

		if (parent != null && updateLocal){
			updateLocalTransform();
		}
	}

```
However, transforming gets more expensive because scaling is, so it would probably not be a good idea.

GoranM:
It may be best to have our own custom transform (a Matrix4f on GameObject), and have everything else operate based on that, instead of operating directly on the bullet body transform.

[...]

PR #824, opened 2016-10-27
(#812) GameObject.transform() now scaled GameObject.transform

rafcolson:
As good as finished. I need to test some scenarios to be totally sure everything's working as it should. In the meantime you can take a look. I fully documented the changes. I think it should be easier to check the code that way.


GoranM:
The "curr" prefix seems unnecessary, as does changing "localTransform" to "transformLocal".

rafcolson:
PR updated.

rafcolson:
PR updated. Improved calculation method for `GameObject.join()` by using matrix functions (faster and minimal drain on garbage collection).

GoranM:
As far as I can tell, this looks fine, but since it's a fairly large change, I think it would be best for @SolarLune to give it a quick test, just to be sure.

If everything seems to work on his end, I'll merge it in.

SolarLune:
OK, so there's a couple of problems.

1) This throws stuff out of whack for me physics-wise for joined objects, and 
2) Children of children aren't updated (so a child of a child of a dynamic object stays completely still).

Here's a blend file that demonstrates the issues.

https://www.dropbox.com/s/xfgho0sbi5nbq0s/physicsissues.blend?dl=0

You can see the first issue by the ground objects - although the ground objects are merged together and the joined physics body is updated, the ground panels are physically displaced and scaled down. For the second issue, you can see that Plane B is parented to Plane A, which is parented to Guy, but Plane B doesn't move at all.

_____

Also, ant gave me a warning that using "_" as a variable name might not be supported after Java 8 - might want to change that to something else.

rafcolson:
>Oh, and thanks for waiting.

Thanks for your review.

> 1) This throws stuff out of whack for me physics-wise for joined objects

I forgot to update the motion state and transform of the body.

>2) Children of children aren't updated

I forgot that I should be updating `childrenRecursive()` of parents with dynamics and active bodies.

Previously `Scene.updateChildBodies()` would update all parents with active bodies, not just those with dynamics. But with the current implementation no extra updates for `NO_COLLISION`, `STATIC` and `SENSOR` are needed, which is better for performance.

After some further investigation I found it to be better to always get the center of mass offset transform, so in case that the `bodyType` is changed, it wouldn't be necessary to replace the motion state, which wasn't happening right now.

So joining objects now also works with `CONVEX_HULL` `RIGID_BODY`'s.

I also changed the initialization method of instances so that the scale of the master object is multiplied by the scale of the group instance object (#827).

I've pushed the changes first, so it's easier to evaluate.

rafcolson:
PR updated but still have to check something. I'll keep you posted. And @SolarLune, the issues are solved, right?

rafcolson:
PR update and good to go if @SolarLune approves of it.

SolarLune:
OK, so that does fix the first one. As for the second issue, it's better, but still broken - Plane B is parented to Plane A in the demo, but it doesn't follow Plane A should Plane A move. The fix simply keeps the Planes' offsets to the parent (Cube).

rafcolson:
>The fix simply keeps the Planes' offsets to the parent (Cube).

Yes, of course, only those with `parent == null` will update their children. I think I'll be returning to the original code for now, and try to optimize it later.

rafcolson:
PR updated. I'm sorry for the trouble. Had to know better. @SolarLune if you would like to check one more time? Thanks a lot.

SolarLune:
OK, that solved the last issue, but I notice that scaling something to 0 makes it "unrecoverable" from this state (you can't scale it back afterwards).

rafcolson:
>but I notice that scaling something to 0 makes it "unrecoverable"

We could scale with a low number instead, but I wonder if there are even cases where objects should have zero scale.

Up until now BDX didn't support objects with zero scale, or did it? I didn't take this into account before. Try running after scaling Sacky on the Z axis. A `ZeroDivisionError` will occur while exporting object dimensions.

SolarLune:
> We could scale with a low number instead, but I wonder if there are even cases where objects should have zero scale.

My use case is an energy gauge - it starts off at 1.0 scale, but goes to 0 after using energy. After that, it can't grow under this PR.

> Up until now BDX didn't support objects with zero scale, did it? I didn't take this into account before. Try running after scaling Sacky on the Z axis. A ZeroDivisionError will occur while exporting object dimensions.

I guess I've never tried exporting objects with 0 scale.

rafcolson:
>My use case is an energy gauge - it starts off at 1.0 scale, but goes to 0 after using energy. After that, it can't grow under this PR.

I see. So what do you suggest? I don't think there's any other way than using a small number instead.

SolarLune:
> I see. So what do you suggest? I don't think there's any other way than using a small number instead.

I don't really have any other suggestions - using a small number sounds like the only solution. You can cap the scale set to a minimum value if scale's greater than 0 and a maximum of -(that value) if scale is negative. Note that you can't use Float.MIN_VALUE - looks like it rounds it down to 0 for some reason, so you'll need to use an arbitrary number. 

As long as we explain why the number can't ever equal 0 and why the developer can't add 0.0000001 to the scale of an object, then it should be fine, haha.

rafcolson:
PR updated. See changes in: `Matrix4f.scale(Vector3f v)`. `MIN_VALUE = 0.00001f`.

SolarLune:
Great, that's fixed. 

I found another bug - when you resize group instances, the object's physics mesh seems to be resized by double, and the object becomes intangible.

In this demo, the planes are the same (just group instances of an existing "Ground" object), but the right one is intangible. Also notice that the physics mesh for the right plane doesn't cover it, but extends a bit further out.

https://www.dropbox.com/s/25cjrna78j6z1wh/physicsissues_2.blend?dl=0

Nice work so far, @rafcolson.

rafcolson:
>Nice work so far

Well, I didn't expect those bugs, so it's not smooth. Looked at the code a couple of times, tried things, but can't find the source of that last bug. I'll have to redo this commit, so it'll take some time. So you guys know.

GoranM:
> I also changed the initialization method of instances so that the scale of the master object is multiplied by the scale of the group instance object (#827).

Maybe that has something to do with the latest bug?

In general, I think it would be best to first replicate previous functionality.

Further changes/additions should probably be in a different PR.

rafcolson:
>Maybe that has something to do with the latest bug?

I'm not sure. I remember that reverting this change didn't solve it. In any case, the issue appears with cloning. The physics body size is doubled, also when adding objects to the scene.

>In general, I think it would be best to first replicate previous functionality. [...]

Yes. That would be best.

I'll take some time this week to continue.

rafcolson:
Six months already... I think I am ready to continue.

GoranM:
Yeah, it goes fast. Glad to know you're still with us.

rafcolson:
So this PR is updated, finally. I couldn't find any issues, but I'd appreciate it if you could test it out once more, @SolarLune.

rafcolson:
@SolarLune did you find any problems yet?

SolarLune:
I can't test it with Gearend - you have to rebase off of master.

rafcolson:
>Hey, sorry for the late response. I can't test it with Gearend - you have to rebase off of master / bring it up to the latest commit.

There are no conflicts with the base branch, are there? Or do you mean to rebase off of your repository?

SolarLune:
OK, sorry, it was an issue on my end.

So, it seems like logic usage is way heavier in Gearend with this commit. 

With this commit, the game runs at ~47 FPS. The highest profiler category is Logic, and it's at ~70%.

Without it, the game runs at 60 FPS. The highest profiler category is Outside, and it's at ~40%. Second highest are Render and Logic, which are both around 20%. 

Also, the BDX profiler doesn't show up anymore for some reason.

Besides that, I believe the game appeared to play the same.
_____

Running Visual VM showed me that about half of the execution time is spent updating transforms from setting position - here's a couple of profiler snapshots I exported from it. One is before the transformation change, and the other is afterwards. 

[GearendTransformChangeVisualVMSnapshots.zip](https://github.com/GoranM/bdx/files/1050715/GearendTransformChangeVisualVMSnapshots.zip)

Notice that after the Transform change, the heaviest function is the LockMotion component for its GameObject.position() calls, and accounts for 33% of the execution time. Before the change, that function accounts for 0.6% of the execution time.

rafcolson:
>Also, the BDX profiler doesn't show up anymore for some reason.

The new commit doesn't effect the profiler on my end. Anything in the console?

>Notice that after the Transform change, the heaviest function is my LockMotion component for its GameObject.position() calls, and accounts for 33% of the execution time. Before the change, that component's main function call accounts for 0.6% of the execution time.

Didn't expect these bad results. It's quite strange because as far as I know the new transformation method isn't that different from the old method. I'll have to look into this.

>If you need me to make a test, I can. From this, though, it just seems like the transform / updateTransform() / updateChildTransforms() functions are slower - it probably would be as easy as just making a ton of objects that change their positions to be able to easily test it.

No, that's not necessary. I'll be doing some tests on my own. But thanks to this information I know in which area I should do those. Thanks a lot.

rafcolson:
Added an Empty to the second layer and changed `Sacky.java`:
```java
package com.raco.test;

import java.util.ArrayList;

import com.nilunder.bdx.*;
import com.nilunder.bdx.utils.Random;

	ArrayList<GameObject> l;
	
	public void init(){
		l = new ArrayList<GameObject>();
		for (int i = 0; i < 4000; i++){
			l.add(scene.add("Empty"));
		}
	}

	public void main(){
		for (GameObject g : l){
			g.position(Random.vector());
		}
	}
	
}
```
4000 objects is no realistic amount. Can't imagine that Gearend would need so much. But there's a considerable difference. On my end current master shows logic at around 25%, whereas the repo shows around 35%. That's an increase of 10%. I'm wondering if the 50% increase (20%-70%) in Gearend is only due to setting positions... I'll look into this further.

rafcolson:
So I've found the problem. My branch also contained a fix on updating AaBb of NO_COLLISION objects when updating the transform, which was slowing things down.

Also, there's no reason to update the modelInstance if no scaling is done. Also the transforms of the body and motionState doesn't need updating when only scaling is done. After making these changes I'm getting these results:

| Operation | Before | After | Difference |
| --- | ---: | ---: | ---: |
| *position* | 21% | 16% | -5% |
| *rotation* | 67% | 26% | -41% |
| *scaling* | 28% | 17% | -11% |

Seems even better!

There might also be done improvements with reducing the updates of the visuals in `Scene.updateVisuals()`. Visuals only need updating when performing transformations (happens in `GameObject.updateTransform()` and `Scene.updateTransforms()`). Instead of updating all visuals, relevant visuals could be collected and updated. But this will be a separate suggestion.

rafcolson:
@SolarLune I'd appreciate it if you could do another test to see if the issue is solved. Let me know if it can be merged. Thanks.

SolarLune:
OK, this seems pretty much just as performant as otherwise. Nice work!

It seems like making the profiler visible or invisible in-game makes the Outside bar disappear / reappear. Not sure of the cause, exactly - here's a test case.

[TransformTest.zip](https://github.com/GoranM/bdx/files/1063270/TransformTest.zip)

rafcolson:
The old method is to get the transforms of the visuals in [`Scene.updateVisuals()`](https://github.com/GoranM/bdx/pull/824/commits/6735e1babfc56b4b76995a820002f2c6cb48c0f9#diff-343d1f1fde98dfa03b3b102030c3d4e5L814) by first getting the scale with `modelInstance.transform.getScale(Vector3 scale)` but [`Matrix4.getScale(Vector3 scale)`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Matrix4.java#L1073) doesn't support negative scaling. Consequently, for visible objects, even when the old method for scaling an object seems to support negative scaling (although [partially](https://github.com/GoranM/bdx/pull/824/commits/6735e1babfc56b4b76995a820002f2c6cb48c0f9#diff-49b2fa87135c8e3035fc454ff6ceb44eL645), because of [`Matrix4.scl(float x, float y, float z)`](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Matrix4.java#L1073)), this is canceled out by `Scene.updateVisuals()`.

This explains why removing support for negative scaling in `Matrix4f.scale(Vector3f)` fixes this issue with the Profiler's `"outside"` subsystem. Remember the changes after [this comment](https://github.com/GoranM/bdx/pull/824#issuecomment-257472824) implied support for negative scaling, but it doesn't matter for `GameObject.modelInstance` or `GameObject.body`. **We could store a field `GameObject.scale` to be able to preserve negative numbers**. But then the question remains, is this really necessary?

Because, with current master, negative scaling is not exported. So when you scale Sacky to `(-1, -1, -1)` in the Properties shelf in Blender, `Sacky.scale()` will return `(1, 1, 1)`. But if you would scale Sacky in game with `Sacky.scale(-1, -1, -1)` it will return `(-1, -1, -1)`. It doesn't seem consistent whilst with the new transformation method it is consistent (negative values are ignored always), so what should I do @GoranM and @SolarLune ?

rafcolson:
@SolarLune 
@GoranM 
Any thoughts on how to go forward here?

SolarLune:
Hmm, I dunno. 

> But then the question remains, is this really necessary?

Yeah, I think so. Ideally, negative scale values should be fully supported in BDX and Blender. 

Sometimes the developer might want to scale an object to negative values (to flip a 2d billboarded sprite, for example), so I think it best that we do what we can to support this. If there's any alternative ideas that get the job done without an extra field, that would be fine, too, though.

GoranM:
I agree with SolarLune. Negative scaling should be supported, and I don't really understand why it's troublesome to provide that support. As far as I know, Matrix4f is perfectly capable of storing values that represent an inverse scale transform, as is libgdx's Matrix4.

> it seems that in my previous post I can't leave a message to `@GoranM`. Strange

I think it's a case sensitive system, so you have to use `@GoranM`, not `@goranm`

rafcolson:
>I agree with SolarLune. Negative scaling should be supported, and I don't really understand why it's troublesome to provide that support. As far as I know, Matrix4f is perfectly capable of storing values that represent an inverse scale transform,

Ideally the transform matrix of the GameObject would also support negative scaling, but I just don't have a clue on how this could work with `Matrix4f`. For one, it surely isn't intended to be used with negative scaling, as `Matrix4f.getScale()` [will not return any negative values](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Matrix4.java#L1142).

I'm pretty sure it's not just a matter of getting the signs of the axis values. Could you maybe give me some pointers?

> as is libgdx's Matrix4.

I might not understand a thing or two about how negative scaling can be supported by `Matrix4`, but as far as I can understand I don't think it currently is, for the [aforementioned reason](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/math/Matrix4.java#L1144).

Would it be okay to just store the scale in a private variable `GameObject.scale` and update this value whenever the scale is changed? `Matrix4f` wouldn't support negative scaling but at least `GameObject` would.

GoranM:
The matrix itself should be able to hold arbitrary data. The scaling restrictions seem to be enforced solely by the scale methods, but you don't have to use those.

Can't you just make and use new scale methods, such that they won't eliminate relevant scaling information?

rafcolson:
>Can't you just make and use new scale methods, such that they won't eliminate relevant scaling information?

Well, I would like to but I don't know how, because [_negative_ components in a matrix show up naturally as a result of rotations](https://gamedev.stackexchange.com/questions/54505/negative-scale-in-matrix-4x4/54508#54508). So how would I then be able to store negative scale values without losing those after doing operations to change rotation?

Example:
```java
private void test(){
	
	// construct and print an identity matrix with negative scaling
	
	Vector3f pos = new Vector3f();
	Vector3f sca = new Vector3f(-1, -1, -1);
	Matrix3f ori = Matrix3f.identity();
	Matrix4f m = Matrix4f.transform(pos, sca, ori);
	
	System.out.println(m);
	
	/** result:
	 * -1.0, -0.0, -0.0, 0.0
	 * -0.0, -1.0, -0.0, 0.0
	 * -0.0, -0.0, -1.0, 0.0
	 * 0.0, 0.0, 0.0, 1.0
	 */
	
	
	// rotate this matrix 360 degrees on the X-axis
	
	ori.rotX((float) Math.PI);
	ori.rotX((float) Math.PI);
	m = Matrix4f.transform(pos, sca, ori);
	
	System.out.println(m);
	
	/** result:
	 * -1.0, -0.0, -0.0, 0.0
	 * -0.0, 1.0, -8.742278E-8, 0.0
	 * -0.0, 8.742278E-8, 1.0, 0.0
	 * 0.0, 0.0, 0.0, 1.0
	 */
}
```
The `Matrix3f.rotX()` method produces a negative value (sine) for [`m12`](https://github.com/hharrison/vecmath/blob/master/src/javax/vecmath/Matrix3f.java#L1406), as does `Matrix3f.rotY()` for [`m20`](https://github.com/hharrison/vecmath/blob/master/src/javax/vecmath/Matrix3f.java#L1433), and  `Matrix3f.rotZ()` for [`m01`](https://github.com/hharrison/vecmath/blob/master/src/javax/vecmath/Matrix3f.java#L1451).

The same with BDX's custom method `Matrix3f.rotation(axis, angle)` due to [`Matrix3f.set(AxisAngle4f)`](https://github.com/hharrison/vecmath/blob/master/src/javax/vecmath/Matrix3f.java#L878).

rafcolson:
Even Python's `Mathutils.Matrix` is not able to preserve negative scaling. The [API](https://docs.blender.org/api/current/mathutils.html?highlight=to_scale#mathutils.Matrix.to_scale) notes:

> This method does not return negative a scale on any axis **because it is not possible to obtain this data from the matrix alone**.

Which seems to imply that Blender's method for preserving negative scaling also involves keeping a separate variable for scale.

rafcolson:
PR updated.

The flipping feature provides compatibility for the Test game tutorial (99 bottles). Scaling a GameObject with a negative scale will still work as before, except it is now only applied to the modelInstance (which is all we need, right?).

GoranM:
Oh, wait, I misread, and thought that you were talking about not being able to set a negative scale.

It's true that you can't get signed scale info from the matrix, but you don't need it, because the ability to set a negative scale is what provides the flip utility that SolarLune wants (if I understood his argument correctly).

I would agree with you that "flip" methods would probably be more sensible, because setting a negative scale, without preserving sign info for the subsequent gets, would probably confuse people.

I don't know if @SolarLune would be ok with that, or if he would prefer existing scale semantics.

I'm ok with whatever you guys decide.

rafcolson:
>Oh, wait, I misread, and thought that you were talking about not being able to set a negative scale.

My fault. My initial description of the problem wasn't complete.

>I would agree with you that "flip" methods would probably be more sensible, because setting a negative scale, without preserving sign info for the subsequent gets, would probably confuse people.

My thought exactly. Thanks a lot for reviewing this @GoranM. I'll be waiting for @SolarLune to review this hopefully for the last time.

rafcolson:
Just in case you missed it @SolarLune, the PR was updated. Let us know if you think this can be merged. Thanks again.

SolarLune:
Hey, sorry - I'll check it out today.

SolarLune:
After the last font writer PR is merged, I'll review this one, as it depends on that change for me to fully test this.

rafcolson:
These commits are based on those other commits. So you'ld know.

rafcolson:
PR rebased.

src/com/nilunder/bdx/Scene.java
```diff
@@ -301,7 +301,9 @@ public void init(){
 			}
 
 			com.badlogic.gdx.graphics.Mesh mesh = g.modelInstance.model.meshes.first();
-			float[] trans = gobj.get("transform").asFloatArray();
+			g.flipState(new Vector3f(gobj.get("scale").asFloatArray()));
```
SolarLune: This errors for me if you press P from the 3D view to play the game from Blender's camera. I'm guessing the "default" camera in Blender doesn't have a specific scale associated with it (because there's no object representing it). 

You could default out to just 1,1,1 if there's no scale for this situation, or just provide an exported scale for the default camera.

SolarLune:
Tested again and found an error - commented on the line that causes it.

rafcolson:
I can't seem to find your in-line comment.

SolarLune:
Oh, I guess I have to submit the review fully. That should do it.

rafcolson:
Yes, it seems so. I completely forgot about the 3d view camera. Thanks for tackling this. PR updated. Take your time if you're not sure yet that this can be merged. Thanks so much for reviewing this so many times.

SolarLune:
Nah, no problem.

So I tested again, and the PR seems like it's almost there - the main issue I noticed is scaling while adjusting the direction of an object using alignAxisToVect. Here's a simple test file - press space to scale the pointing arrow. When you do, it goes out of whack.

[RafTransformBranchTest.zip](https://github.com/GoranM/bdx/files/1164838/RafTransformBranchTest.zip)

rafcolson:
So, it turns out that I've been making a crucial mistake with the composition of the `GameObject.transform` matrix. It had an absolute scale applied to it, whilst it should be a rotated scale. I'm working on a solution.

rafcolson:
@SolarLune 
This should be working now, but I'm going to test this thoroughly before passing the final testing to you.

src/com/nilunder/bdx/Scene.java
rafcolson: @SolarLune It's weird that your comment is still present, even though I've updated the branch...

rafcolson:
Update on performance (logic).

| Operation | Before | After | Difference |
| --- | ---: | ---: | ---: |
| *position* | 29% | 21% | -8% |
| *rotation* | 50% | 33% | -27% |
| *scaling* | 35% | 23% | -12% |

So overall this PR should boost performance a bit.

@GoranM 

The last commit contains a couple of additional functions to help improve matrix operations. To just change the positional component of the matrix no matrix multiplication is necessary. Initially there are methods for rotation, such as `Matrix4f.rotX(float angle)`, but these will "set" the rotational component instead of "rotating" it. So `Matrix4f.rotate(Vector3f axis, float angle)` seems useful. For more information, see commit.

@SolarLune 

Here's a [folder containing the test blends for this PR](https://www.dropbox.com/sh/p2709liuerhvnkq/AAAfxjPGA_GtGtohmyd7TN9Fa?dl=0). Final tests are showing consistent results on all transform-like operations. If there are any other issues left, I'm happy to hear. Thanks.

rafcolson:
@SolarLune 
PR rebased.

rafcolson:
@SolarLune
Did you perhaps found the time to do some final testing? In my previous post I've also put a link to a folder with test blends, which may be helpful. Thanks again.

SolarLune:
Hey, I'm super sorry for the late response. 

So I just tested it with Gearend, and things aren't looking so good, haha. It seems the scales of certain objects are way off. Some objects are way larger than they should be, while others are much smaller.

screenshot from 2017-09-20 11 32 39

The background doesn't appear visible, and I feel like it might be because it's way larger than it should be.

Anyway, I'm not certain if this is the only issue, but it seems like scale isn't being properly exported from Blender. It seems to be set in code fine (i.e. setting a cube to scale 2 in Blender makes it something like scaled by 8x, whereas setting the scale to 2 while running truly sets it to 2). Here's a quick test blend (it's just a cube scaled up by 2): https://www.dropbox.com/s/2kexx3pnp7z1hjo/TransformTest2.blend?dl=0

If you want, Raco, I can also send you a copy of a piece of a scene from Gearend to help you debug.

rafcolson:
@SolarLune PR rebased.

SolarLune:
OK, cool. I'll check it out later today.

rafcolson:
@SolarLune
Any chance you have some time next week? It should be okay. After this I'll submit a

rafcolson:
@GoranM 
Perhaps you are able to review this?

SolarLune:
Hello! Sorry for not responding - I forgot to review this and have been having computer trouble. I'll look over it today.

SolarLune:
OK, I checked it out. 

Everything seems just about right with Gearend, which is great. The only issue I found is that it seems like you can't scale anything back after scaling it down to 0.

rafcolson:
PR updated. Changed `Matrix3f.mul(Vector3f v)` and `Matrix4f.mul(Vector3f v)` to ensure that the scaling component will never be zero:
```java
	public final void mul(Vector3f v){
		float MIN_VALUE = 0.00001f;
		float f;
		f = (float) (v.x < 0 ? Math.min(v.x, -MIN_VALUE) : Math.max(v.x, MIN_VALUE));
		m00 *= f; m10 *= f; m20 *= f;
		f = (float) (v.y < 0 ? Math.min(v.y, -MIN_VALUE) : Math.max(v.y, MIN_VALUE));
		m01 *= f; m11 *= f; m21 *= f;
		f = (float) (v.z < 0 ? Math.min(v.z, -MIN_VALUE) : Math.max(v.z, MIN_VALUE));
		m02 *= f; m12 *= f; m22 *= f;
	}
```
I didn't make changes to the static method `Matrix4f.transform(Vector3f pos, Vector3f sca, Matrix3f ori)` because it can be assumed that scale will never be zero initially, right?

SolarLune:
Welp, I tested this with Gearend, and I didn't see any issues this time around. It should be good to go. 

Nice work, @rafcolson!

rafcolson:
Thanks for all your efforts to test this over and over, @SolarLune

GoranM:
Just a few days shy of being open for a full year :)

rafcolson:
Yes. Finally! Thanks for your patience @GoranM.

PR #831, opened 2016-11-15
+Text alignment.

SolarLune:


SolarLune:
Updated PR with a bugfix to make it reliably align correctly regardless of capacity.

PR #834, opened 2016-11-17
-Bug-fix: Use active area for screenPositionNormalized.

SolarLune:
Otherwise, the position is normalized across the entirety of the game window, rather than just the gameplay area where the viewport can be.

To explain a bit further, imagine a game with the viewport set to letterbox. If you stretch the window in a direction, black areas will extend along the extended axis (which is fine). screenPositionNormalized() didn't take these black areas into account in its readings (so it would weight the readings towards the center of the screen regardless of the position of the vector "in the game world"). 

I'm not sure if this fix should be applied to `screenPosition()` as well.

### DIFF
--- a/src/com/nilunder/bdx/Text.java
+++ b/src/com/nilunder/bdx/Text.java
@@ -3,8 +3,6 @@ package com.nilunder.bdx;
 import com.badlogic.gdx.utils.JsonValue;
 import com.badlogic.gdx.graphics.*;
 
-import javax.vecmath.Matrix4f;
-
 public class Text extends GameObject{
 
 	public enum Alignment {
@@ -18,6 +16,7 @@ public class Text extends GameObject{
 
 	public JsonValue font;
 	public int capacity;
+	private float lineHeight = 1.0f;
 
 	public void text(String txt){
 		// Reform quads according to Angel Code font format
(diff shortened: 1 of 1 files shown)


GoranM:
In my view, everything should be viewport relative, but it's possible that @rafcolson had something else in mind ... ?

SolarLune:
Yeah, I think it should be relative to the viewport as well, but I'm fine with waiting to see what ideas @rafcolson has.

rafcolson:
If `Camera.screenPosition(Vector3f p)` derived from `com.badlogic.gdx.graphics.Camera.project(Vector3 worldCoords, float viewportX, float viewportY, float viewportWidth, float viewportHeight)` is supposed to return the screen coordinates of the Camera's Viewport, than yes. However, the API states that [the viewport is assumed to span the whole screen](https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/Camera.html#project-com.badlogic.gdx.math.Vector3-float-float-float-float-), so the coordinates are not viewport relative. You could do a simple test with the default scene (enable the Profiler).
```java
package com.raco.test;

import com.nilunder.bdx.*;

	public void init(){
		Bdx.profiler.subsystemsVisible(false);
		
		// transform the viewport
		scene.viewport.sizeNormalized(0.25f, 0.25f);
		scene.viewport.positionNormalized(0.75f, 0.75f);
	}
	
	public void main(){
		// print screen position of every object
		for (GameObject g : scene.objects){
			Bdx.profiler.props.put(g.name(), scene.camera.screenPosition(g.position()));
		}
	}
	
}
```
The position will be relative to `Bdx.display.size()`, not `Scene.viewport.size()`.

In any case, if `Camera.screenPositionNormalized()` should be relative to `Scene.viewport.size()`, than also `Camera.screenPosition()` should be.

>Yeah, I think it should be relative to the viewport as well

I think `screenPosition()` should be relative to the screen, not the viewport. Adding extra methods `viewportPosition()` and `viewportPositionNormalized()` would be better, I think. If at some point the screen position methods seem redundant, these could be removed. But for the moment I can't say these would be useless.

PR #844, opened 2016-12-06
+ Better responsiveness on export.

SolarLune:
- Set mouse cursor to wait on export.
- Print when scenes are exported.
This way, the export process doesn't go unnoticed in case you mis-press P (like I usually do).

### DIFF
--- a/blender/bdx/ops/exprun.py
+++ b/blender/bdx/ops/exprun.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import bpy
 import subprocess
 from .. import utils as ut
@@ -10,6 +11,10 @@ class BdxExpRun(bpy.types.Operator):
     bl_label = "Export and Run"
 
     def execute(self, context):
+
+        # Set the mouse cursor to "WAIT" as soon as exporting starts
+        context.window.cursor_set("WAIT")
+
         j = os.path.join
 
         proot = ut.project_root()
@@ -50,11 +55,15 @@ class BdxExpRun(bpy.types.Operator):
             os.mkdir(inst)
 
         # Export scenes:
-        for scene in bpy.data.scenes:
+        for i in range(len(bpy.data.scenes)):
+            scene = bpy.data.scenes[i]
             file_name =  scene.name + ".bdx"
             file_path = j(asset_dir, "scenes", file_name)
-
+            sys.stdout.write("\rBDX - Exporting Scene: {0} ({1}/{2})                            ".format(scene.name, i+1, len(bpy.data.scenes)))
+            sys.stdout.flush()
             bpy.ops.export_scene.bdx(filepath=file_path, scene_name=scene.name, exprun=True)
+
+        print("")       ## Added blank line for reading comfort
 
         if prof_scene_export:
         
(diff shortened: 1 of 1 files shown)


GoranM:
> This way, the export process doesn't go unnoticed in case you mis-press P (like I usually do).

Wouldn't the waiting mouse cursor indicate that there's an export in progress?

SolarLune:
The mouse cursor only shows up currently when the export process is finished going through all of the scenes - that's why I moved it up, to when the process first starts (when you first press P).

GoranM:
I meant now that you changed it to wait when the process first starts.

With that, you wouldn't need anything else to know that you're exporting.

SolarLune:
Yeah, but it'd still be nice to get some kind of a "percentage" so I can know how many scenes I've exported or have left to export (ergo the print statements). 

Maybe we could add a print statement that just prints a single line (i.e. `Exporting Scenes (1/7)`), which gets overwritten so it doesn't clog the console?

GoranM:
> so I can know how many scenes I've exported or have left to export

Why do you need to know that? You still have to wait until they're all exported, and your wait time is only a few seconds, without too much variation, so what's the point?

SolarLune:
> Why do you need to know that? You still have to wait until they're all exported, and your wait time is only a few seconds, without too much variation, so what's the point?

No particular reason, it's just nice to know how close it is to finishing. No different from a loading gauge being better than a static picture. It's for user experience.

Why does the fact that it takes "only a few seconds" matter? 

My game's kind of small - just a few scenes, and each scene has only a few thousand objects for the tiles (most of which are repeats), and yet it takes a good bit of time to export. For a moderately bigger game, it's easy to see that this would take much more time than just a few seconds.

GoranM:
> Why does the fact that it takes "only a few seconds" matter?

If it took several minutes, that may give you enough time to do some other work, and from that perspective, there may be some utility in being able to see how many scenes remain, because you can check back to see if it's about to be done, or if there's enough time to do some other work.

However, for a few seconds, watching a progress log/bar in the console doesn't really strike me as being better/nicer than just watching blender's wait cursor.

That said: If you really want it, I'll merge it in.

SolarLune:
OK, I updated the PR with a more elegant solution, I think. Let me know what you think of this solution.

Something that's nice about this is that it allows you to see which scenes are the heaviest to export, which can be useful to know in case something's slowing things down more than usual.

If you still don't think it's useful enough / enough utility to see how the export process goes, it's OK - I can live with just the mouse cursor changing earlier. Thanks.

GoranM:
> Something that's nice about this is that it allows you to see which scenes are the heaviest to export, which can be useful to know in case something's slowing things down more than usual.

It's only useful if the information can be used to remedy the situation. The fact that you know that some scenes are "heavy" doesn't give you a path to speed up the export process. You just know that certain scenes take longer to export than others, and you wait just as long as you would if you didn't have that knowledge.

So, I still don't see the utility, but, as I said, if you really want it, I will merge it.

SolarLune:
Hmm, I see, thanks. 

Well, I think I would still like it, if possible.

Issue #154, opened 2015-02-06
Suggestion: music and sound device?


SolarLune:
Hey, there. 

I think it'd be nice to have a static device for playing back music and sound to be able to easily control volume from it. Currently, when you play a sound or music, you can adjust the playback of the sound afterward. This can be a bit tedious if you have a lot of sounds and want to change something later. 

With the device, you could adjust it globally to more easily and consistently control volume and panning. The actual sound and music playback system would be basically the same. For example:

```
Bdx.audio.device.masterVolume = 0.5f;

Bdx.audio.device.musicVolume = 0.75f;

// Default volume of 100%, or 1.0 for soundVolume, musicVolume, and masterVolume

Music levelBGM = Bdx.audio.device.loadMusic("Level1BGM"); 

levelBGM.play(); // Plays back at a volume of 75% of 50%, or 37.5%

Sound jumpSound = Bdx.audio.device.loadSound("Jump"); // Volume of 50%.

jumpSound.play();

jumpSound.volume = 0.5f; // Jump sound is now playing at 25% volume
```

GoranM:
The ability to control overall volume would be useful, but I don't think the implementation would require a significant change in the API. You could just have:

```java
Bdx.music.volume
Bdx.sounds.volume;
Bdx.audioVolume;
```

BTW: `music` and `sounds` are designed to behave like hashmaps, so you can just use get("sound_name") - you're not supposed to call load directly.

As for the API, in general: I didn't want to add more levels of indirection than necessary, but, organizationally, I guess it would make more sense to have everything stem from Bdx.audio:

```java
Bdx.audio.music.volume
Bdx.audio.sounds.volume
Bdx.audio.volume
```

It's a bit longer to type:

```java
Bdx.audio.sounds.get("name").play();
// vs
Bdx.sounds.get("name").play();
```

I guess people could alias, if necessary:

```java
Music music = Bdx.audio.music;
music.get("name").play();
```

But then, there could be a clash between libgdx Music and bdx Music, so maybe we could rename bdx Music to Musics ... ?

Or maybe all this stuff should just be available directly from the GameObject itself?

All our logic runs on the GameObject to begin with, and we already have "scene" there, so why not other things?

SolarLune:
I like the idea of just putting that stuff in the music and sound stores. There's not many locally available options for music and sound in LibGDX (just panning, volume, and looping), so it wouldn't be difficult or too invasive to put it in there.

[...]

PR #849, opened 2017-01-10
AudioDevice for controlling playback of sounds and music.

SolarLune:


SolarLune:
Hey, here's an attempt at making a device class for audio playback that allows for tweaking volume, panning, and pitch from a global location. Let me know what you think.

GoranM:
I remember we discussed something along these lines in #154, but we didn't decide on a specific approach.

I think we should hash it out there before making a PR.

SolarLune:
That sounds fine, yeah.

SolarLune:
OK, PR updated.

GoranM:
The changes to GameObject.java (the added import statements) don't provide shortcuts to the relevant objects on Bdx. I'm pretty sure you would need to create the relevant fields on GameObject, and then set them to existing audio object on Bdx.

You could do that, or you could just leave Bdx.sounds and music to point at Bdx.audio.sounds and Bdx.audio.music.

I think the latter would probably be a sufficient shortcut, while not breaking the existing API.

---

I can see that you're tracking currently playing music streams, but I don't see the code that tracks currently playing sounds streams.

Did I just miss it, or was it omitted for some reason?

---

I don't think higher level effects, like "fading", belong in the engine. If someone wants to change volume over time, in some specific way, that's their responsibility.

SolarLune:
> The changes to GameObject.java (the added import statements) don't provide shortcuts to the relevant objects on Bdx. I'm pretty sure you would need to create the relevant fields on GameObject, and then set them to existing audio object on Bdx.

Yo, that's right - I was going to add shortcuts, but decided against it, as it added code for not that much of a payoff (in my mind). I missed it when I cleaned things up - my bad. PR updated to fix it.

> I can see that you're tracking currently playing music streams, but I don't see the code that tracks currently playing sounds streams.

> Did I just miss it, or was it omitted for some reason?

I would have tracked currently playing sound streams, but it's not possible to tell when a sound stream is finished playing. I could alter looping sound streams, since they won't stop unless the developer interjects, but already playing sound streams could be looped after initial play, or, conversely, looping sounds could be set to not loop after starting. So, I'd need to keep track of both playing and looping streams just to adjust looping sound streams. 

That was so much work and additional code that I just ended up omitting automatic live adjustment of sounds - sounds only alter their properties on first play, which isn't a huge deal (since you won't generally dynamically, in-game, alter global sound playback properties).

> I don't think higher level effects, like "fading", belong in the engine. If someone wants to change volume over time, in some specific way, that's their responsibility.

I would generally agree for high level effects, but fading a BGM in and out is fairly common in games and extremely useful, so I feel like it's a key usability feature.

GoranM:
There is also a fairly common need for sprite animation, but we don't encode that feature directly into the engine core, because sprite animation is simply the manipulation of texture coordinates, and that makes it a higher level feature that can be encoded in a component.

We ship a default implementation of that component not just because there's a fairly common need for the feature, but also because the implementation is not trivial, and therefore not something that people could easily write on their own.

Fade, on the other hand, is simply manipulating a single value (volume) over time, and typically in a linear fashion, so I think one would have to stretch to justify inclusion even as a component, let alone engine code.

SolarLune:
> the implementation is not trivial, and therefore not something that people could easily write on their own.

I see what you mean, but we have other components that are fairly trivial to write, like the Halo and Track components. The component and state classes themselves are also very simple. I don't think "it's hard to write on your own" should be the qualifier for whether a particular piece of code is implemented into the engine or not - I think it should be whether it would be useful to most users. 

I don't mind too much moving fading into a component, but I feel like it's useful enough and utilized in enough games to justify inclusion somehow.

GoranM:
Track and Halo mostly serve as familiar examples for how one could write basic components. They don't represent a baseline for inclusion.

The component and state classes are engine-level abstractions that provide the foundation for writing components, so I don't see how their simplicity relates to what we're discussing here.

I'm not arguing that implementation effort should be the only metric for component inclusion, but I think it should be a factor, because we don't want to clutter the engine with components that do overly trivial things (ie: incrementing/decrementing a value).

For now, I think it would be best if you make the fade component a part of your helper library. That way, if anyone wants to use it, it's available, and if there are more users, and most of them want to use it, we can make it a default that ships with the engine.

SolarLune:
Fair enough.

PR updated.

GoranM:
Instead of an update method that has to run in the main engine loop, on each frame, you could simply have setters that would update playing streams.

---

Considering that you're already looping through all elements in playingMusic, and removing those that are not currently playing, it seems redundant to also remove them in pause and stop methods.

SolarLune:
Ah, I missed those points. Fixed.

GoranM:
I don't think we need to run an update method on every frame.

The list of tracked streams doesn't have to be exclusively a list of currently playing streams, as long as those are included, so I think it's probably sufficient to remove non playing streams in the play method.

SolarLune:
OK, I think I got it - I updated it again.

PR #857, opened 2017-02-01
Bug-fix - internal Java src saving fix.

SolarLune:
We have to save out the internal files _before_ exporting each scene.
To accomplish this, I'm adding a function to erase exported internal java files.


GoranM:
I don't understand the need for a new function.

SolarLune:
OK, PR updated.

GoranM:
Why can't you simply use `saved_out_files`?

---

The commit message no longer describes the changes being made.

SolarLune:
> Why can't you simply use saved_out_files?

Because I wanted to separate the run portion from the export portion to be able to run them independently. 

To do this, we need to export the Java classes from Blender's text editor _before_ the export function runs (like it did originally, before 535ca022649a6008bdd5461b09fbde2a2eaa0b2e was merged), but delete the created java files _after_ the run function finishes, assuming we want to clean up the files afterwards (which this PR actually doesn't do, so I'll fix that too). 

Is there a huge need to delete the files exported after running? Maybe we should just export and overwrite the Blender text files without removing them, and document the change on the wiki.

GoranM:
A bug fix should just solve the immediate issue in the most straightforward way possible. It shouldn't be used as an implicit staging area for other things, because that can be really confusing.

As it stands, your PR is functionally equivalent to simply moving the save out line further up, so you should just do that.

If that needs to change in the future,  to accommodate other features, we can address it at that point.

> Is there a huge need to delete the files exported after running?

I don't think we should make users manage files that they previously didn't have to.

Are you having some problems with deleting those files from the run thread?

The source files only need to exist on the filesystem during the building process, so if you were to build the project in the main thread (`gradlew desktop:classes`), and then only run the application on the other thread (`gradlew desktop:run -a -x classes`), I think that would probably work.

SolarLune:
> It shouldn't be used as an implicit staging area for other things, because that can be really confusing.

I'm not sure what you're referring to.

> Are you having some problems with deleting those files from the run thread?

No, the problem was that you can run the game multiple times after a single export, so deleting them wasn't really an option.

> The source files only need to exist on the filesystem during the building process, so if you were to build the project in the main thread (gradlew desktop:classes), and then only run the application on the other thread (gradlew desktop:run -a -x classes), I think that would probably work.

OK, that seems like it works (just `gradlew -a desktop:run`. The added `-x` seems to break it?) , though invoking Java adds at least a second to the export time, which hurts the workflow a bit when auto-export on save is on. That makes it a slightly less than ideal solution.

> I don't think we should make users manage files that they previously didn't have to.

I guess I'm wondering why delete the files in the first place, though? If we didn't do that, then we also wouldn't need the "Make Internal Java Files External" button, since the export process would do that.

GoranM:
When you said "which this PR actually doesn't do, so I'll fix that too", I though you meant relating to deletion from a different thread, because this PR actually seems to cleanup fine.

I assumed you wanted to add some code for the threaded case in this PR, which doesn't really seem like it would belong in a bug fix.

> No, the problem was that you can run the game multiple times after a single export, so deleting them wasn't really an option.

Ah, I think I understand now.

> just gradlew -a desktop:run. The added -x seems to break it?

Did you try the exact ordering I posted: `gradlew desktop:run -a -x classes`?

That should only run the project, as opposed to building and running, as the default run task would usually do.

Of course, that wouldn't really help you if you want to build and run without exporting first, so it's probably not something you need to explore further.

>  I'm wondering why delete the files in the first place, though?

If you don't delete them, and the user deletes/renames the internal file, that leaves orphaned files on the filesystem, which basically become garbage that the user has to manage.

Although, given what you're trying to do, I guess it complicates things a bit too much.

... Ok, if you make a new PR, removing the file deletion code, along with the "make internal files external feature", I'll merge it.

SolarLune:
> When you said "which this PR actually doesn't do, so I'll fix that too", I though you meant relating to deletion from a different thread, because this PR actually seems to cleanup fine.

Ah, I see what you mean. It was just a misunderstanding.

> Did you try the exact ordering I posted: gradlew desktop:run -a -x classes?

Yeah, I did - it returned

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':android'.
> The SDK directory '/home/solarlune/android-sdk' does not exist.
```

as an error, which is strange. Doesn't seem to happen with `gradlew -p project_root -a desktop:run`.

> Ok, if you make a new PR, removing the file deletion code, along with the "make internal files external feature", I'll merge it.

OK, I will. Thanks.

Issue #853, opened 2017-01-24
Workflow Improvement


SolarLune:
Hey, there. 

So, I'm getting to the stage where working in BDX is getting to be a bit on the slow side of things. Fortunately, I'm nearly done with Gearend, but BDX could still stand to be improved to deal better with slightly larger projects. Here's my ideas:

- Support for more than one blend file at a time. 

Currently, BDX only works with one blend file at a time. This is fine, but slightly larger projects could be split apart into multiple pieces (say, one blend for each zone, or set of levels). The advantage of doing this would be the ability to export only recently updated blend files, rather than all data from all scenes in the game. Also, Blender wouldn't have to load more data than is necessary - just the data from the current blend file, which could be a potentially much smaller section of the game. This would make the actual moment-to-moment development process go much faster.

For this, I imagine the export process could be altered to push out `.bdx` files for each scene, named according to the name of each blend file (or perhaps put into a folder for each blend file). We could also just disallow scenes that share names across blend files from existing (i.e. no scene named "Background" in both Level1.blend and Level2.blend).

We could also add a "Main Scene" selector to the BDX Render panel - this would be for you to select where the game starts, regardless of which blend file is open. If it's left blank, then it would work like it does now. That's a bit more code, but it'd eliminate the need to return to the intended playable scene when testing, like we have currently. I suppose we would also have to have some kind of settings file that would keep these kinds of project-wide settings available for access from any blend file.

- Ideally, our workflow should be one that allows usage of Blender as an editor while the game is running. 

It's a bit draining to have to wait several seconds for export, for Java to be invoked for compilation, and finally for the game to load just to test even a slight change. By keeping Blender available as an editor while the game is running, you could make changes in Blender, export, and then just restart the running game to see the changes live, so to speak.

As for implementation, my idea would be to rework the export process into a background process (like when you go to export an animation), and execute it on save. You'll be able to see the progress bar on the top as it goes through the process. You could then run the game from Blender, as usual, or from an IDE.

Let me know your thoughts.

GoranM:
Based on our discussions in #839, I think we concluded that having BDX read the .blend files directly would be the best way to address all these issues.

SolarLune:
I'll carry on discussion in that thread, then.

GoranM:
The original purpose of that thread was to answer a question, not to discuss how the exporter could be improved, so I think it's probably best if we continue here.

[...]

PR #861, opened 2017-02-05
#853 - Threading of run process.

SolarLune:
The run process is now threaded. That means that we can no longer report from an operator (even if the operator's passed in manually, because of the lack of bpy data). Still, errors are printed to the console.

Export process finishes with a signal, by outputting a file to HDD. Simply checking the time modified of a .bdx file isn't sufficient, as the file may not be finished writing before that time is updated.

Adding Bdx.restartOnExport to automatically restart (and so, reload) BDX data if a new export is signaled.

### DIFF
--- a/blender/bdx/ops/exprun.py
+++ b/blender/bdx/ops/exprun.py
@@ -2,9 +2,34 @@ import os
 import sys
 import bpy
 import subprocess
+import threading
 from .. import utils as ut
 
+runThread = None
+
+class RunThread(threading.Thread):
+
+    def run(self):
+
+        gradlew = "gradlew"
+        if os.name != "posix":
+            gradlew += ".bat"
+
+        print(" ")
+        print("------------ BDX START --------------------------------------------------")
+        print(" ")
+        try:
+            subprocess.check_call([os.path.join(ut.project_root(), gradlew), "-p", ut.project_root(), "desktop:run"])
+        except subprocess.CalledProcessError:
+            pass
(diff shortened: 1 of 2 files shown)


GoranM:
Aside from the comment which explains the reason for setting the thread daemon flag, the rest seem unnecessary, if not outright confusing:

`// LibGDX won't let us delete internal FileHandles ...`

I'm not sure what you're trying to highlight there: FileHandle objects are managed by the runtime, so they shouldn't require any kind of manual deletion. If you're actually referring to the files those objects represent, it seems like you're able to delete them just fine, because this wouldn't really work if you couldn't.

SolarLune:
So I should remove the other comments?

> I'm not sure what you're trying to highlight there: FileHandle objects are managed by the runtime, so they shouldn't require any kind of manual deletion. If you're actually referring to the files those objects represent, it seems like you're able to delete them just fine, because this wouldn't really work if you couldn't.

LibGDX won't allow you to delete internal files - that's why I have to get the file from the FileHandle first. The FileHandle object has a delete() function, but it won't work if it's an Internal type.

SolarLune:
OK, PR updated to remove a lot of the comments.

GoranM:
> So I should remove the other comments?

Unless there's a good argument to justify their necessity, yes, I think it would be best to remove them.

> The FileHandle object has a delete() function, but it won't work if it's an Internal type.

That would have made it clear. Even more to the point: "Can't use done.delete(), because it doesn't work on internal files".

PR #862, opened 2017-02-07
#853 - Ability to export from multiple blend files.

SolarLune:
Adding UI toggles for multiple blend file exporting and "differential" exporting (exporting just recently changed blend files).


SolarLune:
PR updated.

Previous to this update, scenes from external blend files were appended before export and removed afterwards - this resulted in a considerable wait for Blender to load the scenes in. This also meant issues with overlapping named resources. For example, fonts couldn't be named the same thing, and would be renamed on append, which can cause problems with font generation.

Now the scenes are linked in instead, meaning a much shorter wait and no naming conflicts.

PR #863, opened 2017-02-09
#853 - Fix to save and re-open blend file after export.

SolarLune:
This clears orphan data left over from linked scenes.


SolarLune:
Actually, hold on for this - I forgot that auto-export on save would make an infinite loop for this.

SolarLune:
OK, that should do it.

PR #897, opened 2017-05-17
AABB functions.

SolarLune:


GoranM:
What is the concrete use case for these methods?

SolarLune:
Basically, the physics engine only can detect collision on the surfaces of objects (i.e. if an object touches another), not in the volumes (i.e. if an object resides inside of another). These basic AABB intersection test functions open up basic volume intersection detection.

Currently, I use it to determine if an object is residing inside of water (a huge cube, for coverage purposes) or not in Gearend (rather than using raycasts).

PR #900, opened 2017-05-20
ArrayListGameObject.get__ functions.

SolarLune:
Now they return a list of all compatible objects, rather than just the first one.


GoranM:
Instead of breaking existing code, it seems like it would be better to simply make additional functions that return a list.

So, `getBy*` would return a single object, as it does now, and `getObjectsBy*` could return a list.

SolarLune:
OK. PR updated.

PR #919, opened 2017-06-08
Changes to minimize number of shader switches when rendering.

SolarLune:
- Adding UniformSet to facilitate this. Its role is to allow an entry point to set uniforms from game code after the shader has been switched to in BDX's render loop (so now you don't have to call .begin() and .end() on a ScreenShader or MaterialShader's shaderProgram to set the uniforms).

- When drawing a RenderBuffer to a destination, the shader is set before calling batch.begin(), as that ends the previous shader, if there was one.

-  Don't render to the depth texture if unnecessary.


BluePrintRandom:
Does BDX have render batching?

This could save gearend quite a bit of render time, since the whole game
should fit on a few large sprite sheets/atlas?

SolarLune:
Yeah, BDX does have support to merge meshes together to lessen render load. It's something I'm already taking advantage of for Gearend. Thanks!

PR #927, opened 2017-07-01
Revamping Shader system a bit.

SolarLune:
Now MaterialShader and ScreenShader inherit from a base Shader class (as there are strong similarities between the two).
Shaders now support automatic hotloading if provided FileHandles. You can also send the shader strings yourself to hotload it manually, regardless of the hotloading setting.
Shaders no longer crash the engine if they can't compile.


SolarLune:
Actually, hold off on merging this for a bit - I could get rid of the MaterialShader class entirely and just have the ScreenShader be the "special" class.

SolarLune:
OK, that should do it - now MaterialShaders are just Shaders. 

ScreenShaders now extend from them, since they share similarities (a ScreenShader is just an ordinary Shader that has some special properties specifically for 2D).

Issue #929, opened 2017-07-08
Suggestion: Profiler - copy assets on creation instead of exporting


rafcolson:
On export the scene of the profiler is temporarily loaded and added to the blend file in your bdx project, with `exprun.py`:
```python
if prof_scene_export:
	with bpy.data.libraries.load(j(ut.gen_root(), "profiler.blend")) as (data_from, data_to):
		data_to.scenes = ['__Profiler']
```
This is leaving excess data inside the project's blend file, for which reason the profiler's scene doesn't contain any lights, materials and a world. However, this can't be avoided for meshes and fonts. So ideally, the profiler's scene is not added to the project's blend file so it stays 'untouched', which is to be expected from a user's point of view. If not, it may only be confusing.

Because the profiler's scene is exported also issues with its assets can occur in some cases. One case would be where you would create a new project and overwrite the project's `game.blend` (or add some other blend) and run bdx. The `exporter.py` now expects to find a copy of `BDXFontMono.fntx` instead of the file itself. This issue seems to be related to #571, which may describe other cases.
```python
File "C:\Users\Raf\AppData\Roaming\Blender Foundation\Blender\2.78\scripts\addons\bdx\exporter.py"
line 246, in fntx with open(j(fntx_dir, t.font.name + ".fntx"), 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory:
'D:\\BDX\\Projects\\Test\\Test\\android\\assets\\bdx\\fonts\\BDXFontMono.001.fntx'
```
This issue currently can be solved by first running bdx once after the project is exported, but it is very annoying. During development blend file templates are frequently used to test builds. I'm not sure if this would solve those other issues, but obviously these are also caused while exporting the profiler's scene.

So instead of exporting the profiler's scene, adding to the bdx project by copying its assets on creation time, would solve all issues. With bdx development this would mean that the profiler's assets are now created by the developer and copied to the `gen/profiler` folder. This is easy:

- create a new project
- copy `profiler.blend` and `BDXFontMono.ttf` from the `gen/profiler/dev` folder into the project's `blender` folder
- open `profiler.blend` and run the project to overwrite the assets of the profiler's scene
- copy `__Profiler.bdx`, `BDXFontMono.fntx` and `__FNT_BDXFontMono.png` from the project's `assets` folder into the `gen/profiler` folder

[...]

PR #935, opened 2017-07-11
(#929) Always ignore data block name of font

rafcolson:


SolarLune:
Hey, I think you're assuming the `fntx` file is in the same location as the source `ttf` file - this isn't always the case. For me, at least, I store my fonts external to my project folder (since multiple projects could use them).

SolarLune:
Had a previous post, but that was incorrect. I'm running into an issue with BDX being unable to read the `fntx` file for a font that I'm using a blend file:

```python
RuntimeError: Error: Traceback (most recent call last):
  File "/home/solarlune/.config/blender/2.78/scripts/addons/bdx/exporter.py", line 908, in execute
    return export(context, self.filepath, self.scene_name, self.exprun, self.apply_modifier)
  File "/home/solarlune/.config/blender/2.78/scripts/addons/bdx/exporter.py", line 857, in export
    bdx["models"].update(srl_models_text(ts, fonts_dir))
  File "/home/solarlune/.config/blender/2.78/scripts/addons/bdx/exporter.py", line 261, in srl_models_text
    for t in texts}
  File "/home/solarlune/.config/blender/2.78/scripts/addons/bdx/exporter.py", line 261, in <dictcomp>
    for t in texts}
  File "/home/solarlune/.config/blender/2.78/scripts/addons/bdx/exporter.py", line 249, in fntx
    with open(j(fntx_dir, font_name(t.font) + ".fntx"), 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/solarlune/Documents/Projects/BDX/NewTest/android/assets/bdx/fonts/please1.fntx'

```

I believe the issue is that FontWriter is being passed the font's name as the output file name, I think, here: https://github.com/GoranM/bdx/blob/master/blender/bdx/exporter.py#L765

The commands being passed includes the font's name, when you want it to be the font's filename, right?

rafcolson:
> The commands being passed includes the font's name, when you want it to be the font's filename, right?

I don't understand completely why it has been working on my end without this change, but of course it only makes sense if also the FontWriter uses the file name. Hopefully this will cut it for you. And thanks.

rafcolson:
Putting this on hold for just a few minutes.

GoranM:
Waiting for @SolarLune to confirm.

SolarLune:
I don't see any issues with this, on testing, so it should be good to go.

PR #943, opened 2017-07-17
Bug-fix: Should be "post-export program", not "pre-export".

SolarLune:


GoranM:
I don't think a commit message is an appropriate place for self-deprecating remarks.

Just write a message that describes the change, and let others judge for themselves.

SolarLune:
OK, that should do it.

rafcolson:
Hilarious.

Issue #944, opened 2017-07-18
Profiler Issues


SolarLune:
I'm getting inconsistent results with the new profiler additions - I get the following error after deleting the profiler to re-export it:

```
Traceback (most recent call last):
  File "/home/solarlune/.config/blender/2.78/scripts/addons/bdx/ops/exprun.py", line 267, in execute
    export(self, context, bpy.context.scene.bdx.multi_blend_export, bpy.context.scene.bdx.diff_export)
  File "/home/solarlune/.config/blender/2.78/scripts/addons/bdx/ops/exprun.py", line 216, in export
    bpy.data.objects.remove(bpy.data.objects["__PBars"]);
KeyError: 'bpy_prop_collection[key]: key "__PBars" not found'
```

After that, I can export fine, but when I go to actually enable the profiler and run, I get this:

```
Exception in thread "LWJGL Application" java.lang.NullPointerException
	at com.nilunder.bdx.utils.Profiler.init(Profiler.java:353)
	at com.nilunder.bdx.Scene.init(Scene.java:202)
	at com.nilunder.bdx.Bdx$ArrayListScenes.add(Bdx.java:111)
	at com.solarlune.icontest.BdxApp.create(BdxApp.java:20)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:147)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:124)
```

SolarLune:
OK, that PR should resolve the first one - I think the second one was caused by me not deleting my previous BDXMono font file.

rafcolson:
Perhaps it would be better to remove all data after all blend files are exported?

SolarLune:
What do you mean?

rafcolson:
What I meant was to remove all data of the profiler after all blend files are exported, because now the issue with the profiler remains with multi blends. I'll have a look. Also, sorry that I didn't check if that commit would also work with multi blends.

SolarLune:
Yeah, the profiler data is removed after all the blend files are exported, it's that when multi-blend mode is on, Blender saves and reloads after export (as it appends the scenes from the other blend files in the directory into the exporting blend file). This means that the objects the profiler appends are also removed before that block at the bottom, which is what throws the error. The PR I submitted resolves this issue, though.

No problem.

rafcolson:
[...]

PR #945, opened 2017-07-18
(#944) Don't remove Profiler objects if multiblend export's on.

SolarLune:


GoranM:
Aren't you just removing the unnecessary semicolons?

SolarLune:
No, I tabbed it out so that it executes the `prof_scene_export` block only if multi blend export is off.

GoranM:
I was reading the diff in split view, so it was more difficult to spot (for me, at least).

PR #949, opened 2017-07-27
GameObject AABB enhancements.

SolarLune:
aabbContains(Vector3f) now can take an AABB for reference. If it is set to null, it will use the calling GameObject's AABB points.

aabbContains(float[][]) and aabbContainsAny(float[][]) now can also take another AABB for reference. If it is set to null, it will use the calling GameObject's AABB points.

getAABBPoints() now can take a float value for an additional margin around the returned AABB.

### DIFF
--- a/src/com/nilunder/bdx/GameObject.java
+++ b/src/com/nilunder/bdx/GameObject.java
@@ -1195,22 +1195,38 @@ public class GameObject implements Named{
 
 	}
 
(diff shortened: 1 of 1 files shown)


GoranM:
Shouldn't this all be in a single commit?

SolarLune:
Whoops, my mistake. Thanks, fixed.

PR #951, opened 2017-08-02
ArrayListNamed improvements.

SolarLune:
Now Objects can be re-arranged more easily using the slot() function, which moves the object into its new position while not removing other objects.

They can also now easily be moved forwards or backwards in the list by using the moveUp() and moveDown() functions.

Note that the main application of this is to more easily control the order of scenes in the scene list (so you can, for example, ensure a certain scene is on top).


GoranM:
it seems like "moveTo" would be a more descriptive/fitting name when compared to "slot".

SolarLune:
Ah, that does seem better. Thanks, PR updated.

PR #958, opened 2017-08-09
Random.selection().

SolarLune:
Random.selection()'s role is to provide a random number (within provided bounds, if desired) of the arguments back. Basically a plural version of Random.choice(). See the example:

```java
 String[] t = new String[]{"A", "B", "C", "D", "E"};  
        
if (Bdx.keyboard.keyHit("space"))
    System.out.println(Random.selection(t));

// [A, C]
// [C, B, A]
// [C, E]
// [A]
// [B, C, E, D, A]
```

There's also a series of variants to work with lists, and another one that takes a variable amounts of arguments.

### DIFF
--- a/src/com/nilunder/bdx/utils/Random.java
+++ b/src/com/nilunder/bdx/utils/Random.java
@@ -1,6 +1,8 @@
 package com.nilunder.bdx.utils;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 
 import javax.vecmath.Vector3f;
 
@@ -64,6 +66,74 @@ public class Random{
 				Random.random(0, 1),
 				Random.random(0, 1),
 				Random.random(0, 1));
+	}
+
+	public static <T> ArrayList<T> selection(int minNum, int maxNum, T[] args, boolean repeats) {
+
+		ArrayList<T> out = new ArrayList<T>();
+
+		ArrayList<T> interim = new ArrayList<T>();
+
+		Collections.addAll(interim, args);
+
+		for (int i = 0; i < Math.floor(random(minNum, maxNum + 1)); i++) {
+			if (interim.size() == 0)
+				break;
+			T ch = choice(interim);
+			if (!repeats)
+				interim.remove(ch);
+			out.add(ch);
+		}
+
+		return out;
+
+	}
+
+	public static <T> ArrayList<T> selection(int minNum, int maxNum, T[] args) {
+		return selection(minNum, maxNum, args, false);
+	}
+
+	public static <T> ArrayList<T> selection(int num, T[] args) {
+		return selection(num, num, args, false);
+	}
+
+	@SafeVarargs
+	public static <T> ArrayList<T> selection(T... args) {
+		return selection(1, args.length, args, false);
+	}
+
+	public static <T> ArrayList<T> selection(int minNum, int maxNum, List<T> args, boolean repeats) {
+
+		ArrayList<T> out = new ArrayList<T>();
+
+		ArrayList<T> interim = new ArrayList<T>();
+
+		interim.addAll(args);
+
+		for (int i = 0; i < Math.floor(random(minNum, maxNum + 1)); i++) {
+
+			if (interim.size() == 0)
+				break;
+			T ch = choice(interim);
+			if (!repeats)
+				interim.remove(ch);
+			out.add(ch);
+		}
+
+		return out;
+
+	}
+
+	public static <T> ArrayList<T> selection(int minNum, int maxNum, List<T> args) {
+		return selection(minNum, maxNum, args, false);
+	}
+
+	public static <T> ArrayList<T> selection(int num, List<T> args) {
+		return selection(num, num, args, false);
+	}
+
+	public static <T> ArrayList<T> selection(List<T> args) {
+		return selection(1, args.size(), args, false);
 	}
 
 }


GoranM:
The need for this functionality doesn't seem common enough to justify engine inclusion.

SolarLune:
Really? It seems pretty useful - for example, to randomly select a collection of behaviors for an enemy or boss, or select from a group when randomly generating objects (i.e. select anywhere from 2 - 5 cities when designating a zone when doing random generation, which is my use case). 

Thinking about it now, I guess having the duplicates argument is just a convenience, since if you use it, it's basically the same as just constructing a list of choice() calls, so it's not really helpful. But the power of the function comes from the selection not containing duplicates.

GoranM:
One would expect that an enemy or boss would have a well defined number of behaviors, implemented in a fairly simple system (like a state machine).

Procedural generation (I assume that's what you mean by "random generation") is usually picking one item from a well defined group or groups, not from derived, randomly generated groups.

SolarLune:
> One would expect that an enemy or boss would have a well defined number of behaviors, implemented in a fairly simple system (like a state machine).

For a boss, I could randomly pick a sequence of states for such a state machine before executing them in order, like (Jump, Attack, Shoot), (Shoot, Attack, Jump), etc. That could help to keep the boss's behavior fresh and unpredictable.

> Procedural generation (I assume that's what you mean by "random generation")

Yeah, procedural generation is what I meant.

> is usually picking one item from a well defined group or groups, not from derived, randomly generated groups.

Sure, but there's not really any reason why you would never have need of picking multiple objects from a collection. 

For example, this function could also be useful for things like selecting monsters to create encounters - you could tag monsters created beforehand according to difficulty, and then, on generation, randomly select a certain number from those collections to spawn them in to create encounters.

GoranM:
> For a boss, I could randomly pick a sequence of states for such a state machine before executing them in order, like (Jump, Attack, Shoot), (Shoot, Attack, Jump), etc. That could help to keep the boss's behavior fresh and unpredictable.

Anyone could want to do anything, but that's not really the core of the argument.

Also, there's a fine like between "fresh" and "annoying":

In order to achieve a greater level of mastery, players need the ability to confidently predict enemy behavior, based on observable game state. A few surprises here and there can be fun, but too much is just annoying, and I think that's probably why "fully random" is not a predominant behavior design methodology.

> Sure, but there's not really any reason why you would never have need of picking multiple objects from a collection.

I'm not trying to argue that one would "never have need" for this feature. I'm just not sure that it's a common enough need to justify additional engine code.

rafcolson:
It may not be common enough for it to be added to the engine, but a cool feature nonetheless! If not added, maybe this could be part of a tutorial: More Random methods...

SolarLune:
Ah, I forgot to respond to this.

> Anyone could want to do anything, but that's not really the core of the argument.

> Also, there's a fine like between "fresh" and "annoying":

> In order to achieve a greater level of mastery, players need the ability to confidently predict enemy behavior, based on observable game state. A few surprises here and there can be fun, but too much is just annoying, and I think that's probably why "fully random" is not a predominant behavior design methodology.

"But that's not really good boss design" is not the point. I find it to be a good idea - repeating boss behaviors in a set sequence is boring. I don't think there's a reason to restrict BDX to "what you or I think a good game would need". "What you and I think a game, in general, would need", yeah.

> It may not be common enough for it to be added to the engine, but a cool feature nonetheless! If not added, maybe this could be part of a tutorial: More Random methods for AI ...

Thanks, but a tutorial wouldn't really be a good fit for this - the entire point is that it's a utility function. Teaching people to make it isn't really as useful as the function itself.

rafcolson:
>Thanks, but a tutorial wouldn't really be a good fit for this - the entire point is that it's a utility function. Teaching people to make it isn't really as useful as the function itself.

I'm not sure that I understand. How wouldn't this be useful if the utility is not going to be part of BDX? I would think, from a user's point of view, that it's better to at least have an example of how such utility could be created.

For such reason I made the [ProfilerToggler component tutorial](https://github.com/GoranM/bdx/wiki/Component-Tutorials%3A-ProfilerToggler). Was this a bad idea?

GoranM:
I don't know how you would define "generally needed", but in my view, it's something that correlates highly with what we can see in most existing games.

Whether or not it's because of the design reasons I listed is not as relevant (in this context) as the fact that it doesn't seem to be required by most games.

So, it's not about "people could want to do x", or "I think x is a good idea based on my personal taste" - It's about what would most people really need, based on what most developers actually used in the past.

SolarLune:
> It's about what would most people really need, based on what most developers actually used in the past.

OK, so here's [a video](https://www.youtube.com/watch?v=0YcyGV4o6hU) of an example of what this kind of function could have helped with - a boss from Megaman 4. The boss's behavior is always the same, and while the player in the video does have a bit of trouble with him, it's pretty trivial to see the boss's pattern, making it just a matter of time before you defeat him. 

Randomizing the sequence of behaviors would have made this boss impossible to cheese. To balance this, the ring could move slower to allow the player to properly foresee its behavior, but it would have made the entire battle much more enjoyable and dynamic, rather than a process of rote memorization.

> I'm not sure that I understand. How wouldn't a tutorial be useful if the utility is not going to be part of BDX? I would think, from a user's point of view, that it's better to have at least an example of how such utility could be created.

> For such reason I made the ProfilerToggler component tutorial. Was this a bad idea?

I think I feel like if it's not going to be included, I shouldn't really mention it or point to it. It implies BDX is missing something, and if it's going to be documented as a good thing, why haven't we added it?

I think your ProfilerToggler's not a bad component, and could be included in BDX as an additional component (maybe with a bit of tweaking). Being able to click a corner to toggle and cycle the profiler seems like a simple and solid additional component to add, and something very useful to have mid-development.

I'd still like to see more components packaged in BDX as standard solutions for logic (like my notification system idea, or your ProfilerToggler, or a UV scrolling component).

rafcolson:
>It implies BDX is missing something

Not necessarily, at least, that's my own opinion. Some functionalities might be very useful, but not so much generally useful. I think that's the point of the argument we have to focus on. For example, the fact that you were mentioning that "maybe with a bit of tweaking" the ProfilerToggler component could be included, shows to me that it is just "not general enough". Otherwise it would be good as is, unless you meant that the code itself could use some improvements.

I really like the Random.selection() functionality and I would perhaps make use of this technique, even for procedural world generation. For which reason I personally would prefer it to be part of BDX. I acknowledge that this functionality might perhaps not be common enough to justify engine inclusion. As we are seeing more and more games with procedural generation, I guess only time will tell if this it will become common enough.

An [interesting video](https://www.youtube.com/watch?v=TgbuWfGeG2o) on this development. With the right amount of "rules" procedural generation might become "tasteful" for everyone.

@GoranM 

>A few surprises here and there can be fun, but too much is just annoying, and I think that's probably why "fully random" is not a predominant behavior design methodology.

It seems to me that your point also implies that there would be a general application for random selections, as long as it wouldn't be used excessively, right?

GoranM:
I'm not really sure how your belief, that the behavior of this boss is substandard, relates to my core argument.

But, if you want to discuss game design in general: I don't think there's anything wrong with the behavior of that megaman boss. It's highly predictable, but that's actually the point, because once the player figures it out, they still have to follow through with execution. So, it's not just about "rote memorization", and knowing exactly what happens and when - It's also about how effectively you can deal with that (like in speedruns).

> It seems to me that your point also implies that there would be a general application for random selections, as long as it wouldn't be used excessively, right?

Doesn't seem that way to me, because unexcessive use wouldn't require anything more than the basic random generation capabilities we already have. I mean, if you want to randomize the order of set behaviors, you can just do it with `Collections.shuffle`. If you want to decide on an action based on some aspect of player state, while including some random factor, you can typically just use a random number, or choose a single element from a set list.

The need to have something like "selection", to get random subgroups, doesn't seem common enough to justify engine inclusion.

---

I think it would be unfortunate to include something if it can't be justified, but if you guys really have strong feelings about this feature, and you want to have it in the engine, I will include it.

SolarLune:
> I mean, if you want to randomize the order of set behaviors, you can just do it with Collections.shuffle

This basically would do what Random.selection would do. That's fine, then - I can just use that if I need it. I'm fine with closing this PR now.

> I'm not really sure how your belief, that the behavior of this boss is substandard, relates to my core argument.

If your core argument was "The need for this functionality doesn't seem common enough to justify engine inclusion.", my argument was basically "I have need for this functionality, and pre-existing games that lacked this functionality could've been improved by it". 

> Otherwise it would be good as is, unless you meant that the code itself could use some improvements.

Yeah, that's what I meant - mainly that it seems like the mouse cursor has to be in the top left if you press the P key to toggle the profiler.

rafcolson:
>I think it would be unfortunate to include something if it can't be justified

So let's think about this further. I can be wrong, but would my earlier point prove that `Random.selection` could be generally useful if seed is set to 0?

>Doesn't seem that way to me, because unexcessive use wouldn't require anything more than the basic random generation capabilities we already have.

I agree that with what we already have we could create all kinds of random behavior. However, I think that `Random.selection` may be a standard solution in many cases. But I think it is already implied that the random selection shouldn't be containing doubles. It then also doesn't make sense to have a function `Random.selection(int len,  ArrayList<T> list)` since this would be the same as shuffling the list. So I think this would be better:
```java
	public static <T> ArrayList<T> selectionStatic(int min, int max, ArrayList<T> list){
		ArrayList<T> out = new ArrayList<T>();
		ArrayList<T> tmp = new ArrayList<T>(list);
		Collections.shuffle(tmp);
		for (int i = 0; i < Math.floor(random(min, max + 1)); i++){
			out.add((T) tmp.get(i));
		}
		return out;
	}

	public static <T> ArrayList<T> selectionStatic(ArrayList<T> list){
		return selectionStatic(1, list.size(), list);
	}
```
An example: with an area of 16 sections, for every section, compare `Random.selection(items.size(), items.size(), items)` (same as shuffling the selection).

random selection fixed

with `Random.selection(items)`.

random selection

So this would be generally useful, I think, as long as the number of items in the selections always vary.

rafcolson:
>Yeah, that's what I meant - mainly that it seems like the mouse cursor has to be in the top left if you press the P key to toggle the profiler.

I've checked this. But this doesn't seem to be the case. Only when the mouse is clicked (or a finger is tapped), then the profiler is not toggled when hitting P. Just, so you'd know.

GoranM:
> I'm fine with closing this PR now.

Ok ... I assume you didn't close it already because you want me to do it?

> So this would be generally useful, I think

I don't really see how anything in your post would even imply that. Maybe we just don't agree on the meaning of "generally needed/useful", so I'll repeat what I said before: it's not about "people could want to do x", or "I think x is a good idea based on my personal taste" - It's about what would most people really need, based on what most developers actually used in the past.

SolarLune:
> Ok ... I assume you didn't close it already because you want me to do it?

Nah, I wanted to let the discussion continue.

I don't really see the utility now that I know you have access to Collections.shuffle and List.subList - between those two, I think you can get the same behavior as Random.selection(), basically. So I'll close this, then.

PR #968, opened 2017-08-27
InputMaps improvements.

SolarLune:
You can now use a collection to describe input descriptors to an InputMap.

You can now pass a map of aliases for input keys. Using this, you can easily automatically add inputs if other inputs are present. For example, you can define the d-pad's inputs to be aliases for the left analog stick's. If any of one group's inputs are detected, the corresponding input for the other group is added.

Inputs now turn into a readable string.


BluePrintRandom:
:D this is just like what I do for wrectified and manic mack, but I use
lists.

GoranM:
> For example, you can define the d-pad's inputs to be aliases for the left analog stick's. If any of one group's inputs are detected, the corresponding input for the other group is added.

Couldn't you simply add the relevant descriptors for both the d-pad and the left analog stick?

> Inputs now turn into a readable string.

Changing what toString returns doesn't change the object type.

SolarLune:
> Couldn't you simply add the relevant descriptors for both the d-pad and the left analog stick?

Nah, it would require checking to see if an input that uses the left analog stick was detected, and if so, setting the corresponding input for the d-pad, and vice versa. Basically, it'd be what I'm doing here, so I think it's better to add this functionality in for future situations like this. 

Generally, I suppose you'd always want both left analog stick and d-pad to respond to the same inputs, so an alternative might be to change the d-pad's inputs to something else (like, say, "dpad-left", or something) and build some kind of functionality to have general directions (like "left") correspond to either of those inputs ("ls-left" or "dpad-left"). But that would probably be better abstracted out to work with anything, not just the left analog stick and d-pad.

> Changing what toString returns doesn't change the object type.

Yeah, I meant that they are now readable, rather than being just a hashcode that doesn't really help you too much in cases of trying to see which Inputs are set in the InputMaps.

GoranM:
I'm still not sure if I fully understand. It would be helpful if you could provide a scenario that clearly describes the problem you're trying to solve with aliases.

Assuming for a second that you really need something along those lines: There seems to be no requirement (or significant benefit) to have this in the engine, because you can transform the descriptor collection in whatever way you want before passing it to put.

SolarLune:
The basic use case is I want to detect inputs on the left analog stick or d-pad. 

Rather than checking all inputs to see if any of them have "ls" in their names and then adding the corresponding d-pad inputs, and detecting the opposite, I can just pass a hash map of aliases that will do this for me.

GoranM:
> I want to detect inputs on the left analog stick or d-pad.

This would do that:

```java
imaps.put("up", "g:ls-up", "g:up");
imaps.put("down", "g:ls-down", "g:down");
imaps.put("left", "g:ls-left", "g:left");
imaps.put("right", "g:ls-right", "g:right");
```

Unless there's some other requirement that I don't know about, there seems to be no need for "checking all inputs" for "ls", or anything else.

SolarLune:
It's because I have input binding as well, so I would have to check each binding to see if it has "ls" in there, and if so, add the d-pad binding, and vice-versa. 

I guess I could just assume the player would want these controls on the left stick and d-pad, and not allow them to be bound.

I'll close this.

Issue #874, opened 2017-03-29
Utilizing Delta Time / Removing FPS Lock


SolarLune:
Hey, there. 

So basically, BDX is locked at, and expects to run at, 60 FPS. It might be a good idea to "undo" this FPS lock (particularly because on lower-end devices, anything below 60 becomes unplayable, as the game runs more slowly in all respects). I know this isn't trivial, but I think it's worth thinking about. I'm not certain if the ["update delta time spiral of death"](http://bulletphysics.org/Bullet/phpBB3/viewtopic.php?t=10293) has something to do with this particular issue, but if so, there might be a simple solution that allows us to unlock the FPS rate and still keep the engine's update core relatively simple.

Also, maybe we could have a function that controls whether the FPS is locked to a target value (old-style behavior) or unlocked with the delta given (proposed new-style behavior)?

GoranM:
> I'm not certain if the "update delta time spiral of death" has something to do with this particular issue

That's one of the core reasons for the flat 60 target.

> if so, there might be a simple solution that allows us to unlock the FPS rate and still keep the engine's update core relatively simple.

What did you have in mind?

SolarLune:
Well, you mentioned:

> Given a call like stepSimulation(delta, 7), where delta is 1/60f, I assume that there would be only one substep, and if **I'm running my game logic in the simulation tick callback**, that logic would only run once for this frame.

So I was thinking of just separating the logic and physics calls? Or maybe just having one physics substep (for now, at least, just to allow variable / unlocked FPS values)? 

We could also try spawning another thread for physics, on desktop, at least. That's neither cross-platform nor particularly simple, though it would probably be pretty great for performance, so that's something to keep in mind for later.

GoranM:
Limiting to a single substep would automatically lose time in cases where the simulation delta is larger than the fixed time step. If you want to run at 30 fps, and have bullet simulate over a 1/30 second interval, with a fixed time step of 1/60, you need 2 substeps to properly simulate the full interval.

Running logic outside the substep callback means that gamecode will miss all the bullet state changes that occurred in all but the last substep.

SolarLune:
> If you want to run at 30 fps, and have bullet simulate over a 1/30 second interval, with a fixed time step of 1/60,

Why not give Bullet 1/30 for the fixed time step in that case? 

Or are you saying "want to run at" to mean the game's running slowly, and we're aiming for 60 FPS?

Also, what exactly do you mean by "lose time"?

[...]

PR #979, opened 2017-09-28
(#874) Bdx.delta() function for delta timing.

SolarLune:


GoranM:
Why do we need this?

SolarLune:
I guess we don't actually (for now, anyway)~

PR #987, opened 2017-10-23
Bug-fix: Fixed newline handling.

SolarLune:
Before, \n characters were being eaten by Text, while Java considers
them to be actual characters. Now they're treated the same (\n becomes
a space for Text).
Also a bit more Text error handling.

### DIFF
--- a/src/com/nilunder/bdx/Text.java
+++ b/src/com/nilunder/bdx/Text.java
@@ -62,6 +62,9 @@ public class Text extends GameObject{
 
 		for (String l : lines) {
 
+			if (l != lines[lines.length - 1])
+				l += " "; // Java counts \n as an actual, single character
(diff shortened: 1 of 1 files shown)


rafcolson:
Did you check if this messes up the Profiler? Shouldn't `\n` be treated as a new line?

SolarLune:
Thanks for the reminder - just checked and the profiler seems fine.

`\n` is treated as a newline separator still, it just used to be eaten entirely, which causes issues. 

In Java a newline is explicitly a character. For example, in the string, `ABC\nDEF`, the fourth character should be the newline character, not `D`. This PR adds spaces to simulate the newline character's "existence".

rafcolson:
Thanks. I understand.

PR #988, opened 2017-10-24
"Immediate" render function ( Scene.renderMesh() ).

SolarLune:
This adds a function, "Scene.renderMesh()", which allows you to render a mesh with a specific transform using only code. The idea is that this would prove especially useful when you need to place objects in a specific location (like place a button icon in a message window) and don't want to have to keep track of adding and removing objects.

Here's a code example:

```java

package com.solarlune.tests;

import com.nilunder.bdx.*;
import javax.vecmath.Vector3f;

public class Plane extends GameObject {

    public void main() {
    
        if (Bdx.keyboard.keyDown("space")) {
        
            Vector3f vec = new Vector3f();
        
            for (int i = 0; i < 240; i++) {
                vec.set((float)Math.sin(Bdx.time + i) * 4, (float)Math.cos(Bdx.time + i) * 4, i * .061f);
                scene.renderMesh("Cube", vec);                
            }
            
        }
            
    }
    
}

```

To get this:

peek 2017-10-23 22-39

Let me know if this makes sense at all and could be useful enough for engine inclusion.


GoranM:
> don't want to have to keep track of adding and removing objects

You're doing it for everything else anyway, so it doesn't really seem like a big deal.

SolarLune:
Yeah, I suppose so. I'd like it to be easier to draw and render things, but I guess it's not too difficult currently - I'll have to examine the current approach to creating another SpriteBatch or ModelBatch for rendering inside of BDX.

Issue #964, opened 2017-08-22
Join() vs. ModelCache


SolarLune:
Yo, @rafcolson, have you seen LibGDX's [ModelCache](https://github.com/libgdx/libgdx/wiki/ModelCache) class? 

It seems like it might be a simpler way to do what GameObject.join does for us?

rafcolson:
Worth to check this out, but I'm assuming that model cashing in LibGdx uses merging methods which are not preserving normal data. If that would be the case than adapting the cashing methods to our own methods with our own version of ModelInstance would probably work. But it seems like an extensive change to BDX's code.

rafcolson:
>I'm assuming that model cashing in LibGdx uses merging methods which are not preserving normal data

Luckily my assumption seems to be wrong. ModelCache doesn't use the transform methods of [Mesh](https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/Mesh.java#L920) but of the [MeshBuilder](
https://github.com/libgdx/libgdx/blob/master/gdx/src/com/badlogic/gdx/graphics/g3d/utils/MeshBuilder.java#L486) itself.

SolarLune:
Yo, 'sup! 

So I've been getting this working, and it seems like it might be a better solution, as it's not limited to just 32768 vertices (which is fairly low, when you think about it from the perspective of joining static meshes together). 

The issue I'm running into now, though, is doing the matrix math to properly set the transforms of the meshes to be added prior to adding them into the ModelCache. Multiplying the meshes' transforms by the inverse of the calling object's transform _almost_ works - it just fails in a few edge cases.

@rafcolson, since you wrote the original join process, could you take a look at the matrix code and lemme know what I'm doing wrong? I saw that you did some stuff with matrices in the previous join code, but I don't know if all of that's still necessary since we're just dealing with whole Meshes, rather than their individual vertices?

```java

Matrix4f inverted = transform();
inverted.setIdentity();

for (Map.Entry<Mesh, ArrayList<Matrix4f>> entry : data.entrySet()) {

	for (Matrix4f t : entry.getValue()) {

		ModelInstance m = entry.getKey().getInstance();
		t.mul(inverted);
		m.transform.set(t.get());
		modelCache.add(m);

	}

}

```

rafcolson:
Hi @SolarLune
 
[...]

Issue #917, opened 2017-06-04
Suggestion: Joiner upgrade - JoinData


rafcolson:
The joining functions `GameObject.join(objects)` and `GameObject.join(map)` can keep draw calls to a minimum. Especially on low-end and mobile devices this can be useful.

However, there's much more that could be done by using this joining functionality, that is, if adding or removing vertices could be done efficiently. This was already discussed in [a previous suggestion related to the joining functionality](https://github.com/GoranM/bdx/issues/809#issuecomment-253866699).

Imagine a puzzle game with hundreds of pieces laying around on the table. If all of those pieces were to be separate objects, the game could become slow because of two problems. First, there's the rendering issue with a lot of draw calls. Fortunately the joiner functionality already tackles this problem. But the other problem is physics calculation. Hundreds of invisible static ghost (yet detectable) objects could have a large impact on game play.

You can test [this apk](https://www.dropbox.com/s/05j2wpa4rwcy62s/512%20invisible%20static%20ghost%20planes.apk?dl=0) with 512 invisible static ghost planes on an Android device. I'm getting barely 10 fps.

This problem could be tackled by changing the detection method of the pieces, reducing physics calculation to a minimum. But this would also require to edit the joined mesh object, when removing or adding segments.

In order to edit the joined mesh faster the transformed vertex arrays should be stored initially. The purpose of this upgrade would be to provide not only this, but also a basis by which the joined data is accessed and modified, stored and restored rather easily.

The result is [this apk](https://www.dropbox.com/s/sb41mbuamm8g1sj/512%20segments%20with%20JoinData.apk?dl=0) with 512 segments, which can be removed or added. Additionally [the corresponding blend file](https://www.dropbox.com/s/ms5lcs303guru3v/512%20segments%20with%20JoinData.blend?dl=0), which works with the referenced branch of my repository (based on the Mesh upgrade commits).

More information about the changes will follow. Tell me what you guys think.

rafcolson:
Thanks @GoranM. Commit updated with additional information.

SolarLune:
Huh, I think I get the idea. Yeah, OK, this seems fine.

rafcolson:
PR submitted.

rafcolson:
PR merged. Thanks!

PR #993, opened 2017-11-07
(#917) Joiner upgrade

rafcolson:
Summary:

- `JoinData` enables faster updates with adding and removing meshes, transforms or transforms of meshes, as only the relevant parts of the joined mesh are updated before building the `Mesh`
- `Mesh.getJoinedModel(JoinData)` accesses `JoinData.parts` to build the `MeshParts` for the `Model`
- `JoinData` can be serialized and deserialized

@SolarLune 
I'd appreciate it if you could evaluate these changes. After this is merged we could look into using `ModelCache` (see Issue #964).

### DIFF
--- a/src/com/nilunder/bdx/utils/JoinData.java
+++ b/src/com/nilunder/bdx/utils/JoinData.java
@@ -1,0 +1,252 @@
+package com.nilunder.bdx.utils;
+
+import java.util.Map;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.ArrayList;
+
+import javax.vecmath.Matrix4f;
+
+import com.badlogic.gdx.utils.Json;
+import com.badlogic.gdx.utils.JsonValue;
+import com.badlogic.gdx.utils.JsonReader;
+
+import com.nilunder.bdx.Bdx;
+import com.nilunder.bdx.Scene;
+import com.nilunder.bdx.GameObject;
+import com.nilunder.bdx.gl.Mesh;
+import com.nilunder.bdx.gl.Material;
+
+public class JoinData extends HashMap<Mesh, ArrayList<Matrix4f>>{
+	
+	public HashMap<Material, JoinData.Part> parts;
+	
+	private Matrix4f relativeTransform;
+	
+	public class Part extends LinkedHashMap<Matrix4f, float[]>{
+		private int numVertices;
+		
+		public int numVertices(){
+			return numVertices;
+		}
+		public int numIndices(){
+			return numVertices / Bdx.VERT_STRIDE;
+		}
+		@Override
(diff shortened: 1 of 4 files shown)


src/com/nilunder/bdx/utils/JoinData.java
```diff
@@ -0,0 +1,252 @@
...
+		for (GameObject g : objects){
+			m = g.mesh();
+			l = map.get(m);
+			if (l == null){
+				l = new ArrayList<Matrix4f>();
+				map.put(m, l);
+			}
+			l.add(g.transform());
+			if (endObjects){
+				g.end();
```
SolarLune: This should be endNoObjects(), rather than end(). This way, it doesn't end objects more than once if the child and parent are in the list of objects to be joined and the parent is before the child.

SolarLune:
OK, sorry - saw this request and then closed it and couldn't find it again, haha. 

I did a cursory review and posted my thoughts on a review. Thanks.

SolarLune:
OK, thinking about this some more, should GameObjects even have a `join()` function if the JoinData has data about GameObjects and can create a Mesh by itself?

rafcolson:
PR updated with `endNoChildren`.

>should GameObjects even have a join() function if the JoinData has data about GameObjects and can create a Mesh by itself?

We won't if we modify `GameObject.mesh(Mesh newMesh)` to call `GameObject.calcOriginDimNoScale()` also if `newMesh.name().equals(GameObject.mesh().name())` and `newMesh` is a joined mesh. But there's no way to determine this, once the mesh is created with `new Mesh(JoinData data, Scene scene, String name)`. Unless we would add another argument to the method: `GameObject.mesh(Mesh newMesh, boolean containsJoinData)`.

That said, even if there wouldn't be any issues with creating a joined object without a join function on `GameObject` it still would be convenient to have it anyway, in my own opinion.

SolarLune:
> Unless we would add another argument to the method: GameObject.mesh(Mesh newMesh, boolean containsJoinData).

We could just add a `containsJoinData` field to Mesh to indicate that it contains the result from a JoinData operation.

> Or we might have the convention to always give an identical name to the joined mesh...

That'd also be fine - maybe we should always do that by default for all meshes, and just allow the name to be changed after the fact using a public name field.

rafcolson:
PR updated.

`GameObject.join(...)` methods are removed, which only leaves `Mesh(JoinData joinData, Scene scene, String name)` as the constructor to create a joined mesh, which then can be assigned to an object with `GameObject.mesh(Mesh newMesh)`. If necessary the body then can be updated with `GameObject.updateBody()`.

@SolarLune Let us know if this can be merged. Thanks a lot!

SolarLune:
Sounds good to merge! No problem.

PR #999, opened 2017-11-23
(#874) - Delta timing / Unlocked FPS.

SolarLune:
The desired target framerate is now exported from the currently active blend file, though the framerate is only explicitly set for the Desktop target. On other platforms, it should remain at 60 FPS. Substeps have also been implemented. As follows is the proposed write-up on substeps and the API additions.

## What's a substep?

The graphical framerate is now unlocked from the logical and physics tickrate through the addition of _substeps_. If the framerate drops such that the game cannot maintain its graphical framerate, the physics world will be stepped through again, and logic re-calculated, as well. This keeps the game world up to speed with the graphical representation. So, if the game normally runs at 60 FPS, but is slow at rendering and now is rendering at 20 FPS, then BDX can try using substeps to keep the game _running_ at full speed, even if the game's _rendering_  slowly. (60 FPS target / 20 FPS minimum = 3 substeps needed to keep game running at full speed)

Put more simply, __a substep is basically just an event where BDX calculates physics and logic__. Before, BDX was built to use only one substep for each rendered frame. Now, it can use more, and will automatically use the appropriate number between the minimum and maximum substep counts. Physics and logic are both executed in a substep, and with more substep execution comes a finer physics simulation. If you're making a precise physics-focused game, you could consider using higher minimum substep counts for a more precise physics simulation.

## How often will BDX use substeps?

How often BDX will use substeps is automatically determined, with the goal being to keep the physics world and logic execution up to speed with the target FPS as best as possible. It will be a value between the specified minimum and maximum substep counts. The maximum number of substeps is exported from the currently active blend file and read when BDX first starts. The minimum is, by default, 1.

_______

API additions:

- Bdx.display.averageFPS() - Returns average FPS (Gdx.graphics.getFramesPerSecond()).
- Bdx.rawDelta() - Returns the delta between the last rendered frame and this one.
- Bdx.delta() - Returns rawDelta() divided by the number of substeps for this frame. It's also capped to a maximum of `TICK_TIME * maxSubsteps()` (so multiplying movement distances by delta() won't result in insane distances covered in odd cases where the physics engine tries to simulate 1 full second with just a few substeps).
- Bdx.substepCount() - The number of substeps for the current game frame.
- Bdx.substepIndex() - The substep index for the current game frame under execution.
- Bdx.minSubsteps() / Bdx.minSubsteps(int) - Gets and sets the minimum number of substeps per frame.
- Bdx.maxSubsteps() / Bdx.maxSubsteps(int) - Gets and sets the maximum number of substeps per frame.

_______

For existing projects to take advantage of unlocked FPS, you'll need to recreate a project to grab the new BdxApp and DesktopLauncher source files (or just add the missing code lines to your project's existing BdxApp and DesktopLauncher files).

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -852,6 +852,7 @@ def export(context, filepath, scene_name, exprun, apply_modifier):
         "clearColor": clear_color,
         "resolution": [scene.render.resolution_x, scene.render.resolution_y],
         "frame_type": scene.game_settings.frame_type,
+        "maxSubsteps": scene.game_settings.physics_step_max
     }
 
     if exprun:
--- a/blender/bdx/gen/BdxApp.java
+++ b/blender/bdx/gen/BdxApp.java
@@ -8,8 +8,11 @@ import com.nilunder.bdx.*;
 
 public class BdxApp implements ApplicationListener {
 
+	public int TICK_RATE = 60;
+
 	@Override
 	public void create(){
+		Bdx.TICK_RATE = this.TICK_RATE;
 		Bdx.init();
 
 		Scene.instantiators = new HashMap<String, Instantiator>();
--- a/blender/bdx/gen/DesktopLauncher.java
+++ b/blender/bdx/gen/DesktopLauncher.java
@@ -11,9 +11,15 @@ public class DesktopLauncher {
 		config.title = "Project Name";
 		config.width = 666;
 		config.height = 444;
+		config.foregroundFPS = 60;
+		config.backgroundFPS = 60;
+		config.vSyncEnabled = false;
+
+		BdxApp app = new BdxApp();
+		app.TICK_RATE = config.foregroundFPS;
 		config.addIcon("bdx/icon_128.png", Files.FileType.Internal);
 		config.addIcon("bdx/icon_32.png", Files.FileType.Internal);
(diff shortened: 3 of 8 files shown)


rafcolson:
Looking forward to this! Nice work.

>On other platforms, it should remain at 60 FPS.

Any idea if there is a possibility to lower the fps on Android? In any case, I'll do some investigation there.

SolarLune:
> Looking forward to this! Nice work.

Thanks!

> Any idea if there is a possibility to lower the fps on Android? In any case, I'll do some investigation there.

I remember looking around for this information and finding that it wasn't possible, that Android had to have 60 FPS. I'm looking around now, though, and seeing that [it should be possible to set the FPS in an AndroidApplicationConfiguration](https://stackoverflow.com/questions/23901398/libgdx-too-much-fps-how-to-limit).

rafcolson:
>seeing that it should be possible to set the FPS in an AndroidApplicationConfiguration.

I didn't test if `AndroidApplicationConfiguration.foregroundFPS` actually exists. It's not exposed in the API. Maybe that's why it was said in that thread: "I don't know if this is a new feature or an old one."

SolarLune:
Oh, OK, maybe that's the case.

rafcolson:
I don't know. I first have to check if it exists.

PR #1001, opened 2017-11-24
Gamepad PS3 profile, auto-profile function, gamepad crash fix.

SolarLune:
Adding a PS3 profile.
Also, adding an automatic profile function (`autoProfile()`) that attempts to provide the name of the profile the gamepad should probably be using.
Also, don't crash if you try to feed the profile an inappropriate button (i.e. "cross" on an "XBOX360" profile).


rafcolson:
@SolarLune It seems that `autoProfile()` is actually `detectProfile()`. So I've updated the API.

SolarLune:
Ah, thanks. My mistake.

PR #1013, opened 2017-11-29
Remove objects from camera's ignoreObjects on end.

SolarLune:


SolarLune:
I was going to push this PR, but thinking about it, this might slow things down considerably, considering very few, if any, objects are actually in camera's ignoreObjects list, so it's probably better to leave it up to the developer.

Issue #1010, opened 2017-11-28
Suggestion: store origin and dimensionsNoScale in Mesh


rafcolson:
It doesn't really make sense to be storing these values on the `GameObject`. First, setting the origin in Blender will be an operation on the mesh (although you'll have to be in Object Mode). Secondly, `dimensionsNoScale` are actually the dimensions without any transformation (position, orientation and scale). `GameObject.dimensionsNoScale` would become `Mesh.dimensions`.

Making this change would also solve another issue. When a mesh is transformed, currently, the dimensions and origin are not updated. This is currently only done when replacing the mesh.

This would also open up the possibility to calculate the frustum without using the body's AaBb, which is also not updated with `NO_COLLISION` objects, solving that issue.

Thoughts?

GoranM:
Yeah, I guess that would make sense.

rafcolson:
>This would also open up the possibility to calculate the frustum without using the body's AaBb, which is also not updated with `NO_COLLISION` objects, thus solving that issue.

This was a wrong assumption, `body.getAabb()` will work regardless.

rafcolson:
PR merged. Thanks a lot!

PR #1015, opened 2017-12-01
(#1010) Mesh.median & Mesh.dimensions & Bullet upgrade

rafcolson:
Preceded by some clean-up.

### DIFF
--- a/blender/bdx/exporter.py
+++ b/blender/bdx/exporter.py
@@ -102,18 +102,26 @@ def mat_tris(mesh):
 
     return m_ps
 
+def median(obj):
+    return list(sum([mt.Vector(l) for l in obj.bound_box], mt.Vector()) / len(obj.bound_box))
+
+def dimensions(obj):
+    return [d / s for d, s in zip(obj.dimensions, obj.scale)]
+
 def srl_models(objects, use_mesh_modifiers):
     name_model = {}
 
     tfs = 3 * 8 # triangle float size: 3 verts at 8 floats each
 
     for o in objects:
-        if o.type != "MESH":
+        if not (o.type == "MESH" and o.data.name not in name_model):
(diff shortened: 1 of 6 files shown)


rafcolson:
Better put this on hold because I think I've found a way to fix the collision shape issue for primitives when origin is off-center.

rafcolson:
PR updated.

Aside of moving data about origin and dimensions from `GameObject` to `Mesh` this commit will also fix the bounds of primitive shapes when origin is off-center.

In some cases it might be useful to exploit the origin, for example with dynamically positioning static objects. _Primitive shapes_ will always take the exact center of the primitive shape to get the center of gravity. This makes the collision bounds off-center if the object's origin is also off-center.

For a comparison, this was how such an off-center object would show up currently:

debug_aabb_before

And this is how it will be with this patch:

debug_aabb_after

Here's my [Debug](https://www.dropbox.com/sh/2hi7uswegitciis/AADNCEl_HegGIL6cmeScaTEJa?dl=0) folder, where I'm gathering templates for debugging. Click with the mouse on screen to walk through the tests. When comparing between branches, re-open the blend.

API changes:
- removed `GameObject.origin` and `GameObject.dimensionsNoScale` fields
- added `public Matrix4f GameObject.boundsTransform()` -> gets world transform of mesh.median
- added `public Vector3f Mesh.median` -> resembles Blender's median coords
- added `public Vector3f Mesh.dimensions`
- added `public void Mesh.updateMedianDimensions()` -> corrects bounds transform for primitive shapes

As a side-note: unfortunately I couldn't fix the remaining issue with scaled objects of type `SPHERE`, `CYLINDER` and `CONE`. I think we should document that objects should not be scaled when using these shapes. But I'll put that in another suggestion.

Issue #1019, opened 2017-12-04
Suggestion: Refine InputMaps Gamepad descriptor search


rafcolson:
Currently, if a [descriptor](https://github.com/GoranM/bdx/wiki/API-InputMaps#customDescriptorsHDU) contains the character `"g"` it will be detected as `Gamepad`, which is not so convenient. I'd like to refine the search to `{"g", "g0", "g1", "g2", "g3", "g4", "g5", "g6", "g7", "g8", "g9"}`.

GoranM:
...?

The "g" is supposed to specify the gamepad device.

Please provide a use-case that clearly illustrates the problem at hand.

rafcolson:
>Please provide a use-case that clearly illustrates the problem at hand.

Of course. If I'm adding `"vgp"` or `"virtual_gamepad"` or some description containing a `"g"` to [customDescriptorsHDU](https://github.com/GoranM/bdx/wiki/API-InputMaps#customDescriptorsHDU), [InputMaps](https://github.com/GoranM/bdx/wiki/API-InputMaps) will not detect the descriptor properly, because it will specify it as a gamepad device.

>The "g" is supposed to specify the gamepad device.

Everything will stay the same, each item in `["g", "g0", "g1", ...]` will still specify a gamepad device.

GoranM:
Oh, I see.

Ok.

rafcolson:
PR merged. Thanks!

SolarLune:
Hey, "g:left" no longer identifies an input as "left" on the gamepad - you have to specify a number, which is incorrect.

rafcolson:
Ok. I see.

rafcolson:
PR submitted.

Gerkinfeltser:
Hello! I attempted to use your patched bdx build @rafcolson & I'm still having issue (though I could be confused as to what is going on or proper usage of imaps). This also happens with the latest offical release of BDX as well.

`Bdx.imaps.put("jump", "g:ls-right");` results in the following error and the game ending as soon as it starts in-spite of the "BUILD SUCCESSFUL".:
```
Controllers: added manager for application, 1 managers active
Exception in thread "LWJGL Application" java.lang.RuntimeException: Invalid descriptor "g:ls-right".
        at com.nilunder.bdx.inputs.InputMaps$Input.<init>(InputMaps.java:132)
        at com.nilunder.bdx.inputs.InputMaps.put(InputMaps.java:283)
[...]

PR #1021, opened 2017-12-07
(#1019) Refining InputMaps Gamepad descriptor search

rafcolson:
API change in [customDescriptorsHDU](https://github.com/GoranM/bdx/wiki/API-InputMaps#customDescriptorsHDU)
- The String argument is not expected to be one of the following: `"k"`, `"m"`, `"g"`, `"g0"`, `"g1"`, `"g2"`, ...

### DIFF
--- a/src/com/nilunder/bdx/inputs/InputMaps.java
+++ b/src/com/nilunder/bdx/inputs/InputMaps.java
@@ -1,6 +1,7 @@
 package com.nilunder.bdx.inputs;
 
 import java.util.*;
+import java.util.regex.Pattern;
 
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.utils.Array;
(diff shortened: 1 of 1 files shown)


GoranM:
Would `hdu == null` ever evaluate to true? That doesn't really seem clear to me ...

The whole PR seems kinda convoluted; Couldn't you just replace the relevant "contains" call with a function that checks specifically for a "g" or a "gx" where x is a digit from 0 to 9?

rafcolson:
Yes, it's also not working. I made an assumption, even a bad one. I corrected it, and I'm sorry that I didn't properly check the commit before submitting the PR.

>Couldn't you just replace the relevant "contains" call with a function that checks specifically for a "g" or a "gx" where x is a digit from 0 to 9?

I probably could do a for loop, but I want to avoid as many checks as possible. Right now keyboard, mouse, the first gamepad and custom input will need three checks each. Gamepads 2 to 10 need only four checks. I think that's the best I can do.

So, PR updated!

GoranM:
Performance is not really a concern in this context, but readability always matters, so I think it would be best to just make it a simple 2 line change:

```diff
+ import java.util.regex.Pattern;
...
- }else if (d[0].contains("g")){
+ }else if (Pattern.compile("^g[0-9]$").matcher(d[0]).find()){
```

rafcolson:
You're probably right. I didn't know about that function, so thanks! PR updated. (and it works as I've tested it).

PR #1031, opened 2018-02-02
Profiler: LinkedHashMap for props

rafcolson:
Minor change, so I'm assuming opening an issue isn't necessary.

In any case, this change will give a little more control over how Profiler props are visualized.


rafcolson:
Thanks.

